PR #6, opened 2018-06-07
emissary in docker

ets-mlb:
Add a dockerfile for the assembly dist.


mjwall:
@ets-mlb the docker stuff that was in there was designed for developers, especially on Windows.  This docker implementation looks to be designed for running a single node.   Any thoughts on how to support development via docker?

Would be nice if some of the things were configurable, like the port and number of agents.  Would not want to rebuild from maven, so maybe read from environment variables.  Could expand the entrypoint.

Also, how much space is saved if you run `rm -rf /var/cache/yum` in the file?

To support a cluster, would probably want a dockerfile that runs a feeder as well.

ets-mlb:
I lean towards using the maven:3.5.3-jdk-8 to build Emissary and not maintain a developers image. If custom functionality is needed, then a dev can inherit from the maven image and do whatever they need. 

The base image is geared towards running a standalone Emissary, hence the default 8001 port. I added a docker-compose file to show how to use Emissary in cluster mode. If you bind to the host network, you don't have to worry about exposing ports.

ssfoster:
+1 Docker Emissary in cluster mode

PR #7, opened 2018-06-07
emissary transaction manager

ets-mlb:
Create an aspectj transaction manager to watch the data flow through Emissary.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -247,6 +247,16 @@
         <version>${dep.xmlbeans.version}</version>
       </dependency>
       <dependency>
+        <groupId>org.aspectj</groupId>
+        <artifactId>aspectjrt</artifactId>
+        <version>1.8.13</version>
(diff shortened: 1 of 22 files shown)


mjwall:
@ets-mlb a transaction manager is certainly a welcome addition.  Since there is no ticket, can you talk about why you choose aspectJ instead of just putting the code in the classes.  Didn't look like that many pointcuts from a quick review.

ets-mlb:
We wanted to leave the transaction manager out of the main code and just use it as a watcher for now. AspectJ was chosen to weave the code into the existing code base, leaving minimal code changes to Emissary. This way we could integrate it in later, or easily rip it out if we don't like the implementation.

ssfoster:
Any news on this PR?  Is the concern about adding another dependency?

ets-mlb:
Newer pr will be submitted

PR #9, opened 2018-06-20
Allow more flexibility when going beyond the buffer sizes

pcgrenier:
1) WindowedSeekableByteChannel size() is unreliable so removed use from the NIOSessionParser
2) Resizes data array after filling from channel
3) Allow WindowedSeekable to grow to not roll off a set position


drewfarris:
We may consider closing this in favor of #15, but I really like the notion of the growable `WindowedSeekableByteChannel` - we might want to introduce this while retaining the original implementation that does not grow.

PR #13, opened 2018-07-11
Off-heap/high capacity base data objects

adyoun2:
I'm happy to discuss the changes, but to summarise, there are 3 intents:

1. If configured, it is possible that BaseDataObjects could be backed externally to the heap (files or off heap memory), allowing a lower memory footprint for objects not actively being processed.
2. Data objects are not limited to being 2GB.
3. All changes are compatible with existing bytecode.

While this pull request doesn't entirely complete part 2 (Pickup etc would have to be changed to use the streaming apis), I think it is a large step to allowing the processing of large objects. Places that may have to deal with large objects would also benefit from the use of the streaming apis.

Could someone take a look at this and get back to me if you have any questions?


adyoun2:
@mjwall @drewfarris , I am blind to what the process/schedule is for pull requests to be reviewed. Could you let me know on what sort of timescales this will be looked at?

drewfarris:
Hi @adyoun2 - is this change good to go from your perspective, or do you anticipate additional changes? There merge schedule PR's is sort of ad-hoc depending on what else is in the release etc. Since this is a somewhat major change, we may want to isolate it to a release by itself, e.g: emissary 6.x?

If you are happy with its current state and ready for a detailed review - let us know. Personally, I wasn't sure if you were done yet, so I'm glad you asked.

adyoun2:
@drewfarris 

I am certainly happy that without configuring an instance to use any of the more complicated functionality, there will be very little visible change, so I would consider it to be ready for review. Happy to talk through any of the changes.

Cheers

adyoun2:
Thanks for getting back to me

adyoun2:
I've had a look at what would be required to have the rest of emissary use the streaming APIs, and therefore gain the ability to use the capacity increase. Luckily this also demonstrates some of the features I've added where the need isn't as immediately obvious.

@drewfarris @mjwall , are you able to see my local diff:
https://github.com/adyoun2/emissary/compare/master...adyoun2:UseStreamingAPI?expand=1

adyoun2:
Also note I will be unavailable for a week now, so will not be able to answer questions raised immediately.

adyoun2:
The large file issue is becoming critical for us. Could you let me know whether a review of this pull request has been scheduled?

@drewfarris

drewfarris:
@adyoun2 - yes this is in the queue, have people lined up to look at it starting this week.

fbruton:
The following test classes have failures:
AllDataContainerTest
TikaFilePlaceTest
UnixFilePlaceTest

adyoun2:
Odd, they all work for me. What failures are you seeing? Hoping it is just test configuration being environment specific

adyoun2:
The only failures I'm seeing don't involve my changes, and will be because of my dev environment not being as expected:

[ERROR] Failures: 
[ERROR]   ResourceReaderTest.testResourceLocation:35 All data resources not found expected:<4> but was:<0>
[ERROR]   ResourceReaderTest.testResourceLocationAsTest:26 All test resources not found expected:<4> but was:<0>

fbruton:
Listed below is a sample of some of the failed test:
AllDataContainerTest.testClone:383 expected:<[Base Dat]a with append> but was:<[]a with append>
TikaFilePlaceTest>IdentificationTest.testIdentificationPlace:92 Cannot run test emissary/id/TikaFilePlaceTest/BMP.dat: emissary.core.blob.DataException: Data exceeds the maximum size configured for array usage, size=202074 max=3000
UnixFilePlaceTest>IdentificationTest.testIdentificationPlace:92 Cannot run test emissary/id/UnixFilePlaceTest/RICH_TEXT.dat: emissary.core.blob.DataException: Data exceeds the maximum size configured for array usage, size=25895 max=3000

adyoun2:
Oddly the ....FilePlaceTests weren't being run, are there flags I need to send to maven to cause them to be included?

AllDataContainerTest is working for me, but the failure you are seeing looks very odd. 

Get a data container with contents "Base Data"
clone it
Append " with append to the original"
Confirm the contents if both the original and the clone.

Somehow the append to the original is loosing all but the last byte before appending.

It is a parameterised test, do you have more info on which sub-class it is testing?

adyoun2:
Discovered the above failures in ResourceReaderTest were causing tests not to run. Apart from the unknown failure in AllDataContainer test for you, they should new all work.

fongrx7:
I am seeing the same failure here as well: 
 AllDataContainerTest.testClone:383 expected:<[Base Dat]a with append> but was:<[]a with append>
[ERROR]   AllDataContainerTest.testClone:383 expected:<[Base Dat]a with append> but was:<[]a with append>
[ERROR]   AllDataContainerTest.testClone:383 expected:<[Base Dat]a with append> but was:<[]a with append>
[ERROR]   AllDataContainerTest.testClone:383 expected:<[Base Dat]a with append> but was:<[]a with append>

adyoun2:
I haven't managed to reproduce these locally. If they are still occurring after the latest push, could you attach the surefire xml file for the failed test?

mjwall:
When there are inconsistent runs, please document the following
1 - OS and version
2 - Java and maven version
3 - exact command used, ie 'mvn clean test'

If there are still differences, adding -X to the maven command will dump lots of useful info.  Could be environment variable differences as well.

adyoun2:
mvn -version
Apache Maven 3.5.4 (1edded0938998edf8bf061f1ceb3cfdeccf443fe; 2018-06-17T19:33:14+01:00)
Maven home: /Users/P41793/maven/apache-maven-3.5.4
Java version: 1.8.0_181, vendor: Oracle Corporation, runtime: /Library/Java/JavaVirtualMachines/jdk1.8.0_181.jdk/Contents/Home/jre
Default locale: en_GB, platform encoding: UTF-8
OS name: "mac os x", version: "10.13.6", arch: "x86_64", family: "mac"

I've made some tweaks that, if they fix it, it will make sense that it was failing for you, but not that it was working for me!

adyoun2:
Have the changes caused it to build?

adyoun2:
Could you let me know if you are happy with this as a general solution to the issue, or whether you think a different approach would be preferable?

adyoun2:
It has been 2 months since I have heard anything on this. Is there any intent to pick it up again?

drewfarris:
@adyoun2  - yes definitely. I'm taking a look at it today.

drewfarris:
Thanks for the contribution, this looks excellent so far. I still have much to review, and I'm still building my understanding of the design, but I wanted to share a couple things to consider sooner rather than later.

1) The Emissary design tends to consider the `byte[]` in the IBDO as the primary view, but as you can see in the API, there is a mechanism for storing a number of other `byte[]`s using the alternate view mechanism. Often, the data stored in an alternate view is derived from the primary view -or- the contents of the primary view are moved to an alternate view while the primary view is replaced with the results of a processing place.  Consider updating the IBDO and related classes to use the `IDataContainer` interface for the alternate view api as well?

2) Consider breaking `LegacyContainerWrapper` in `IDataContainer` out into its own top level class. I suspect I understand why it's nested in `IDataContainer` - it is not really designed to be a 'publically' consumable API. Its presence in the `IDataContainer` makes that code a little harder to read.

3) Emissary code tends to generally avoid `@author` tags - I suspect this comes from the fact that some of the contributors are used to the Apache Software Foundation rules. Attribution can be obtained from git annotations / commit logs if we're trying to track down folks responsible for code. We'll probably end up getting rid of the other `@author` tags some time soon.

adyoun2:
1. I completely agree, the intent was to deliver iteratively, first to break the back of the primary storage mechanism, then to get the framework itself to interact with the binary using streaming mechanisms, then to repeat with views. I felt a pull request with all 3 would be a bit too large.

2. Totally happy to do that. Would you prefer I do it now or when the rest of your comments are made?

3. IDE automation, I'll remove them. Again I'll wait and bundle changes together.

adyoun2:
So for 1, this shows how views would be supported, please ignore the lack of tests/comments, as it is not production ready:

https://github.com/adyoun2/emissary/compare/master...adyoun2:Views?expand=1

adyoun2:
Having had a look at using IDataContainers for views, there is one subtle difference in their behaviour in the 2 contexts: When it it the primary byte array, the value is never null, where it can be for views, and a 0 length array is in fact assumed to mean null. This makes sense, but introduces a few corner cases.

Would you want
1. IDataContainer never has null data, and it is the presence or not of the IDataContainer on the view that signifies the view's presence. (This is what I have assumed in the current branch).
2. The primary IDataContainer of a base data object is nullable, but guarded against on the existing base data object data methods. This feels uglier, but is in some respects more consistent.
3. Similar to 1, but consider zero length arrays distinct from null. If I was designing from scratch, this would probably be my preferred option, but it would be hard to introduce now.

I've made the assumption of option 1, but am happy to be steered in another direction.

adyoun2:
The part that may have some issues is the output.filter package, as both the XmlOutputFilter and JsonOutputFilter try to serialise the entire object graph in memory. Are these classes used for all data?

adyoun2:
Given this has been tagged for 6.0.0 release, would you prefer I roll my follow-on changes into the same pull request?
I had planned to phase it as.

1. Enabling changes allowing streaming over BaseDataObjects of unlimited size (This PR)
2. Changes to make the framework use the streaming apis itself, sich as in the Pickup and Tika places.
3. Allow Views to use the same mechanisms.
4. The Framework's use of views is all streaming.

I have all 4 combined on this branch:
https://github.com/adyoun2/emissary/compare/master...adyoun2:UseStreamingViewApi?expand=1

Would you prefer to review and commit this in the stages listed, or all in one go?

jpdahlke:
Closing in favor of new effort starting with https://github.com/NationalSecurityAgency/emissary/pull/204

jpdahlke:
Closing in favor of new effort starting with https://github.com/NationalSecurityAgency/emissary/pull/204

Issue #18, opened 2018-08-06
WhatCommand -h/--header parameter doesn't work as expected.


smcgrattan:
With JCommander, it seems a boolean parameter with a default value of true must have an arity =1  specified in order to provide a false value to override the default value.

PR #19, opened 2018-08-07
18: WhatCommand -h/--header parameter doesn't work as expected. Added…

smcgrattan:
… arity = 1 to parameter definition. true or false must be specified when using parameter now.

Apparently  a JCommander boolean parameter without a value does not toggle the value. Without specifying a value, a boolean with a default value of false will toggle to true., a boolean with a default value of true will not toggle to false. In order to override a boolean with a default value of true, the parameter must be dfined with an arity = 1 and then true or false must be passed as the value.

### DIFF
--- a/src/main/java/emissary/command/WhatCommand.java
+++ b/src/main/java/emissary/command/WhatCommand.java
@@ -45,7 +45,7 @@ public class WhatCommand extends BaseCommand {
         return COMMAND_NAME;
     }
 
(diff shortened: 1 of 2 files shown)


justin-cowen:
Looks legit. 

http://jcommander.org
Section: 2.1. Boolean

justin-cowen:
In the future, please use the Github auto-close using keywords in your commit messages. 
https://help.github.com/articles/closing-issues-using-keywords/

ie. Closes #18 - then the rest of your method 

Whoever merges this please update the commit message.

PR #21, opened 2018-08-20
Closes #20 - Configure RollManager with a passed Configurator

### LINKED ISSUE #20: RollManager creation based on pass Configurator

justin-c-bt:


src/main/java/emissary/roll/RollManager.java
```diff
@@ -95,6 +103,19 @@ public void update(Observable o, Object arg) {
     }
 
     public static RollManager getManager() {
+        if (RM == null) {
```
drewfarris: Perhaps we should synchronize both `getManager()` and `getManager(Configurator)` so that it's not possible to unintentionally create two roll managers if multiple threads should attempt to call the methods simultaneously and each passes the null check.

justin-c-bt:
updated per comment @drewfarris

justin-c-bt:
updated per comment @drewfarris

drewfarris:
Merged by hand after minor update to synchronization - thanks @justin-c-bt

PR #22, opened 2018-08-20
Further nio parser improvements

drewfarris:
- avoids allocating the entire `maxChunkSize` up-front, allocates `minChunkSize` and then re-allocates as necessary to read an entire session in `NIOSessionParser`
- eliminates some of the confusion around returning null vs. throwing an exception in loadNextRegion in `NIOSessionParser`
- adds documentation to `NIOSessionParser`
- `FillingNIOParser` now properly slides through the chunk, dropping the sessions that are successfully parsed. This should keep the size of the allocated chunkBuffer in `NIOSessionParser` to a minimum.

### DIFF
--- a/src/main/java/emissary/parser/NIOSessionParser.java
+++ b/src/main/java/emissary/parser/NIOSessionParser.java
@@ -11,22 +11,38 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Provide a basic NIO based session parser.
+ * Provide a basic NIO-based session parser that reads data in chunks from the underlying channel. A chunk might have
+ * zero or more complete sessions within it. The chunk buffer will begin at minChunkSize, but grow as large as
+ * maxChunkSize in order to accomodate a complete session. Sessions larger than maxChunkSize will lead to
+ * ParserExceptions
  */
 public abstract class NIOSessionParser extends SessionParser {
     // Logger
     private final static Logger logger = LoggerFactory.getLogger(NIOSessionParser.class);
 
-    // the input channel we will read from
+    protected static final int MIN_CHUNK_SIZE_DEFAULT = 2 * 1024 * 1024; // 2Mb
+    protected static final int MAX_CHUNK_SIZE_DEFAULT = 40 * 1024 * 1024; // 40Mb
+
+    /** The data source for this parser */
     protected SeekableByteChannel channel;
 
-    // Track where we are in the overall file and in the current chunk
+    /** The start position of the current chunk relative to the data source */
     protected int chunkStart = 0;
+
+    /** The current chunk buffer */
     protected byte[] data = null;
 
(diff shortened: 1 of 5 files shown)


src/main/java/emissary/parser/NIOSessionParser.java
```diff
@@ -11,22 +11,38 @@
...
+    protected int writeOffset = 0;
+
+    /** Min chunk buffer size. */
+    protected int minChunkSize = MIN_CHUNK_SIZE_DEFAULT;
+
+    /** Max chunk to read at a time, we will never be able to read a file with a session larger than this. */
+    protected int maxChunkSize = MAX_CHUNK_SIZE_DEFAULT;
+
+    /** When we grow the chunk buffer to accomodate additional data, we will grow the buffer by this increment */
+    protected int chunkAllocationIncrement = (10 * 1024 * 1024) + 100;
```
justin-c-bt: Why 100MB + 100?

src/main/java/emissary/parser/NIOSessionParser.java
```diff
@@ -68,49 +84,76 @@ public void setChunkSize(int value) {
...
-            data = new byte[chunkSize];
+        if (data == null) {
+            logger.debug("allocating new byte[] of size {}", minChunkSize);
+            data = new byte[minChunkSize];
+        }
+
+        if (writeOffset >= data.length) {
+            // grow the byte buffer to accomodate more data
+            int newSize = data.length + chunkAllocationIncrement;
+            if (newSize > maxChunkSize) {
```
justin-c-bt: So the chuckAllocationAgreement (100MB) is already greater than the default maxChuckSize (40MB). 

This should always evaluate to true in the default configuration. Is that intentional?

src/main/java/emissary/parser/NIOSessionParser.java
```diff
@@ -68,49 +84,76 @@ public void setChunkSize(int value) {
...
+                newSize = maxChunkSize;
+            }
+
+            if (data.length >= maxChunkSize) {
```
justin-c-bt: maxChunk is 40MB. We wont process anything larger than 40MB?

src/main/java/emissary/parser/NIOSessionParser.java
drewfarris: This is sort of arbitrary, but I did not want to make it an even multiple of the default maxChunkSize, as the comments suggest - I'm shooting for something along the lines of ByteArrayOutputStream's `grow()` implementation eventually, but that will be another PR

src/main/java/emissary/parser/NIOSessionParser.java
drewfarris: 40MB is the default setting and should be managed by configuration to something larger for other use cases.

PR #24, opened 2018-10-06
Filename Tika detection and multiple Tika Signatures

mabruraas:
Added possibility to configure multiple Tika signature files, and TikaFilePlace now includes filename in media type detection.

### DIFF
--- a/src/main/java/emissary/id/TikaFilePlace.java
+++ b/src/main/java/emissary/id/TikaFilePlace.java
@@ -114,6 +114,7 @@ public class TikaFilePlace extends emissary.id.IdPlace {
     private MediaType detectType(IBaseDataObject d) throws Exception {
         Metadata metadata = new Metadata();
         InputStream input = TikaInputStream.get(d.data(), metadata);
+        metadata.set(Metadata.RESOURCE_NAME_KEY, d.getFilename());
         MediaType mediaType = mimeTypes.detect(input, metadata);
         logger.debug("Tika type: " + mediaType.toString());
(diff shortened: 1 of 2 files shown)


src/main/java/emissary/id/TikaFilePlace.java
```diff
@@ -86,23 +83,46 @@ protected void configurePlace() throws IOException {
...
      * Use the Tika mime type (magic) detector to identify the file type
-     * 
+     *
      * @param d the IBaseDataObject payload to evaluate
      * @return mediaType
      */
     private MediaType detectType(IBaseDataObject d) throws Exception {
         Metadata metadata = new Metadata();
         InputStream input = TikaInputStream.get(d.data(), metadata);
+        metadata.set(Metadata.RESOURCE_NAME_KEY, d.getFilename());
```
drewfarris: Does Tika use the filename in determining the type of a file in some cases? If so, it would be nice to be able to disable this behavior via emissary configuration. There are contexts where this behavior would not be desirable.

mabruraas: Tika applies the filename as [a hint, if provided](https://github.com/apache/tika/blob/master/tika-core/src/main/java/org/apache/tika/mime/MimeTypes.java#L500): If the filename hint is the only MimeType found or it's equal to or a sub type of, an existing hint; it will be used. If not the filename hint will be ignored. 

If you still think it's necessary, I can add an option to disable the filename check in the Emissary configuration?

PR #26, opened 2018-10-17
Added ability to reset byte data

apmoriarty:
Added ability to reset byte data operated on by ByteMatcher and KeywordScanner

### DIFF
--- a/src/main/java/emissary/util/search/ByteMatcher.java
+++ b/src/main/java/emissary/util/search/ByteMatcher.java
@@ -1,5 +1,6 @@
 package emissary.util.search;
 
+import java.nio.charset.Charset;
 
 /**
  * This class provides some simple string matching functions on byte arrays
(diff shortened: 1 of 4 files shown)


src/main/java/emissary/util/search/ByteMatcher.java
```diff
@@ -13,14 +13,30 @@
     public static final int NOTFOUND = -1;
 
 
+    public ByteMatcher(String data) {
+        this(data.getBytes());
```
adyoun2: It would be better to specify a charset, or ensure one is provided by the client

apmoriarty: Added the ability to specify a charset by name

apmoriarty:
Request that final merge is a "squash & merge", @mjwall @justin-c-bt

PR #29, opened 2018-11-14
Added default empty constructors for ByteMatcher and KeyWordScanner.

apmoriarty:


drewfarris:
Hi @apmoriarty - no apparent functional issues here, but can you provide any context as to why the default empty constructors are needed?

apmoriarty:
> Hi @apmoriarty - no apparent functional issues here, but can you provide any context as to why the default empty constructors are needed?
In the case of instantiating the matchers up front and then setting/resetting the data I found myself writing the 'new byte[0]' as a constructor arg often enough that I wanted a default empty constructor.

PR #34, opened 2019-01-28
update to use java nio

dev-mlb:
This is an old branch that I had and rebased. Most of this update is to use the updated java nio package and move away from FileInputStream. Why? -- FileInputStream/FileOutputStream can be harmful and updating to Files.new can help with I/O efficiency.

Note -- significant change to Executrix.

### DIFF
--- a/src/main/java/emissary/config/ConfigUtil.java
+++ b/src/main/java/emissary/config/ConfigUtil.java
@@ -3,14 +3,16 @@ package emissary.config;
 import java.io.BufferedOutputStream;
 import java.io.ByteArrayInputStream;
 import java.io.File;
-import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
+import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Properties;
 import java.util.Set;
@@ -404,9 +406,13 @@ public class ConfigUtil {
      */
     public static long getConfigFileLastModified(final String name) {
         final String sname = getConfigFile(name);
-        final File f = new File(sname);
-        if (f.exists() && f.canRead()) {
-            return f.lastModified();
+        final Path f = Paths.get(sname);
+        if (Files.exists(f) && Files.isReadable(f)) {
+            try {
+                return Files.getLastModifiedTime(f).toMillis();
+            } catch (IOException e) {
+                logger.warn("There was an error getting the last modified time for {}", f, e);
+            }
         }
         return -1L;
     }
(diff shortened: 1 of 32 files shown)


adyoun2:
This has implications to the use of https://github.com/NationalSecurityAgency/emissary/pull/13/, and looks good.

src/main/java/emissary/config/ConfigUtil.java
```diff
@@ -431,7 +429,7 @@ public static InputStream getConfigStream(final String name) throws IOException
         File f = new File(sname);
```
drewfarris: Consider starting with a `Path` here and then using the static `Files` api method `isReadable` with checks for both existence and readability.

src/main/java/emissary/config/ConfigUtil.java
```diff
@@ -621,29 +619,15 @@ public static InputStream getConfigData(final String f) throws IOException {
      * @throws IOException on read or write problems
      */
     public static void getConfigDataLocal(final String f, final String outputPath) throws IOException {
-        final InputStream is = getConfigData(f);
-        final BufferedOutputStream os = new BufferedOutputStream(new FileOutputStream(outputPath));
-        try {
+        try (final InputStream is = getConfigData(f);
+                final BufferedOutputStream os = new BufferedOutputStream(Files.newOutputStream(new File(outputPath).toPath()))) {
```
drewfarris: Consider replacing `new File(outputPath).toPath()` with `Paths.get(outputPath)` - its a bit more concise.

src/main/java/emissary/config/ConfigUtil.java
drewfarris: .. there are a bunch of other cases where we could work with `Path` instead of `File` objects and the `Files` api methods in this class. I won't annotate each one. I wonder if we could drop all imports of `java.io.File`.

src/test/java/emissary/place/MainTest.java
```diff
@@ -34,11 +35,9 @@
     public void setUp() throws Exception {
 
 
-        try {
-            File dataFile = new File(TMPDIR + "/testmain.dat");
-            FileOutputStream ros = new FileOutputStream(dataFile);
+        File dataFile = new File(TMPDIR + "/testmain.dat");
```
drewfarris: `Paths.get(TMPDIR, "testmain.dat");` would work well here. Yay varargs.

src/test/java/emissary/place/ServiceProviderPlaceTest.java
```diff
@@ -217,7 +218,7 @@ private void runFileConfiguredTest(boolean usePackage, int ctorType) throws Emis
             } else {
                 cfg = new File(CFGDIR + "/MyFileConfigedTestPlace" + ConfigUtil.CONFIG_FILE_ENDING);
```
drewfarris: `Paths.get(CFGDIR, "MySomething" + ConfigUtil.CONFIG_FILE_ENDING)` maybe?

src/test/java/emissary/config/ServiceConfigGuideTest.java
```diff
@@ -615,4 +611,25 @@ public void testFindStringEntry_WithDefaultAndConfiguredNull_Emissary201() {
...
+        // Write the config bytes out to a temp file
+        final String dir = System.getProperty("java.io.tmpdir");
+        final String priname = dir + "/output/testing.out";
+
+        try {
+            Executrix.writeDataToFile(new String("OUTDIRTEST = " + dir + "/copilot/../output/testing.out").getBytes(), priname);
+            ServiceConfigGuide guide = new ServiceConfigGuide(priname);
+
+            String canonical = guide.findCanonicalFileNameEntry("OUTDIRTEST", "");
+            System.out.println(canonical);
```
ets-mlb: remove println

dev-mlb:
broken up into smaller prs #102 - #106

PR #35, opened 2019-02-04
extensibility updates

ets-mlb:


adyoun2:
I slightly worry this further extends IBO being all things to all people. See https://github.com/NationalSecurityAgency/emissary/pull/13/ and https://github.com/adyoun2/emissary/compare/master...adyoun2:Views?expand=1 for examples of where I have tried to pull capability from the single class to a full object graph.

Having said that, ID certainly is something the class itself should be responsible for, so maybe the others should too.

PR #43, opened 2019-04-23
Updated WorkBundle class for TransactionManager

ssj4ndrew:
Changes include addition of array for files that failed to parse and a list of transaction IDs to prevent recalculating UUIDs.

### DIFF
--- a/src/main/java/emissary/pickup/WorkBundle.java
+++ b/src/main/java/emissary/pickup/WorkBundle.java
@@ -48,13 +48,22 @@ public class WorkBundle implements Serializable, Comparable<WorkBundle> {
     boolean simpleMode = false;
 
     // List of filenames comprising this bundle
-    List<String> fileNameList = new ArrayList<String>();
+    List<String> fileNameList = new ArrayList<>();
+
+    // List of file transaction IDs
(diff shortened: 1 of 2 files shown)


src/main/java/emissary/pickup/WorkBundle.java
```diff
@@ -496,9 +592,18 @@ private static WorkBundle buildWorkBundle(Document jdom) {
         if (serr != null && serr.length() > 0) {
             wb.setErrorCount(Integer.parseInt(serr));
         }
+        String sFilesProcessedCount = root.getChildTextTrim("filesProcessedCount");
+        if (sFilesProcessedCount != null && sFilesProcessedCount.length() > 0) {
+            wb.setErrorCount(Integer.parseInt(sFilesProcessedCount));
```
fbruton: I think this is using the wrong setter

ssj4ndrew: Thanks, updated setter.

ssj4ndrew:
Closed this PR. Added Jason's suggestion of a WorkUnit class. https://github.com/NationalSecurityAgency/emissary/pull/46

PR #54, opened 2019-07-25
Added files for creation of Docker images and Jenkins pipelines

bekelm:
Added Dockerfiles, docker-compose.yml files and Jenkins pipeline files to provide functionality for creation of Docker images and build automation using Jenkins. Docker images allow for Emissary to be run in standalone mode and in cluster mode using docker-compose. The Docker files were based on the "emissary in docker #6" PR and are used to create docker images for use in standalone and cluster mode. Two Jenkinsfiles were provided, one to process Emissary on it's own and one to process and package an Emissary JAR for use in a downstream build job.


src/test/resources/jenkins_config/settings-security.xml
```diff
@@ -0,0 +1 @@
+<settingsSecurity><master>{KlTvz8oxYrMHlSFrVYlLuyeimGK4Pp6NytqDwATt8dX87pciTZOrwqzkgHHoKlk0}</master></settingsSecurity>
```
mjwall: you shouldn't need this

src/test/resources/jenkins_config/settings.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
mjwall: nor should you need this

.gitignore
```diff
@@ -12,6 +12,9 @@ mvn_test.out
 # used when running ./emissary script
 .mvn-classpath
 
+# used when running ./Docker_compose_test_script.sh
+test_results
```
mjwall: consider moving this to target/test_results

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN chmod 766 /opt
+RUN mkdir -p /opt/emissary-${PROJ_VERS}/localoutput/json
+RUN chmod -R a+rw /opt/emissary
+
+EXPOSE 8001
+
+
+ENTRYPOINT ["./emissary"]
+
+CMD ["server", "-a", "2", "-p", "8001"]
```
mjwall: might be useful to pass in # agents, as well other options

src/test/resources/jenkins_config/settings-security.xml
bekelm: Removed as suggested.

src/test/resources/jenkins_config/settings.xml
bekelm: Removed as suggested.

.gitignore
bekelm: Fixed test_results output as suggested.

Dockerfile
bekelm: I agree it might be useful to pass in agents and/or other options.  The beauty of using the Dockerfile in this manner is that these are only the default options for running this Dockerfile. For any use case where a different number of agents or other options need to be specified when doing a "docker run emissary:latest [CMD]" or similar command.

mjwall: :+1:

.gitignore
```diff
@@ -12,6 +12,9 @@ mvn_test.out
 # used when running ./emissary script
 .mvn-classpath
 
+# used when running ./Docker_compose_test_script.sh
+target/test_results
```
mjwall: target should be gitignore already

PR #64, opened 2019-10-29
EM-4461: Fix directory name collisions

danielsaenz:
Replace use of System.currentTimeMillis() with System.nanoTime() to generate part of temporary directory names to avoid name collisions when these are generated by from different threads at about the same time.

### DIFF
--- a/src/main/java/emissary/util/io/FileManipulator.java
+++ b/src/main/java/emissary/util/io/FileManipulator.java
@@ -6,6 +6,7 @@ package emissary.util.io;
 
 import java.io.File;
 import java.io.Serializable;
+import java.security.SecureRandom;
 
 /**
  * A class of utility methods for manipulating files.
@@ -21,10 +22,14 @@ public class FileManipulator implements Serializable {
      */
     private static String FS = System.getProperty("file.separator", "/");
 
+    // Use ThreadLocal to allocate a SecureRandom instance per thread instead
+    // of Math.random() which uses the same instance of Random for all threads
+    // that causes resource contention and is not as secure.
+    private static final ThreadLocal<SecureRandom> secureRandomThreadLocal = ThreadLocal.withInitial(() -> new SecureRandom());
+
     public static String mkTempFile(final String dirPath, final String file) {
-        // Date now = new Date(System.currentTimeMillis());
-        String theFullPath = "";
-        String tempName = file + (int) (Math.rint(Math.random() * 10000));
+        String theFullPath;
+        String tempName = file + (secureRandomThreadLocal.get().nextLong() & Long.MAX_VALUE);
         if (dirPath.endsWith(FS) || dirPath.endsWith("/")) {
             theFullPath = dirPath + tempName;
         } else {
@@ -34,8 +39,7 @@ public class FileManipulator implements Serializable {
         // Check to make sure file does not already exist
         File theTempFile = new File(theFullPath);
         if (theTempFile.exists()) {
-            tempName = "temp" + (int) (Math.rint(Math.random() * 10000));
-
+            tempName = "temp" + (secureRandomThreadLocal.get().nextLong() & Long.MAX_VALUE);
             if (dirPath.endsWith("/")) {
                 theFullPath = dirPath + tempName;
             } else {
--- a/src/main/java/emissary/util/shell/Executrix.java
+++ b/src/main/java/emissary/util/shell/Executrix.java
@@ -129,7 +129,7 @@ public class Executrix {
     public String[] makeTempFilenames() {
         final String[] names = new String[7];
         final String dir = FileManipulator.mkTempFile(this.tmpDir, this.placeName);
-        final String base = Long.toString(System.currentTimeMillis());
+        final String base = Long.toString(System.nanoTime());
         names[DIR] = dir;
         names[BASE] = base;
         names[BASE_PATH] = dir + File.separator + base;
--- a/src/test/java/emissary/util/shell/ExecutrixTest.java
+++ b/src/test/java/emissary/util/shell/ExecutrixTest.java
@@ -15,8 +15,12 @@ import java.io.File;
 import java.io.RandomAccessFile;
 import java.nio.channels.FileChannel;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.IntStream;
 
 import emissary.test.core.UnitTest;
 import org.junit.After;
@@ -78,6 +82,26 @@ public class ExecutrixTest extends UnitTest {
         assertNotNull("names array has null OUTPATH", names[Executrix.OUTPATH]);
         assertTrue("names must use out file ending", names[Executrix.OUT].endsWith(".out"));
         assertTrue("names must use in file ending", names[Executrix.IN].endsWith(".in"));
+    }
+
+    @Test
+    public void testExecutrixUniqueBase() {
+        this.e.setInFileEnding(".in");
+        this.e.setOutFileEnding(".out");
+        this.e.setOrder("NORMAL");
+
+        final int COUNT = 1000;
+        final Set<String> basePathSet = Collections.synchronizedSet(new HashSet<>(COUNT));
+
+        // Generate COUNT sets of names
+        IntStream.range(0, COUNT).parallel().forEach(number -> {
+            final String[] name = this.e.makeTempFilenames();
+            assertNotNull("name null DIR", name[Executrix.DIR]);
+            assertNotNull("name null BASE", name[Executrix.BASE]);
+            assertNotNull("name null BASE_PATH", name[Executrix.BASE_PATH]);
+            basePathSet.add(name[Executrix.BASE_PATH]);
+        });
+        assertEquals("Some BASE_PATH entries mismatch", COUNT, basePathSet.size());
     }
 
     @Test


src/test/java/emissary/util/shell/ExecutrixTest.java
```diff
@@ -80,6 +81,22 @@ public void testExecutrixParams() {
         assertTrue("names must use in file ending", names[Executrix.IN].endsWith(".in"));
     }
 
+    @Test
+    public void testExecutrixUniqueBase() {
+        this.e.setInFileEnding(".in");
+        this.e.setOutFileEnding(".out");
+        this.e.setOrder("NORMAL");
+
+        final String[] names1 = this.e.makeTempFilenames();
```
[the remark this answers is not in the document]
danielsaenz: Changed to iterate over 100 instances.

src/main/java/emissary/util/shell/Executrix.java
```diff
@@ -129,7 +129,7 @@ protected void configure(final Configurator configGArg) {
     public String[] makeTempFilenames() {
         final String[] names = new String[7];
         final String dir = FileManipulator.mkTempFile(this.tmpDir, this.placeName);
```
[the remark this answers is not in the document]
danielsaenz: The use of Math.random() used to build the Executrix DIR string was changed to use ThreadLocalRandom to ensure that each thread has a unique instance of Random and avoid collisions.

src/test/java/emissary/util/shell/ExecutrixTest.java
```diff
@@ -80,6 +82,30 @@ public void testExecutrixParams() {
...
+            assertNotNull("names array has null BASE", names[number][Executrix.BASE]);
+            assertNotNull("names array has null BASE_PATH", names[number][Executrix.BASE_PATH]);
+        });
+
+        // Verify that the names generated do not match the next one. This does
+        // not necessarily need to be done in parallel.
+        for (int i = 0; i < 99; i++) {
+            assertNotEquals("DIR matches", names[i + 1][Executrix.DIR], names[i][Executrix.DIR]);
+            assertNotEquals("BASE matches", names[i + 1][Executrix.BASE], names[i][Executrix.BASE]);
+            assertNotEquals("BASE_PATH matches", names[i + 1][Executrix.BASE_PATH], names[i][Executrix.BASE_PATH]);
```
fbruton: Recommend adding ```names[][Executrix.BASE_PATH]``` to a Set and verify that the Set size matches the number of expected unique file paths.

manester:
Just read the following on the performance diff between System.currentTimeMillis() and System.nanoTime(). We should check the performance diff on the target OS if we think this could be an issue for us.

https://stackoverflow.com/questions/19052316/why-is-system-nanotime-way-slower-in-performance-than-system-currenttimemill

src/test/java/emissary/util/shell/ExecutrixTest.java
danielsaenz: Updated testcase to validate unique strings using Sets.

danielsaenz:
> 
> 
> Just read the following on the performance diff between System.currentTimeMillis() and System.nanoTime(). We should check the performance diff on the target OS if we think this could be an issue for us.
> 
> https://stackoverflow.com/questions/19052316/why-is-system-nanotime-way-slower-in-performance-than-system-currenttimemill

It's true. A simple multi-threaded test on a Mac Pro using some of the code from the post showed that it takes on average 2 times longer to run System.nanoTime() than System.currentTimeMillis(). On a similar note, the same multi-threaded test on a Mac Pro showed that using ThreadLocalRandom.current().nextLong() is on average 3 times faster than System.currentTimeMillis(), it is unique, but doesn't follow an increasing pattern as time. 

The fixes included in this pull-request aim to make both DIR and BASE unique, which are used to build the BASE_PATH (i.e. BASE_PATH = DIR + BASE). It may be possible to continue using System.currentTimeMillis() to generate the BASE portion of the BASE_PATH in order to minimize the processing time impact, as long as its acceptable to have DIR be the only unique portion of the BASE_PATH since BASE may not be unique.

src/test/java/emissary/util/shell/ExecutrixTest.java
```diff
@@ -80,6 +82,30 @@ public void testExecutrixParams() {
...
+            names[number] = this.e.makeTempFilenames();
+            assertNotNull("names array has null DIR", names[number][Executrix.DIR]);
+            assertNotNull("names array has null BASE", names[number][Executrix.BASE]);
+            assertNotNull("names array has null BASE_PATH", names[number][Executrix.BASE_PATH]);
+        });
+
+        // Verify that the names generated do not match the next one. This does
+        // not necessarily need to be done in parallel.
+        for (int i = 0; i < 99; i++) {
+            assertNotEquals("DIR matches", names[i + 1][Executrix.DIR], names[i][Executrix.DIR]);
```
fbruton: Gettting the following error:
```ExecutrixTest.testExecutrixUniqueBase:108 Total DIR entries missmatch expected:<100> but was:<99>```

File: src/test/java/emissary/util/shell/ExecutrixTest.java [hunk 5]
```diff
@@ -80,6 +83,33 @@ public void testExecutrixParams() {
...
+        IntStream.range(0, COUNT).parallel().forEach(number -> {
+            names[number] = this.e.makeTempFilenames();
+            assertNotNull("names array has null DIR", names[number][Executrix.DIR]);
+            dirSet.add(names[number][Executrix.DIR]);
+            assertNotNull("names array has null BASE", names[number][Executrix.BASE]);
+            baseSet.add(names[number][Executrix.BASE]);
+            assertNotNull("names array has null BASE_PATH", names[number][Executrix.BASE_PATH]);
+            basePathSet.add(names[number][Executrix.BASE_PATH]);
+        });
+        assertEquals("Total DIR entries missmatch", COUNT, dirSet.size());
```
fbruton: Typo: missmatch

src/test/java/emissary/util/shell/ExecutrixTest.java
```diff
@@ -80,6 +83,29 @@ public void testExecutrixParams() {
...
+        this.e.setInFileEnding(".in");
+        this.e.setOutFileEnding(".out");
+        this.e.setOrder("NORMAL");
+
+        final int COUNT = 100;
+        final Set<String> baseSet = new HashSet<>();
+        final Set<String> basePathSet = new HashSet<>();
+
+        // Generate 100 sets of names
+        IntStream.range(0, COUNT).parallel().forEach(number -> {
```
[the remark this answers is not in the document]
danielsaenz: Issue has been addressed, thanks.

src/test/java/emissary/util/shell/ExecutrixTest.java
```diff
@@ -80,6 +83,31 @@ public void testExecutrixParams() {
...
+        final Set<String> concurrentBaseSet = concurrentBaseMap.newKeySet();
+        final ConcurrentHashMap<String, Integer> concurrentBasePathMap = new ConcurrentHashMap<>();
+        final Set<String> concurrentBasePathSet = concurrentBasePathMap.newKeySet();
+
+        // Generate 100 sets of names
+        IntStream.range(0, COUNT).parallel().forEach(number -> {
+            final String[] name = this.e.makeTempFilenames();
+            assertNotNull("name null DIR", name[Executrix.DIR]);
+            assertNotNull("name null BASE", name[Executrix.BASE]);
+            assertTrue(concurrentBaseSet.add(name[Executrix.BASE]));
```
fbruton: testExecutrixUniqueBase(emissary.util.shell.ExecutrixTest)  Time elapsed: 0.088 s  <<< FAILURE!
java.lang.AssertionError
	at emissary.util.shell.ExecutrixTest.testExecutrixUniqueBase(ExecutrixTest.java:99)
Caused by: java.lang.AssertionError
	at emissary.util.shell.ExecutrixTest.lambda$testExecutrixUniqueBase$0(ExecutrixTest.java:103)

src/test/java/emissary/util/shell/ExecutrixTest.java
```diff
@@ -80,6 +83,27 @@ public void testExecutrixParams() {
...
 
+    @Test
+    public void testExecutrixUniqueBase() {
+        this.e.setInFileEnding(".in");
+        this.e.setOutFileEnding(".out");
+        this.e.setOrder("NORMAL");
+
+        final int COUNT = 1000;
+        final ConcurrentHashMap<String, Integer> concurrentBasePathMap = new ConcurrentHashMap<>();
+        final Set<String> concurrentBasePathSet = concurrentBasePathMap.newKeySet();
```
fbruton: Recommend replacing with:
```Collections.synchronizedSet(new HashSet<>(COUNT))```

File: src/test/java/emissary/util/shell/ExecutrixTest.java [hunk 9]
```diff
@@ -80,6 +83,27 @@ public void testExecutrixParams() {
...
+
+        // Generate 100 sets of names
+        IntStream.range(0, COUNT).parallel().forEach(number -> {
+            final String[] name = this.e.makeTempFilenames();
+            assertNotNull("name null DIR", name[Executrix.DIR]);
+            assertNotNull("name null BASE", name[Executrix.BASE]);
+            assertNotNull("name null BASE_PATH", name[Executrix.BASE_PATH]);
+            assertTrue(concurrentBasePathSet.add(name[Executrix.BASE_PATH]));
+        });
+        assertEquals("Some BASE_PATH entries miss match", COUNT, concurrentBasePathSet.size());
```
fbruton: Typo: Update to mismatch.

src/test/java/emissary/util/shell/ExecutrixTest.java
```diff
@@ -80,6 +83,27 @@ public void testExecutrixParams() {
...
+        final ConcurrentHashMap<String, Integer> concurrentBasePathMap = new ConcurrentHashMap<>();
+        final Set<String> concurrentBasePathSet = concurrentBasePathMap.newKeySet();
+
+        // Generate 100 sets of names
+        IntStream.range(0, COUNT).parallel().forEach(number -> {
+            final String[] name = this.e.makeTempFilenames();
+            assertNotNull("name null DIR", name[Executrix.DIR]);
+            assertNotNull("name null BASE", name[Executrix.BASE]);
+            assertNotNull("name null BASE_PATH", name[Executrix.BASE_PATH]);
+            assertTrue(concurrentBasePathSet.add(name[Executrix.BASE_PATH]));
```
fbruton: Remove this assert statement. Will be handled on line 104.

src/test/java/emissary/util/shell/ExecutrixTest.java
danielsaenz: Great, that one looks better. Done.

src/test/java/emissary/util/shell/ExecutrixTest.java
danielsaenz: Done.

src/test/java/emissary/util/shell/ExecutrixTest.java
danielsaenz: Done.

src/main/java/emissary/util/io/FileManipulator.java
```diff
@@ -21,10 +22,14 @@
...
+    // of Math.random() which uses the same instance of Random for all threads
+    // that causes resource contention and is not as secure.
+    private static final ThreadLocal<SecureRandom> secureRandomThreadLocal = ThreadLocal.withInitial(() -> new SecureRandom());
+
     public static String mkTempFile(final String dirPath, final String file) {
-        // Date now = new Date(System.currentTimeMillis());
-        String theFullPath = "";
-        String tempName = file + (int) (Math.rint(Math.random() * 10000));
+        String theFullPath;
+        String tempName = file + secureRandomThreadLocal.get().nextInt(10000);
```
fbruton: Replace with ```nextLong()```

src/main/java/emissary/util/io/FileManipulator.java
```diff
@@ -34,8 +39,7 @@ public static String mkTempFile(final String dirPath, final String file) {
         // Check to make sure file does not already exist
         File theTempFile = new File(theFullPath);
         if (theTempFile.exists()) {
-            tempName = "temp" + (int) (Math.rint(Math.random() * 10000));
-
+            tempName = "temp" + secureRandomThreadLocal.get().nextInt(10000);
```
fbruton: Replace with ```nextLong()```

src/main/java/emissary/util/io/FileManipulator.java
danielsaenz: Done. Just had to mask the first bit to keep the number positive.

src/main/java/emissary/util/io/FileManipulator.java
danielsaenz: Done.

PR #76, opened 2020-02-05
Emissary shutdown updates

dev-mlb:


src/main/java/emissary/pickup/IPickUp.java
```diff
@@ -0,0 +1,12 @@
+package emissary.pickup;
+
+import emissary.util.ClassComparator;
+
+public interface IPickUp {
+
+    void shutDown();
+
+    static boolean isImplementation(Class<? extends Object> clazz) {
```
jpdahlke: I think we could get around this by just using isAssignableFrom when necessary and keep this clean.

PR #87, opened 2020-10-01
Resource Watcher Fix

LightSeekerSC:
Currently the reset deletes the namespace and any active timers will not report once they complete. The new system deletes saved values while preserving the namespace.

### DIFF
--- a/src/main/java/emissary/core/ResourceWatcher.java
+++ b/src/main/java/emissary/core/ResourceWatcher.java
@@ -1,5 +1,6 @@
 package emissary.core;
 
+import java.lang.reflect.Field;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Queue;
@@ -8,6 +9,8 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 
+import com.codahale.metrics.ExponentiallyDecayingReservoir;
+import com.codahale.metrics.Histogram;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import emissary.place.IServiceProviderPlace;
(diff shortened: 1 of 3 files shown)


fbruton (changes requested):
Change logStats to use mean instead of median.

fbruton:
Should we consider updating to 4.1.2?

src/main/java/emissary/core/ResourceWatcher.java
```diff
@@ -148,15 +151,23 @@ public void run() {
...
-        for (final Map.Entry<String, Timer> e : this.metrics.getTimers().entrySet()) {
-            this.metrics.remove(e.getKey());
+        // We use reflection to reset the histograms that track finished events, but leaves the namespace for active timers
+        for (Timer timer : this.metrics.getTimers().values()) {
+            try {
+                Field histogramField = Timer.class.getDeclaredField("histogram");
+                histogramField.setAccessible(true);
+                histogramField.set(timer, new Histogram(new ExponentiallyDecayingReservoir()));
+            } catch (NoSuchFieldException | IllegalAccessException e) {
+                e.printStackTrace();
```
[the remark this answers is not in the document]
fbruton: Remove ```printStackTrace()``` call

fbruton (changes requested):

fbruton (approved):

LightSeekerSC:
> Should we consider updating to 4.1.2?

Some of the sub-projects we use are not accommodating of this upgrade

drewfarris (approved):

LightSeekerSC:
I believe this PR is ready for merge. Please let me know if there are any outstanding issues preventing a merge

src/main/java/emissary/core/ResourceWatcher.java
```diff
@@ -148,15 +151,23 @@ public void run() {
...
-        // no reset but remove:
-        // https://github.com/codahale/metrics/issues/399
-        for (final Map.Entry<String, Timer> e : this.metrics.getTimers().entrySet()) {
-            this.metrics.remove(e.getKey());
+        // We use reflection to reset the histograms that track finished events, but leaves the namespace for active timers
+        for (Timer timer : this.metrics.getTimers().values()) {
+            try {
+                Field histogramField = Timer.class.getDeclaredField("histogram");
+                histogramField.setAccessible(true);
+                histogramField.set(timer, new Histogram(new ExponentiallyDecayingReservoir()));
```
dev-mlb, line 167: Is reflection the only way to do this?  Is there anything in the new version that fixes what we are trying to solve? Is there a different library that we could use? I'm just wondering if we are using the library is a way that it wasn't intended, so is there another option that would be better long term? Also, I'm afraid of a change like this affecting us being able to upgrade to a new version in the future.

LightSeekerSC: @dev-mlb 
I have done a second due-dillegence check. The newest version of this library has the same limitations and there are no ways other than reflection to accomplish the desired effect.

There is a rolling-metric library developed by a third party (https://github.com/vladimir-bukhtoyarov/rolling-metrics) that would allow resetting histograms when we perform the getSnapshot() in the MetricsFormatter. However, this solution adds two additional dependencies to our project and it is unclear what future maintenance for these libraries will look like.

Histograms as part of the timers has been a core part of the library infrastructure since inception. This structure would be seemingly unlikely to change and shouldn't prevent an upgrade. Any upgrade issues could be easily resolved by updating the reflection usage or implementing any new functions that allow the desired behavior.

I added a new line to the ResourceWatcherTest to ensure that the reset occurs as expected.

I recommend proceeding with the current changes. Any alternative solutions will be time-intensive to implement and in my opinion incur just as much if not more risk and challenge for future upgrades.

LightSeekerSC:
Any updates on taking this ticket?

jpdahlke:
https://github.com/NationalSecurityAgency/burrito-grande/issues/975

PR #89, opened 2020-10-21
Integrate 'TheFramework' to handle data larger than what can fit into a Java Array

jdcove2:
Emissary currently has a couple of data limitations by virtue of storing data in Java arrays:

1. Maximum byte length of data is 2^31-1.
2. Data is stored in heap memory.

It is desired to allow Emissary to overcome these limitations, which this Pull Request proposes to do based on the Open Source library https://sourceforge.net/projects/theframework/.


jdcove2:
Since a Java byte array is no longer the type used to store the data or alternate views, code that relies on references to the byte arrays will fail. Examples of this in Burrito-Grande are:

com.burrito.util.AttachmentListTest.testAdd(...)
com.burrito.util.AttachmentListTest.testAddWithIndex(...)

jdcove2:
Having direct access to the alternate views data structure via the BaseDataObject.getAlternateViews() method is poor design. This is especially true now as what is returned is a copy of the alternate views data structure. Code that relies on manipulating the alternate views data structure directly from outside of the BaseDataObject will fail. An example of this in Burrito-Grande is:

com.burrito.extract.Utf8TokenizerPlace.removeRedundentView(...)

drewfarris:
I spent a small amount of time looking at `TheFramework` library referenced in this pull request. Interesting stuff! One non-technical thing we will need to resolve in order to move forward on this is a license related issue. 

`TheFramework` is licensed under LGPLv2 which is incompatible with the Apache Public License under which emissary is licensed. 

@jdcove2 - as you are one of the authors of `TheFramework` is there a path forward here? Would the copyright holders be willing to relicense `TheFramework` under the APL?

jdcove2:
There does not appear to be any interest in this right now, so I am closing it.

PR #90, opened 2020-10-21
IBDO Data Represented as String

LightSeekerSC:
A function for controlling how the IBDO data byte array is turned into a string and maintaining a cached version as long as the IBDO payload doesn't change

### DIFF
--- a/src/main/java/emissary/core/BaseDataObject.java
+++ b/src/main/java/emissary/core/BaseDataObject.java
@@ -4,6 +4,7 @@ import java.io.Serializable;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.rmi.Remote;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -36,6 +37,11 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
 
     /* Our payload */
     protected byte[] theData;
+
+    /**
+     * A string representation of the data
+     */
+    protected String dataString = null;
 
     /**
      * Original name of the input data. Can only be set in the constructor of the DataObject. returned via the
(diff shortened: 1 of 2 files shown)


LightSeekerSC:
Any thoughts on this PR?

LightSeekerSC:
When someone has the chance can you please review this PR?

fbruton:
I think this needs to be coordinated with other potential BaseDataObject data storage updates, such as PR-90. Also, were you able to do any profiling with metrics on the potential benefits of this change.

src/main/java/emissary/core/BaseDataObject.java
```diff
@@ -37,6 +38,11 @@
     /* Our payload */
     protected byte[] theData;
 
+    /**
+     * A string representation of the data
+     */
+    protected String dataString = null;
```
drewfarris, line 44: Although we don't serialize this, it might be reasonable to mark this `transient` as a indication to a fellow developer that this is expected to be a 'cached' representation of the data as a string.

src/main/java/emissary/core/BaseDataObject.java
```diff
@@ -273,6 +279,20 @@ public void setFilename(final String f) {
...
 
+    /**
+     * Return a String version of the data. If there is already a cached version of the dataString, then return that,
+     * otherwise generate the String to be returned and cache the result
+     *
+     * @return a String representation of the data
+     */
+    @Override
+    public String dataAsString() {
+        if (this.dataString == null) {
```
drewfarris, line 290: We might consider caching only if the string is under a certain size. Otherwise, if the string is too large, we always re-generate the string. This may mitigate some of the concerns over IBDO bloat.

src/main/java/emissary/core/BaseDataObject.java
```diff
@@ -273,6 +279,20 @@ public void setFilename(final String f) {
...
+            this.dataString = new String(this.theData, StandardCharsets.UTF_8);
```
drewfarris, line 291: It is safe to assume that the bytes will always be UTF-8? There may be cases where the data in byte[] is a non UTF-8 string or not even a string. It would be nice to provide a mechanism to pass a character encoding along in `setData()` if the caller know what that is -  or even provide an alternate to `setData()` that takes a `String` as a parameter.

jdcove2:
Currently, the IBaseDataObject interface and corresponding BaseDataObject class are both massive. Since IBaseDataObject specifies a mutable data container, I would suggest keeping this capability simple and just add "public void setDataAsString(String string)" and "public String getDataAsString()" methods that just get/set a simple private String attribute. This capability will get complicated once things are added like different ways to convert byte[] to String, determining whether the String should actually be saved or recreated based on size, adding some sort of "configuration" to the BDO so that the size can be configurable, etc. I think all these "complications" would be best placed outside the BDO.

src/main/java/emissary/core/BaseDataObject.java
jdcove2: Additionally, there should probably be a null check on "this.theData".

LightSeekerSC:
This ticket has been deemed a low-priority and is being closed for the time being to focus developer resources on higher priority efforts. If this ticket is ever re-opened the work done and comments for this PR can be used to gain a head-start on continued development

PR #96, opened 2020-11-13
Make String Match Map Deterministic For Iteration

LightSeekerSC:
The current string match map has non-deterministic behavior when iterating. With minimal added overhead we can make this map into a linked hash map which will make iteration consistent and allow order to matter for our string match map configurations.


jpdahlke:
handled under https://github.com/NationalSecurityAgency/emissary/pull/113

PR #134, opened 2021-04-12
Improve performance of emissary.util.Hexl.toUnformattedHexString()

jdcove2:


src/main/java/emissary/util/Hexl.java
```diff
@@ -14,24 +14,37 @@
...
+            }
+        }
+    }
+
     /**
      * Unformatted hex string
      */
     public static String toUnformattedHexString(byte[] data) {
-        StringBuilder output = new StringBuilder(2048);
+        final char[] characters = new char[data.length * 2];
```
fbruton, line 37: Recommend comparing performance against ```org.apache.commons.codec.binary.Hex.encodeHexString```. If the performance differences are negligible, I recommend going with the Apache commons implementation.

fbruton (changes requested):
Request change based on previous comment.

src/main/java/emissary/util/Hexl.java
jdcove2: The new code and the encodeHexString method are practically the same so their performance should be practically the same; therefore, I have replaced the new code with the encodeHexString method.

fbruton (approved):

jdcove2:
The following screenshots show the number of times and memory allocated for processing a single 20Meg mysqldump file. The first screenshot shows the number & memory created by the original code. The second screenshot shows the number & memory created by the new code.

image

image

jpdahlke (commented):
LGTM, but....

Did we want to take this one step further? If we are just going to call the apache commons library, can we just update all references to this method with a call the apache commons version? It appears to only be a handful. It's a breaking change but we haven't been very good about maintaining that in v6 so far....

Also, this puts the usage of the class and its methods on the brink. I would recommend deprecating this class in the pr at the least and then we can consider removal in a follow-up.

jdcove2:
The code cleanup suggestion will be performed in a subsequent PR.

jpdahlke:
https://github.com/NationalSecurityAgency/burrito-grande/issues/1463

jpdahlke (approved):

PR #135, opened 2021-04-12
Improve performance of emissary.util.ShortNameComparator.compare().

jdcove2:


jdcove2:
The following screenshots are from a 20Meg mysqldump input file. The original ShortNameComparator creates 4433070 objects occupying 184415712 bytes. The new ShortNameComparator does not create any heap objects.

1. Overall profiling results with original ShortNameComparator

image

2. Strings created by original ShortNameComparator

image

3. ArrayLists created by original ShortNameComparator

image

4. Patterns created by original ShortNameComparator

image

5. Matchers created by original ShortNameComparator

image

6. Overall profiling results with new ShortNameComparator

image

7. No strings created by new ShortNameComparator

image

8. No ArrayLists created by new ShortNameComparator

image

9. No Patterns created by new ShortNameComparator

image

10. No Matchers created by new ShortNameComparator

image

cfkoehler (approved):
Thank you for adding the additional unit tests.

src/main/java/emissary/util/ShortNameComparator.java
```diff
@@ -19,34 +19,115 @@
...
+                index2 = nextIndex2;
+
                 continue;
             }
 
             try {
-                int i1 = Integer.parseInt(p1[i]);
-                int i2 = Integer.parseInt(p2[i]);
-                return i1 - i2;
+                int int1 = parseInt(s1, 10, index1, length1);
```
LightSeekerSC, line 46: Comment akin to: Do a number comparison

src/main/java/emissary/util/ShortNameComparator.java
```diff
@@ -19,34 +19,115 @@
...
+                int int2 = parseInt(s2, 10, index2, length2);
+
+                return int1 - int2;
             } catch (NumberFormatException e) {
-                return p1[i].compareTo(p2[i]);
-            }
+                final String substring1 = s1.substring(index1, index1 + length1);
```
LightSeekerSC, line 51: Comment something akin to: "If numeric comparison fails, perform a string comparison"

src/main/java/emissary/util/ShortNameComparator.java
```diff
@@ -19,34 +19,115 @@
...
+        int index1 = s1.indexOf(Family.SEP, 0);
+        int index2 = s2.indexOf(Family.SEP, 0);
 
-        String[] p1 = s1.split(Family.SEP);
-        String[] p2 = s2.split(Family.SEP);
-
-        for (int i = 1; i < p1.length; i++) {
-
-            if (i >= p2.length) {
+        while (index1 >= 0) {
```
LightSeekerSC, line 25: Comment akin to: Loop until we hit the base of the descendant tree

src/main/java/emissary/util/ShortNameComparator.java
```diff
@@ -19,34 +19,115 @@
...
+            if (index1 >= 0 && index2 < 0) {
```
LightSeekerSC, line 26: The index1 check is redundant and can be removed (this is checked by the while loop)

Comment: If string 1 has children and string 2 doesn't have children

src/main/java/emissary/util/ShortNameComparator.java
```diff
@@ -19,34 +19,115 @@
...
                 return 1;
```
LightSeekerSC, line 27: Comment akin to: Return that 2 wins the comparison

src/main/java/emissary/util/ShortNameComparator.java
```diff
@@ -19,34 +19,115 @@
...
-            }
+                final String substring1 = s1.substring(index1, index1 + length1);
+                final String substring2 = s2.substring(index2, index2 + length2);
 
+                return substring1.compareTo(substring2);
+            }
         }
 
-        if (p2.length > p1.length) {
+        if (index1 < 0 && index2 >= 0) {
```
LightSeekerSC, line 58: Comment akin to: If string 1 has no children and string 2 has remaining descendants

src/main/java/emissary/util/ShortNameComparator.java
```diff
@@ -19,34 +19,115 @@
...
             return -1;
```
LightSeekerSC, line 59: Comment akin to: String 1 wins the comparison

src/main/java/emissary/util/ShortNameComparator.java
```diff
@@ -19,34 +19,115 @@
...
         }
+
         return 0;
```
LightSeekerSC, line 62: Comment akin to: Return that the values are equivalent

src/main/java/emissary/util/ShortNameComparator.java
```diff
@@ -19,34 +19,115 @@
...
+        if (radix < Character.MIN_RADIX) {
+            throw new NumberFormatException("radix " + radix + " less than Character.MIN_RADIX");
+        }
+
+        if (radix > Character.MAX_RADIX) {
+            throw new NumberFormatException("radix " + radix + " greater than Character.MAX_RADIX");
+        }
+
+        if (start < 0) {
+            throw new IllegalArgumentException("start < 0 not allowed!");
```
LightSeekerSC, line 81: Keep exceptions consistent with above: (1) change "<" to "less than" (2) remove exclamation point

src/main/java/emissary/util/ShortNameComparator.java
```diff
@@ -19,34 +19,115 @@
...
+        }
+
+        if (length < 0) {
+            throw new IllegalArgumentException("length < 0 not allowed!");
```
LightSeekerSC, line 85: Keep exceptions consistent with above: (1) change "<" to "less than" (2) remove exclamation point

src/main/java/emissary/util/ShortNameComparator.java
```diff
@@ -19,34 +19,115 @@
...
+        }
+
+        if (start + length > s.length()) {
+            throw new IllegalArgumentException("start + length > s.length() not allowed!");
```
LightSeekerSC, line 89: Keep exceptions consistent with above: (1) change ">" to "greater than" (2) remove exclamation point

src/main/java/emissary/util/ShortNameComparator.java
```diff
@@ -19,34 +19,115 @@
...
 
-        if (p2.length > p1.length) {
+        if (index1 < 0 && index2 >= 0) {
             return -1;
         }
+
         return 0;
     }
 
+    // Based on java.lang.Integer.parseInt()
```
LightSeekerSC, line 65: This doesn't belong in this class. This should be in another generic utility class that can be more broadly used across the Emissary codebase. Could add this to StringUtil or create a new IntUtil

LightSeekerSC (changes requested):

src/main/java/emissary/util/ShortNameComparator.java
```diff
@@ -19,34 +19,115 @@
     public int compare(IBaseDataObject obj1, IBaseDataObject obj2) {
         String s1 = obj1.shortName();
         String s2 = obj2.shortName();
```
drivenflywheel, line 21: 1. Does this new implementation return the same result as before if neither s1 nor s2 contain Family.SEP?
2. Does this new implementation return the same result as before if Family.SEP and s2 is "s1 + Family.SEP + [more]"?
3. Does this new implementation return the same result as before if Family.SEP and s1 is "s2 + Family.SEP + [more]"? 
4. Do we have test coverage for the above 3 scenarios to insure parity with previous behavior?

src/test/java/emissary/util/ShortNameComparatorTest.java
```diff
@@ -37,6 +37,15 @@ public void testNonsenseNames() {
         assertEquals("Bad components sort in alpha order", "foo-att-abc", l.get(0).shortName());
     }
 
+    @Test
+    public void testNonsenseNames2() {
```
drivenflywheel, line 41: "Nonsense" is subjective; this test might be better named "testComparisonOfShortNamesThatHaveNoChildren" or a shorter variant

src/test/java/emissary/util/ShortNameComparatorTest.java
```diff
@@ -37,6 +37,15 @@ public void testNonsenseNames() {
...
+        final List<IBaseDataObject> l = new ArrayList<IBaseDataObject>();
+        l.add(DataObjectFactory.getInstance(new Object[] {null, "def"}));
+        l.add(DataObjectFactory.getInstance(new Object[] {null, "abc"}));
+        Collections.sort(l, new ShortNameComparator());
+        assertEquals("Bad components sort in alpha order", "def", l.get(0).shortName());
+    }
+
     private void fillList(final List<IBaseDataObject> l) {
         l.add(DataObjectFactory.getInstance(new Object[] {this.nobytes, this.ba + "1"}));
```
drivenflywheel, line 50: All of the items currently being added to this list contain `Family.SEP`.  Would it be useful to add at least 2 entries for which the shortname does not contain `Family.SEP` so we can assert the correct behavior when comparing those to each other and to the items that contain `Family.SEP`?

src/test/java/emissary/util/ShortNameComparatorTest.java
```diff
@@ -37,6 +37,15 @@ public void testNonsenseNames() {
         assertEquals("Bad components sort in alpha order", "foo-att-abc", l.get(0).shortName());
     }
```
drivenflywheel, line 38: This comment is separate from the changes in this PR, but `testOrdering` and `testImplComparator` appear to be identical tests.

drivenflywheel (changes requested):

jdcove2:
I believe I have made all of the suggested changes. Please let me know if I missed anything.

There is also 100% code coverage by the unit tests as shown in the below screen shots.

image

image

src/test/java/emissary/util/IntegerUtilTest.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Test
+    public void testParseIntLeadingPlus() {
+        assertEquals("Failed leading plus!", 1, IntegerUtil.parseInt("+1", 10, 0, 2));
+    }
+
+    @Test
+    public void testParseIntLeadingMinus() {
+        assertEquals("Failed leading plus!", -1, IntegerUtil.parseInt("-1", 10, 0, 2));
```
drivenflywheel, line 39:
```suggestion
        assertEquals("Failed leading minus!", -1, IntegerUtil.parseInt("-1", 10, 0, 2));
```

src/test/java/emissary/util/ShortNameComparatorTest.java
```diff
@@ -37,13 +37,23 @@ public void testNonsenseNames() {
...
 
+    @Test
+    public void testNoChildren() {
```
drivenflywheel, line 41: Since `ShortNameComparator` is only used to sort items from a single family tree, this test is probably unnecessary, since a family tree can have a single one item that is not a child.

If we want to assert specific behavior, the message in case of failed assertion might be more accurate as "IBDOs that are not children sort in insertion order"

src/main/java/emissary/util/IntegerUtil.java
```diff
@@ -0,0 +1,74 @@
+package emissary.util;
+
+public class IntegerUtil {
+    private IntegerUtil() {}
+
+    // Based on java.lang.Integer.parseInt()
```
drivenflywheel, line 6: This method should have a Javadoc comment explaining how it differs from `Integer.parseInt(String)`  (i.e., its support for substrings).

src/main/java/emissary/util/ShortNameComparator.java
```diff
@@ -19,34 +19,55 @@
...
+            }
         }
 
-        if (p2.length > p1.length) {
+        // If obj1's shortName has fewer levels than obj2's shortName then obj1 < obj2.
+        if (index2 >= 0) {
             return -1;
         }
+
+        // Otherwise, obj1 == obj2
```
drivenflywheel, line 70:
```suggestion
        // Otherwise, obj1 and obj2 compare as equivalent
```

src/test/java/emissary/util/IntegerUtilTest.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Test(expected = NumberFormatException.class)
+    public void testParseIntRadixTooHigh() {
+        IntegerUtil.parseInt("0", Character.MAX_RADIX + 1, 0, 1);
+    }
+
+    @Test
+    public void testParseIntLeadingPlus() {
+        assertEquals("Failed leading plus!", 1, IntegerUtil.parseInt("+1", 10, 0, 2));
```
LightSeekerSC, line 34: Personally not a fan of exclamation marks, but that is not a deal breaker :)

src/test/java/emissary/util/IntegerUtilTest.java
```diff
@@ -0,0 +1,81 @@
...
+    public void testExceedLimitSmall() {
+        final String s = BigInteger.valueOf(2).pow(31).add(BigInteger.valueOf(1)).toString(10);
+
+        IntegerUtil.parseInt(s, 10, 0, s.length());
+    }
+
+
+    @Test(expected = NumberFormatException.class)
+    public void testExceedLimitBig() {
+        final String s = BigInteger.valueOf(2).pow(31).add(BigInteger.valueOf(11)).toString(10);
```
LightSeekerSC, line 77: Not sure I understand how this differs from the previous test testExceedLimitSmall

LightSeekerSC (approved):
All my issues have been addressed. I have a question and a preferred change, but nothing prohibiting this from passing. I approve.

fbruton (changes requested):
Recommend waiting until Emissary is updated to Java 9 or later and making the following API call:
https://docs.oracle.com/javase/9/docs/api/java/lang/Integer.html#parseInt-java.lang.CharSequence-int-int-int-

This will remove the need to create a custom parseInt() utility.

jpdahlke:
> Recommend waiting until Emissary is updated to Java 9 or later and making the following API call:
> https://docs.oracle.com/javase/9/docs/api/java/lang/Integer.html#parseInt-java.lang.CharSequence-int-int-int-
> 
> This will remove the need to create a custom parseInt() utility.

This seems like an appropriate ask if we can stand to wait. Copying the Java 9 method in the short term could get us there too such that after an upgrade to Java it's as easy as simply replacing our copied method w/ the native call. The compare method would have to change regardless to take advantage of the update.

drivenflywheel (approved):
Looks good

src/main/java/emissary/util/ShortNameComparator.java
```diff
@@ -19,34 +19,55 @@
     public int compare(IBaseDataObject obj1, IBaseDataObject obj2) {
         String s1 = obj1.shortName();
         String s2 = obj2.shortName();
+        int index1 = s1.indexOf(Family.SEP, 0);
```
fbruton, line 22: Unnecessary argument. Remove begin index.

src/main/java/emissary/util/ShortNameComparator.java
```diff
@@ -19,34 +19,55 @@
...
+        int index2 = s2.indexOf(Family.SEP, 0);
```
fbruton, line 23: Unnecessary argument. Remove begin index.

fbruton (changes requested):

fbruton (approved):

PR #137, opened 2021-04-19
Improve performance of emissary.output.DropOffUtil.processMetadata() by eliminating String.split()

jdcove2:
This method is called for each BaseDataObject and String.split() is a very expensive call in terms of objects created. The new code does the same job without creating any objects.

### DIFF
--- a/src/main/java/emissary/output/DropOffUtil.java
+++ b/src/main/java/emissary/output/DropOffUtil.java
@@ -1014,7 +1014,7 @@ public class DropOffUtil {
         }
 
         for (final IBaseDataObject p : payloadList) {
-            final int level = p.shortName().split(emissary.core.Family.SEP).length;
+            final int level = StringUtils.countMatches(p.shortName(), emissary.core.Family.SEP) + 1;
             // save specified metadata items for children to grab
             parentTypes.put("" + level, p.getFileType());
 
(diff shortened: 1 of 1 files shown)


jdcove2:
This PR is related to PR#135 in that it eliminates the other set of objects created by String.split() as shown in the first 5 screenshots of PR#135. The following screenshots are analogous to PR#135's screenshots 6-10 and show the elimination of the same objects but from the processMetadata() method. In the 20Meg mysqldump test file, the new code eliminated 6,649,623 objects representing 184,416,212 bytes.

1. Overall profiling results with new code.

image

2. No strings created by the new code.

image

3. No ArrayLists created by the new code.

image

4. No Patterns created by the new code.

image

5. No Matchers created by the new code.

image

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -995,6 +995,19 @@ public Date parseEventDate(final String dateString, final boolean supplyDefaultO
         return date;
     }
 
+    private static int countInstances(String text, String find) {
```
fbruton, line 998: Can we replace with ```org.apache.commons.lang3.StringUtils.countMatches``` instead of creating a custom method?

jdcove2: The implementations are basically the same, so the change has been made.

fbruton (approved):

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -1086,7 +1086,7 @@ public void processMetadata(final List<IBaseDataObject> payloadList) {
                 final List<IBaseDataObject> childObjList = p.getExtractedRecords();
                 Collections.sort(childObjList, new emissary.util.ShortNameComparator());
                 for (final IBaseDataObject child : childObjList) {
-                    final int level2 = child.shortName().split("-att-").length;
+                    final int level2 = StringUtils.countMatches(child.shortName(), emissary.core.Family.SEP) + 1;
                     final int parentLevel = level2 - 1;
```
jpdahlke, line 1090: Can these two lines be combined as:

`final int parentLevel = StringUtils.countMatches(child.shortName(), emissary.core.Family.SEP);`

jdcove2: Good catch. Updated.

jpdahlke (approved):

PR #138, opened 2021-04-19
Improve performance of emissary.output.DropOffUtil.processMetadata() by changing loop…

jdcove2:
This method is called for each BaseDataObject. The current "for" loop syntax creates a lot of iterator objects and should be changed so that it does not.

### DIFF
--- a/src/main/java/emissary/output/DropOffUtil.java
+++ b/src/main/java/emissary/output/DropOffUtil.java
@@ -1006,8 +1006,10 @@ public class DropOffUtil {
         // relies on the attachments being sorted
         final Map<String, String> parentTypes = new HashMap<String, String>();
         final IBaseDataObject tld = payloadList.get(0);
+        final List<String> extended_filetypes = new ArrayList<>();
         parentTypes.put("1", tld.getFileType());
-        for (final String param : this.parentParams) {
+        for (int i = 0; i < parentParams.size(); i++) {
+            final String param = parentParams.get(i);
             if (tld.hasParameter(param)) {
                 parentTypes.put("1" + param, tld.getStringParameter(param));
             }
@@ -1030,9 +1032,9 @@ public class DropOffUtil {
             }
 
             if (p.getStringParameter("EXTENDED_FILETYPE") == null) {
-                final List<String> extended_filetypes = new ArrayList<String>();
+                extended_filetypes.clear();
                 for (final Map.Entry<String, Collection<Object>> entry : p.getParameters().entrySet()) {
-                    final String key = entry.getKey().toString();
+                    final String key = entry.getKey();
                     if (key != null && key.endsWith("_FILETYPE")) {
                         for (final Object value : entry.getValue()) {
                             final String vs = value.toString();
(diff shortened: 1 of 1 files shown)


jdcove2:
DropOffUtil.parseMetadata() uses the "new" style for loop syntax which creates an iterator object for each execution of the loop. Iterating over Lists should not be done this way since the "old" style for loop syntax achieves the same result without creating an object. This is evidenced using the 20Meg mysqldump file as an example where parseMetadata() creates 1,773,237 objects representing 56,743,584 bytes unnecessarily as shown in the following screen shots.

1. ArrayList iterator objects created by the old code.

image

2. No ArrayList iterator objects created by new code.

image

LightSeekerSC (approved):
This looks good to me. There are no processing differences and this reduces the creation of iterators, resulting in saved memory.

jpdahlke:
I would like @fbruton to weigh in on this one. I believe he had some concerns.

fbruton (commented):
I recommend updating the loop for ```parentParams``` and ```extended_filetypes``` and declaring the data type as an ```ArrayList``` instead of ```Lists```.

Also, can you profile ```.clear()``` change separately. I'm not sure if that change is going to make much of a difference.

Also, recommend holding off on updating the loop for ```childObjList``` and ```payloadList``` as part of this merge request.

As we continue to profile and look for performance updates, maybe ```ArrayList``` isn't the desired  data structure for IBDO records.

jdcove2:
To make sure I understand the comments:
1. "declaring the data type as an ArrayList instead of Lists."
    - Change method argument, "payloadList", to be ArrayList instead of List
2. "can you profile .clear() change separately. I'm not sure if that change is going to make much of a difference."
    - The "extended_filetypes.clear()" was not a performance enhancement. It was put in because the previous code created a new ArrayList that is by default empty, so I cleared the list to mimic that behaviour.
3. "recommend holding off on updating the loop for childObjList and payloadList"
    - Revert 1018-1019 (payloadList)
    - Revert 1094-1095 (childObjList)
4. "As we continue to profile and look for performance updates, maybe ArrayList isn't the desired data structure for IBDO records."
    - I'm not sure what data structure will perform better than ArrayList for just holding a list of IBDO's.

fbruton:
> To make sure I understand the comments:
> 
> 1. "declaring the data type as an ArrayList instead of Lists."
>    
>    * Change method argument, "payloadList", to be ArrayList instead of List
> 2. "can you profile .clear() change separately. I'm not sure if that change is going to make much of a difference."
>    
>    * The "extended_filetypes.clear()" was not a performance enhancement. It was put in because the previous code created a new ArrayList that is by default empty, so I cleared the list to mimic that behaviour.
> 3. "recommend holding off on updating the loop for childObjList and payloadList"
>    
>    * Revert 1018-1019 (payloadList)
>    * Revert 1094-1095 (childObjList)
> 4. "As we continue to profile and look for performance updates, maybe ArrayList isn't the desired data structure for IBDO records."
>    
>    * I'm not sure what data structure will perform better than ArrayList for just holding a list of IBDO's.

1. That's if we decide to update the loop for payloadList.
2. ```clear()``` doesn't resize the array list. It could potentially hold more memory if the size of the Arraylist being created differed drastically. Should not be an issue with smaller list of a similar size, which should be the case for that particular field. Overall, I'm okay with the change.
3. Yes
4. That may be the case for specific use cases.  As a framework, it should be flexible enough to meet the needs of various types of use cases.  For example, will the system perform more reads or writes to the records, may determine if using an ArrayList or LinkedList is more efficient. If it's decided that ArrayList is the best option for most/all cases, then I believe the API should reflect that.

jdcove2:
Is the following closer to what you want?

1. Make this code change in a new "issue".
2. Leave this code change in the current PR/Issue. Note that the ArrayList internal array size can be changed with "trimToSize".
3. Make this code change in a new "issue".
4. Make this a consideration in a new "issue".

fbruton (approved):

PR #139, opened 2021-04-19
Improve performance of emissary.output.DropOffUtil.processMetadata() by improving getCookedParameters.

jdcove2:
This method is called for each BaseDataObject and creates a new TreeMap each time along with a bunch of TreeMap$Entry objects. This can be reduced to reusing a single Map object with no Entry objects being created.

### DIFF
--- a/src/main/java/emissary/output/DropOffUtil.java
+++ b/src/main/java/emissary/output/DropOffUtil.java
@@ -679,6 +679,7 @@ public class DropOffUtil {
      * @param metaData map of payload metadata
      * @return the file type (also added to metaData if not present)
      */
+    @Deprecated
     public String getFileType(final Map<String, String> metaData) {
         return getFileType(metaData, null);
     }
@@ -690,6 +691,7 @@ public class DropOffUtil {
      * @param formsArg space separated string of current forms
      * @return the file type (also added to metaData if not present)
      */
+    @Deprecated
     public String getFileType(final Map<String, String> metaData, final String formsArg) {
         String forms = formsArg;
         if (forms == null) {
@@ -736,7 +738,76 @@ public class DropOffUtil {
     }
 
     /**
-     * Parameters that are used to determine filetype in {@link #getFileType(Map, String)}
+     * Get the file type from the metadata or the form string passed in
+     *
+     * @param bdo IBaseDataObject
+     * @return the file type
+     */
+    public static String getFileType(final IBaseDataObject bdo) {
+        return getAndPutFileType(bdo, null, null);
+    }
+
+    /**
+     * Get the file type from the IBaseDataObject or the form string passed in
+     *
+     * @param bdo IBaseDataObject
+     * @param metaData Optional map of metadata that might be modified.
+     * @param formsArg Optional space separated string of current forms
+     * @return the file type
+     */
+    public static String getAndPutFileType(final IBaseDataObject bdo, final Map<String, String> metaData, final String formsArg) {
+        String forms = formsArg;
+        if (forms == null) {
+            forms = bdo.getStringParameter(FileTypeCheckParameter.POPPED_FORMS.getFieldName());
+            if (forms == null) {
+                forms = "";
+            }
+        }
+
+        String fileType;
+        if (bdo.hasParameter(FileTypeCheckParameter.FILETYPE.getFieldName())) {
+            fileType = bdo.getStringParameter(FileTypeCheckParameter.FILETYPE.getFieldName());
+        } else if (bdo.hasParameter(FileTypeCheckParameter.FINAL_ID.getFieldName())) {
+            fileType = bdo.getStringParameter(FileTypeCheckParameter.FINAL_ID.getFieldName());
+            logger.debug("FINAL_ID FileType is ({})", fileType);
+            if (metaData != null) {
+                metaData.put(FileTypeCheckParameter.FILETYPE.getFieldName(), fileType);
+            }
+        } else {
+            if (forms.contains(" ")) {
+                fileType = forms.substring(0, forms.indexOf(" ")).trim();
+                if (metaData != null) {
+                    metaData.put(FileTypeCheckParameter.COMPLETE_FILETYPE.getFieldName(), forms);
+                }
+            } else {
+                fileType = forms;
+            }
+            if (StringUtils.isEmpty(fileType)) {
+                if (bdo.hasParameter(FileTypeCheckParameter.FONT_ENCODING.getFieldName())) {
+                    fileType = "TEXT";
+                } else {
+                    fileType = "UNKNOWN";
+                }
+            }
+
+            if (metaData != null) {
+                metaData.put(FileTypeCheckParameter.FILETYPE.getFieldName(), fileType);
+            }
+        }
+
+        if ("UNKNOWN".equals(fileType) && forms.contains("MSWORD")) {
+            fileType = "MSWORD_FRAGMENT";
+        }
+
+        if ("QUOTED-PRINTABLE".equals(fileType) || fileType.startsWith("LANG-") || fileType.startsWith("ENCODING(")) {
+            fileType = "TEXT";
+        }
+
+        return fileType;
+    }
+
+    /**
+     * Parameters that are used to determine filetype in {@link #getAndPutFileType(IBaseDataObject, Map, String)}
      */
     public enum FileTypeCheckParameter {
         COMPLETE_FILETYPE("COMPLETE_FILETYPE"), FILETYPE("FILETYPE"), FINAL_ID("FINAL_ID"), FONT_ENCODING("FontEncoding"), POPPED_FORMS(
@@ -1021,7 +1092,7 @@ public class DropOffUtil {
                     }
                 }
                 if (!extended_filetypes.isEmpty()) {
-                    final StringBuilder extft = new StringBuilder(getFileType(p.getCookedParameters()));
+                    final StringBuilder extft = new StringBuilder(getFileType(p));
                     for (int j = 0; j < extended_filetypes.size(); j++) {
                         final String s = extended_filetypes.get(j);
                         extft.append("//").append(s);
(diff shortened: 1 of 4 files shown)


jdcove2:
Calling BaseDataObject.getCookedParameters() creates a new TreeMap each time it is called which then creates an entry for each mapping. For the 20Meg mysqldump test file there are 4,433,080 maps and entries created for a total of 180,869,661 bytes. Reusing a single map that is backed by ArrayLists reduces this to "order 10".  This is shown in the following screen shots.

1. TreeMap objects created by old code.

image

2. TreeMap$Entry objects created by old code.

image

3. No TreeMap objects created by new code.

image

4. No TreeMap$Entry objects created by new code.

image

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -995,6 +999,82 @@ public Date parseEventDate(final String dateString, final boolean supplyDefaultO
         return date;
     }
 
+    public static class ArrayListMap<K, V> extends AbstractMap<K, V> {
```
LightSeekerSC, line 1002: Why are we creating this map class? Also this doesn't seem to be a map. This has a lookup speed of O(n), whereas maps are expected to have a lookup of O(1)

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -1006,6 +1086,7 @@ public void processMetadata(final List<IBaseDataObject> payloadList) {
         // relies on the attachments being sorted
         final Map<String, String> parentTypes = new HashMap<String, String>();
         final IBaseDataObject tld = payloadList.get(0);
+        final Map<String, String> cookedParameters = new ArrayListMap<>();
```
LightSeekerSC, line 1089: Why not use a HashMap?

jpdahlke:
Let's hold off on this one for now. Maintaining a custom data set for this application may be a bit much.

LightSeekerSC:
> Let's hold off on this one for now. Maintaining a custom data set for this application may be a bit much.
> 
> Does it make sense that instead of building the whole TreeMap for filetype lookups, just have a getFileType method that performs the lookups of the keys necessary. It could look just like the current method but instead of a metaData.containsKey, do a p.getStringParameter().

I agree that this is the solution that we should pursue. Using getCookedParameters here is unnecessary (duplicating every single parameter to get a limited subset) and considering that this is critical path, we would do good in optimizing the getFileType() method to only check the desired fields.

fbruton (approved):

jpdahlke:
Are we accepting the api breaking change for this?

LightSeekerSC:
> Are we accepting the api breaking change for this?

Are you referring to the fact that the method signature is changing? I think that it is worthwhile, considering the substantial savings this can have at scale

Would this require us to do a major version update?

drivenflywheel:
> > Are we accepting the api breaking change for this?
> 
> Are you referring to the fact that the method signature is changing? I think that it is worthwhile, considering the substantial savings this can have at scale
> 
> Would this require us to do a major version update?

While migrating to the new `DropOffUtil.getFileType` signatures, it might be wise to retain the existing ones and supporting methods, annotated as `@Deprecated`, because this API change will cause upstream build failures.  They should be easily resolved, but there won't be any choice about when to resolve them.  Deprecation would prevent the upstream build failures but still provide a path to leverage the changes via a follow-on ticket.

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -712,52 +712,52 @@ public String makeDotFile(final String fullName) {
     /**
      * Get the file type from the metadata or the form string passed in
      *
-     * @param metaData map of payload metadata
+     * @param bdo IBaseDataObject
      * @return the file type (also added to metaData if not present)
      */
-    public String getFileType(final Map<String, String> metaData) {
```
drivenflywheel, line 718: We should retain this old method signature and annotate it as `@Deprecated` to prevent upstream build failures.

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -712,52 +712,52 @@ public String makeDotFile(final String fullName) {
...
 
     /**
      * Get the file type from the metadata or the form string passed in
      *
-     * @param metaData map of payload metadata
+     * @param bdo IBaseDataObject
      * @param formsArg space separated string of current forms
      * @return the file type (also added to metaData if not present)
      */
-    public String getFileType(final Map<String, String> metaData, final String formsArg) {
```
drivenflywheel, line 729: Retain this old method signature and annotate it as `@Deprecated` to prevent upstream build failures.

drivenflywheel (changes requested):
Need to resolve breaking API changes that will cause upstream build failures.

fbruton (changes requested):
I agree that the original API call should be deprecated.

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -772,7 +774,69 @@ public String getFileType(final Map<String, String> metaData, final String forms
     }
 
     /**
-     * Parameters that are used to determine filetype in {@link #getFileType(Map, String)}
+     * Get the file type from the metadata or the form string passed in
```
fbruton, line 777: Update method description by replacing metadata and form string with IBaseDataObject

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -772,7 +774,69 @@ public String getFileType(final Map<String, String> metaData, final String forms
...
+     *
+     * @param bdo IBaseDataObject
+     * @return the file type (also added to metaData if not present)
+     */
+    public static String getFileType(final IBaseDataObject bdo) {
+        return getFileType(bdo, null);
+    }
+
+    /**
+     * Get the file type from the metadata or the form string passed in
```
fbruton, line 787: Update method description by replacing metadata with IBaseDataObject

src/test/java/emissary/output/DropOffUtilTest.java
```diff
@@ -467,50 +465,51 @@ public void testGetEventDate() {
 
     @Test
     public void testGetFileType() {
```
fbruton, line 467: Recommend including the original test, since you're now deprecating the method instead of removing it.

fbruton (changes requested):

fbruton (approved):

jdcove2:
Observations:
1. getFileType(...,) should not modify BDO passed in.
2. getFileType(...) calculates fileType and completeFileType, so these should be split into two methods.

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -772,7 +774,90 @@ public String getFileType(final Map<String, String> metaData, final String forms
...
+    }
+
+    /**
+     * Get the complete file type from the IBaseDataObject or the form string passed in
+     *
+     * @param bdo IBaseDataObject
+     * @param formsArg space separated string of current forms
+     * @return the complete file type
+     */
+    public static String getCompleteFileType(final IBaseDataObject bdo, final String formsArg) {
```
drivenflywheel, line 841: I'm a bit puzzled by this method name.  Why `getCompleteFileType` as the method name, but completely ignore any bdo parameter stored under the key `FileTypeCheckParameter.COMPLETE_FILETYPE.getFieldName()`?  Admittedly, these are a bit hard to grok since the documentation re: expected values stored under of various parameter keys is very sparse in this codebase.

fbruton (commented):
I need help understanding the ```getCompleteFileType```? Does that need to be part of this merge request?

fbruton (changes requested):
Recommend removing ```getCompleteFileType``` for now and including in a separate effort.

jdcove2:
For systems that use Emissary and the current DropOffUtil.getFileType(...) method and care about whether the Map that is passed in contains the "COMPLETE_FILENAME" key/value pair after DropOffUtil.getFileType(...) is called will not function properly using only the new DropOffUtil.getFileType(...) method. The new DropOffUtil.getCompleteFileType(...) method is also needed to completely replace the current DropOffUtil.getFileType(...) method. Therefore the DropOffUtil.getCompleteFileType(...) method cannot be removed from this PR.

drivenflywheel:
> For systems that use Emissary and the current DropOffUtil.getFileType(...) method and care about whether the Map that is passed in contains the "COMPLETE_FILENAME" key/value pair after DropOffUtil.getFileType(...) is called will not function properly using only the new DropOffUtil.getFileType(...) method. The new DropOffUtil.getCompleteFileType(...) method is also needed to completely replace the current DropOffUtil.getFileType(...) method. Therefore the DropOffUtil.getCompleteFileType(...) method cannot be removed from this PR.

Here's where I've been stuck:  I get that the new `getFileType()` overload no longer populates a value under the `COMPLETE_FILENAME.getFieldName()` key , and that leaves a gap for any downstream components looking for a value under that key.  

I don't see how the new `getCompleteFileType()` method addresses that. The method isn't populating that key, it just wraps the short algorithm to find the value of what would have been shoved into that key by the eliminated side-effect.  If  `getCompleteFileType()` exists so that a downstream component can call it and stuff its return value under the COMPLETE_FILETYPE key, then that component is where the getCompleteFileType() method should be defined and implemented. Its inclusion here is odd, and the downstream component(s) would already have to change.

jdcove2:
After much discussion, it appears that the most support is behind leaving the new getFileType(...) and getCompleteFileType(...) the way they are, changing method names if desired. Then, add a utility method that completes the original getFileType(...) capability by doing the following:

1. takes an IBaseDataObject and a Map and returns void
2. calls the new getFileType(...) and getCompleteFileType(...) methods
3. applies the getFileType()/getCompleteFileType() results to the map, as appropriate.

fbruton (changes requested):
My recommendation is to remove ```getCompleteFileType``` and add the following methods:
```public string getFileType(final IBaseDataObject bdo) // calls getAndPutFileType(bdo,null)```
```public string getAndPutFileType(final IBaseDataObject bdo, final Map<String, String> metaData) // derive fileType from bdo and put to metadata if metadata is not empty```
May need to include a method with ```formsArg```. I'm not sure if that is still necessary.

drivenflywheel (approved):
Approved, assuming the resolution of the most recent method names discussion.

fbruton (approved):

PR #141, opened 2021-04-23
Improve performance of emissary.core.BaseDataObject by moving out MetadataDictionary

jdcove2:
It appears that the MetadataDictionary code within BaseDataObject can take a considerable amount of time to run when accessing parameters so it should be moved out.

### DIFF
--- a/src/main/java/emissary/core/BaseDataObject.java
+++ b/src/main/java/emissary/core/BaseDataObject.java
@@ -610,14 +610,7 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
 
     @Override
     public boolean hasParameter(final String key) {
-        // Try remapping
-        try {
-            final MetadataDictionary dict = MetadataDictionary.lookup();
-            return this.parameters.containsKey(dict.map(key));
-        } catch (NamespaceException ex) {
-            // Remapping not enabled
-            return this.parameters.containsKey(key);
-        }
+        return this.parameters.containsKey(key);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


drewfarris (changes requested):
This LGTM, except for the addition of `BaseDataObjectWithRemapping`. I think we should get rid of this entirely.

That said, this is a significant change, so I think that we should bump the major release number when we incorporate this change. Release notes should emphasize this change for users of the Emissary framework.

If the consensus is that we want to phase this in, how about we do `BaseDataObjectWithoutRemapping` in the next minor release, this will allow framework users to migrate to the new mechanism - and then rename this new implementation to `BaseDataObject` in the next major release. This would make it easier to compare the default implementation of `BaseDataObject` across versions as opposed to having to inspect `BaseDataObjectWithRemapping` to observe the default implementation (as one would have to do after this PR is merged as is).

Regardless, I still advocate for the simpler approach of eliminating the mapping immediately and releasing a new major version - that's the least amount of fiddling required.

jpdahlke:
If this is cumbersome to handle as part of the IBDO, it may be more beneficial to stop doing it at insertion-time and move the concept downstream before output is generated. I think that could potentially be done with less impact to end-users and less fiddling.

jpdahlke:
We also might want to look into removing the namespace binding code from the MetadataDictionary. There doesn't seem to be any usefulness of that once MD is removed from the BDO.

dev-mlb:
PR is 6 months old and sounds like other work is needed before we can take this. Can we close until we've addressed the pre-work and then reopen when we are ready to take the PR?

PR #148, opened 2021-05-07
remove Lookup and WorkSpaceClientEnqueue actions

jpdahlke:


drewfarris:
Generally LGTM, the only removal I question is that of `RollOutputs.action` - Although know this is performed using a timer internally, do we have another mechanism for externally forcing emissary to close all open output files (and reopen new ones?)

jpdahlke:
It appears we do not. As far as I can tell, that's what this was originally intended to do, but it just leverages a method that is not realized in any concrete filter class, so it just uses the AbstractFilter `close()` which is a noop and results in:
`$ curl http://localhost:8001/emissary/roll?filter=JSON --anyauth -u emissary
Output Rolled: JSON ResourceStats`

So, it does have the effect of rolling the stats which may be useful, but we might want to handle those separately from outputs anyway.

If an external close of open output files is a feature we want to actually have moving forward (I could see the utility), I believe we would need to update this to obtain a hook to an AbstractRollableFilter which has a Roller or ask the RollManager to perform the operation for us similar to how it performs a `shutdown()`. 

With that, I will remove this from the PR and we can discuss later how to handle these scenarios.

FineAndDandy (approved):

jpdahlke:
I'm going off of @drewfarris previous LGTM with RollOutput.action reinstated.

PR #150, opened 2021-05-15
README Updates

LightSeekerSC:


README.md
```diff
@@ -6,93 +6,33 @@
 Table of Contents
 =================
 
-* [Emissary](#emissary)
```
LightSeekerSC, line 9: Removed h4 references and sections that were moved to other READMEs

README.md
```diff
@@ -119,44 +59,52 @@ services, their cost and quality and keep the P2P network connected.
 
 ## Internals
 
-### Code organization
+### Code Organization
 
 Emissary is laid out in the following package structure:
 
-* admin - code that starts Places
```
LightSeekerSC, line 126: Made folder names bold to enhance readability

README.md
```diff
@@ -168,305 +116,247 @@ Typical Java threads
...
 are used to control certain aspects of unwrapping and processing and help to ensure that the workflow can always make 
 progress on the task at hand.
 
 There are guidelines on how the workflow stages should be used. These guidelines are not currently enforced but 
 deviations from these guidelines might be an indication that more thought should be applied.
 
-##### Study
-
-The name for this stage came from the idea behind [Perl's Regex study](http://perldoc.perl.org/functions/study.html) 
+Stage Name    | Description                  | Run in Parallel?
```
LightSeekerSC, line 155: Moved table up to give an up-front orderflow

README.md
```diff
@@ -168,305 +116,247 @@ Typical Java threads
...
+COORDINATE    | Coordinate processing        | No
+PRETRANSFORM  | Before transform hook        | Yes
+TRANSFORM     | Transformation phase         | No
+POSTTRANSFORM | After transform hook         | Yes
+ANALYZE       | Analysis/metadata generation | Yes
+VERIFY        | Verify output                | No
+IO            | Output                       | No
+REVIEW        | Finish off                   | No
+
+* #### Study
```
LightSeekerSC, line 168: Bulleted phases to make it clearer that these were subsections of the Workflow section

README.md
```diff
@@ -168,305 +116,247 @@ Typical Java threads
...
 
-Sometimes, for our own convenience, or for dealing with formats provided by up-stream systems,
-the data arrives packaged in various formats. These containers can have multiple items that
-should not be logically related in a family tree. The Emissary Parser framework deals with this
-situation.
+The Emissary Parser framework is designed to extract information from any data containers/wrappers on the payload that may contain information useful for identification or processing.
 
-Parsers need to make a quick identification of the format of the container. This is called
-out by the emissary.parser.ParserFactory and uses the emissary.parser.DataIdentifier as the
+Parsers need to make a quick identification of the container format (i.e. Apache NiFi). This is called
```
LightSeekerSC, line 251: Listed a practical example of a container format to help users better understand how this applies

README.md
```diff
@@ -168,305 +116,247 @@ Typical Java threads
...
 Parser framework is operating on the single-threaded side of the system and handing items over to the
-multi-threaded (MobileAgent) side of the system. This implies that the parsers are fast enough
+multi-threaded (MobileAgent) side of the system. This necessitates that the parsers are fast enough
 to keep the number of configured threads busy.
 
 ## Development
 
 ### Requirements
 
-#### Linux or MacOSX operating system
```
LightSeekerSC, line 325: Made into a preference (see below) instead of a requirement below. Windows now has a unix bash terminal and development there _is_ possible

README.md
```diff
@@ -168,305 +116,247 @@ Typical Java threads
...
-As you go through this guide, keep in mind the steps in the [default lifecycle](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Lifecycle_Reference)
-are cumulative.  If you run *test*, Maven will run *compile* among other
-steps first.  Similarly, when you run *package*, Maven will run
-*compile* and *test* along with other steps first.
-
-We also use [maven profiles](http://maven.apache.org/guides/introduction/introduction-to-profiles.html).
-To activate specific functionality.  To activate profile abc, you would add `-Pabc`.  Mutliple profiles can be
-activated at once if you separate them with a comma.  Something like `-Pprofile-1,profile-2`. 
-
-#### Git 1.8+
```
LightSeekerSC, line 359: User can choose what version of Git they want and if they are on Github, we can treat that as assumed knowledge

README.md
```diff
@@ -168,305 +116,247 @@ Typical Java threads
...
-[Java 1.8](http://www.oracle.com/technetwork/java/javase/downloads/index-jsp-138363.html).  
-OpenJDK should work, although most testing is done with Oracle's JDK.  Be sure to use a recent version of
+Emissary requires [Java 1.8](http://www.oracle.com/technetwork/java/javase/downloads/index-jsp-138363.html).  
+Emissary now supports both OpenJDK and Oracle JDK.  Be sure to use a recent version of
 1.8 as there are bugs in earlier versions.
 
 #### Apache Maven 3.2
 
 Download and install [Apache Maven 3.2](http://maven.apache.org).
-Some maven commands are shown below.  A great place to start with Maven is the
```
LightSeekerSC, line 343: Condensed this info (see below). We aren't maven, so we can just point the user to good resources and save space in our README

README.md
```diff
@@ -168,305 +116,247 @@ Typical Java threads
...
-```
-
-Inspect the output to see what tests have failed.  More information
-about failed tests will be under *target/surefire-tests* should need
-to see more output for a failed test.
 
-As configured, each run will execute the tests in random order.  This
-is intentional to highlight dependencies between tests.
-
-Currently, the POM used .5C for the forkCount when running tests.  This means
```
LightSeekerSC, line 426: Condensed the fork info into a comment

README.md
```diff
@@ -496,57 +385,63 @@ mvn clean verify -Dcode-quality
...
-generation from [above](code-quality)
+### Logging
 
-#### Install
+Logging is handled by [logback](Logback Home).  By default, the file logback.xml in the \<configDir\> will be 
+used.  You can point to another file with the *--logbackConfig* argument.  Feel free to modify the file
+in target/config if you want different logging.  The logback.xml is configured to only log to the console 
+currently.
 
-The *install* task runs after *verify* and copies any included
```
LightSeekerSC, line 504: This is Maven info, not Emissary info. We have already referenced the user to other documents that better explain this

README.md
```diff
@@ -573,441 +468,19 @@ debugger attaches, and you can then step through the code.
...
 ```
 
-#### Potentially speed up the development cycle
+#### Potentially Speed Up the Development Cycle
 
 The above commands all have 'clean' in the command.  It is safest to do that.  But it is 
 possible to omit the 'clean' and maven will then only compile files that have changed.  This can
 speed up things, but for Emissary it is unlikely to speed up much.
 
-### IDE integration
```
LightSeekerSC, line 582: Moved to all Emissary Commands info into a separate README, since there was so much content here. The new README is linked to above so users can easily navigate to it

README.md
```diff
@@ -573,441 +468,19 @@ debugger attaches, and you can then step through the code.
...
-For info on how to setup remote debugging in your IDE, reference the
-following articles:
-
-- for
-[Eclipse](http://help.eclipse.org/kepler/index.jsp?topic=%2Forg.eclipse.jubula.client.ua.help%2Fhtml%2Freference%2Fnode47.html)
-- for [IntelliJ](https://www.jetbrains.com/idea/help/run-debug-configuration-remote.html)
-
 ## Docker
 
-Docker needs to be installed locally to build images. If developing on Linux, the maven plugin for Docker cannot be run
```
LightSeekerSC, line 906: Moved all of the Docker info into a separate README so that it is self-contained and easy to find

README.md
```diff
@@ -1033,21 +506,12 @@ may get redirected for unknown hosts to 92.242.140.21.  To test this run
 nslookup somehost
 ```
 
-If you have this situation, then it is recommended to Google's DNS servers.  [Here](https://developers.google.com/speed/public-dns/docs/using) 
-are some instructions from Google.
-
-### Can't run tests in Eclipse
```
LightSeekerSC, line 1039: This section and the following section don't seem helpful and almost seem condescending

contrib/READMEs/EMISSARY-COMMANDS.md
```diff
@@ -0,0 +1,215 @@
...
+The config directory is defaulted to <projectBase>/config
+but can also be passed in with *(-c or --config)*.  When running from the git checkout, you should use
+*target* as the projectBase.  Feel free to modify config files in target/config before you start.
+
+See the *help -c <commandName>* for each command to get more info.
+
+### What
+
+This command will use the configured engines to identify the file.  Emissary currently only comes with the 
+SizeIdPlace, so the id will TINY or SMALL etc.  See that class for more info.  The *-i or --input* argument
```
ggambrell, line 24: so the id will **be** TINY, SMALL, etc.

contrib/READMEs/EMISSARY-COMMANDS.md
```diff
@@ -0,0 +1,215 @@
...
+this can be high.  You can control the number of agents with *-a or --agents*.  Here 
+is an example run.
+
+```
+./emissary server -a 2
+```
+
+Without further configuration, it will start on http://localhost:8001.  If you browse to that 
+url, you will need to enter the username and password defined in target/config/jetty-users.properties,
+which is emissary and emissary123.
```
ggambrell, line 46: Should we remove the username and password? I understand it is a default, but this seems like bad practice.

contrib/READMEs/EMISSARY-COMMANDS.md
```diff
@@ -0,0 +1,215 @@
...
+
+```
+./emissary server -a 2 -m cluster
+./emissary server -a 2 -m cluster -p 9001
+```
+
+Because these nodes all know about ports 8001, 9001 and 7001, you will see errors in the logs as they
+continue to try to connect.  
+
+Note, in real world deployments we don't run multiple Emissary processes on the same node.  You can configure the
```
ggambrell, line 125: "Note, it is recommended to only run one Emissary process on any single node." ?

contrib/READMEs/EMISSARY-COMMANDS.md
```diff
@@ -0,0 +1,215 @@
...
+Pool in clustered mode also does the same pool in standalone.  It starts at the node (locahost:9001) by default
+then goes to all the nodes it knows about and aggregates a collapsed view of the cluster.  Run it with
+
+```
+./emissary pool --cluster
+```
+
+### Topology (Clustered)
+
+The topology talks to the configured node (localhost:8001 by default) and talks to every it knows about.
```
ggambrell, line 164: and talks to every **node** it knows about.

contrib/READMEs/IDE-SETUP.md
```diff
@@ -0,0 +1,71 @@
...
+
+When running tests in Eclipse, so you will need to set the environment variable PROJECT_BASE to the directory
+eclipse is using to build the project.  So in the run configuration of a test, add the following to the 
+Run Configuration -> Environment -> New
+
+```
+PROJECT_BASE = ${project_loc}/target-eclipse
+```
+
+Unfortunately you will have to do this for every test, that's Eclipse with the M2E plugin for you.
```
ggambrell, line 59: I would recommend removing the last half of this sentence. No need to beat down another product.

ggambrell (changes requested):

contrib/READMEs/EMISSARY-COMMANDS.md
cfkoehler: I think it is ok to have the username and password. If someone goes to start with all defaults then they will have to dig into the config to see what the default is.

cfkoehler (changes requested):
@ggambrell caught all the changes I would like made as well

contrib/READMEs/EMISSARY-COMMANDS.md
LightSeekerSC: There has been some experimentation with running multiple instances on the same node. Maybe we could even remove this guidance? @jpdahlke

LightSeekerSC:
I made the requested changes. We can proceed or wait for @jpdahlke for guidance on the multiple Emissary process documentation

ggambrell (approved):

cfkoehler (approved):

README.md
```diff
@@ -496,57 +385,63 @@ mvn clean verify -Dcode-quality
...
 ```
 
-After a successful build, there will be a bundle located in the target directory:
+### Emissary Commands
 
-```
-target/emissary-<version>-dist.tar.gz
+For information on emissary commands, checkout out the [Emissary Commands README](contrib/READMEs/EMISSARY-COMMANDS.md)
+
+For basic informaion about Emisary commands, you can run
```
jpdahlke, line 430: information, Emissary

README.md
```diff
@@ -496,57 +385,63 @@ mvn clean verify -Dcode-quality
...
+```aidl
```
jpdahlke, line 431: aid1? aidl?

jpdahlke (changes requested):
In general:
- not a fan of stashing these in the contrib directory. Placing them at the same level the README is now should is fine.
- I would put the command related help back into the README
- instead of breaking the IDE specific section out, make a DEVELOPMENT.md that contains what developers would want to know
- is there a way we can fix the markdown so it browses correctly when using a markdown viewer?

README.md
```diff
@@ -496,57 +385,63 @@ mvn clean verify -Dcode-quality
...
 
-Run the *install* task with
+The classpath is setup in one of 2 ways.  
 
-```
-mvn clean install
-```
+* #### Development
+  * If the *emissary* script is a sibling to pom.xml, it is assumed you are running
+from the git checkout. If a file named .mvn-classpath exists, that file is used to setup the classpath. If that file
```
jpdahlke, line 401: This is incorrect now. 

"If the *emissary* script is a sibling to pom.xml, it is assumed you are running from the git checkout. A file named .mvn-classpath will be created using a Maven command then the Emissary class is launched."

README.md
```diff
@@ -168,305 +116,247 @@ Typical Java threads
...
+MobileAgent walks a given session through all the relevant places.
+
+There are two types of MobileAgent:
+* [MobileAgent](src/main/java/emissary/core/MobileAgent.java) - Used for processing a single payload
+* [HDMobileAgent](src/main/java/emissary/core/HDMobileAgent.java) - Used for processing a bundle of payloads
+
+Processing starts when PickUpPlace's assignToPooledAgent() calls [agent.go(payload, startingLocation)](src/main/java/emissary/pickup/PickUpPlace.java#L751)
 
-#### Processing Routing
+The MobileAgent processing loop is handled in [agentControl()](src/main/java/emissary/core/MobileAgent.java#L249)
```
jpdahlke, line 142: I would not put line-numbers in here as that could be brittle.

README.md
jpdahlke: Apache NiFi FlowFile?

README.md
```diff
@@ -168,305 +116,247 @@ Typical Java threads
...
-Java 1.6 and Java 1.7 are no longer supported in this codebase only 
-[Java 1.8](http://www.oracle.com/technetwork/java/javase/downloads/index-jsp-138363.html).  
-OpenJDK should work, although most testing is done with Oracle's JDK.  Be sure to use a recent version of
+Emissary requires [Java 1.8](http://www.oracle.com/technetwork/java/javase/downloads/index-jsp-138363.html).  
+Emissary now supports both OpenJDK and Oracle JDK.  Be sure to use a recent version of
 1.8 as there are bugs in earlier versions.
 
 #### Apache Maven 3.2
 
 Download and install [Apache Maven 3.2](http://maven.apache.org).
```
jpdahlke, line 278: We specify a min version of 3.3.9 in the pom.

README.md
jpdahlke: Version isn't required, but this info seems no less helpful than the other info.

LightSeekerSC:
> In general:
> 
> * not a fan of stashing these in the contrib directory. Placing them at the same level the README is now should is fine.

We can do that. Previously dev-mlb  wanted to cleanup the base directory which is why I opted for the contrib directory. Are we not concerned about these READMEs cluttering the base directory?

> * I would put the command related help back into the README

There was a lot of content for this section and I think it will help to separate it. Anyone interested in the commands will be able to easily find that content and anyone not interested in the commands won't have to scroll past it to get to the content they want. I am open to whatever the group prefers. @cfkoehler and @ggambrell do you have a preference on this?

> * instead of breaking the IDE specific section out, make a DEVELOPMENT.md that contains what developers would want to know

Not quite sure I understand what you mean here. Do you just want me to rename that README to DEVELOPMENT.md or would you like me to change the content of that README as well?

> * is there a way we can fix the markdown so it browses correctly when using a markdown viewer?

I don't seem to have issues with my markdown viewer. Can you help me understand what issues you are running into?

README.md
LightSeekerSC: I hear what you are saying. I do think it is helpful to have links to the functions so that devs can easily see the code referenced.

Based on this stackoverflow article it looks like you can reference a specific commit, then the line numbers wouldn't change: https://stackoverflow.com/questions/18220650/github-link-to-function-in-source

Would that be a good compromise for us? Then we know the line numbers will be valid and the dev can still easily link to the function. The only down-side is that the file they are viewing may not reflect the current code

ggambrell:
Going down the bullet points.
1. I would agree with Jason that the README should be in the base directory. 
2. Agree with Stephen, separating the command help makes the document easier to read.
3. Rename IDE-SETUP.md to DEVELOPMENT.md I believe is what Jason meant.
4. Markdown seems to be fine to me as well.

LightSeekerSC:
I added a "temp" commit to see the changes made and then rebased the changes into a single commit

LightSeekerSC:
> Going down the bullet points.
> 
> 1. I would agree with Jason that the README should be in the base directory.
> 2. Agree with Stephen, separating the command help makes the document easier to read.
> 3. Rename IDE-SETUP.md to DEVELOPMENT.md I believe is what Jason meant.
> 4. Markdown seems to be fine to me as well.

These changes in addition to the changes requested by @jpdahlke were made

DOCKER.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+Alternatively, we can use Docker directly. First run a full maven build and then run the ```docker build``` command:
+
+```
+mvn clean package -Pdist
+docker build -f contrib/docker/Dockerfile . -t emissary:latest
+```
+
+### Run Emissary with Docker
```
drivenflywheel, line 26: This section is a bit confusing, because it uses "emissary" as the name for 3 different things, and that might not be obvious to people who are new to Emissary and/or Docker:

1. the name of the **Docker image** we build
2. the name of the **container** instance we launch from that image
3. the **command** we use within the container to launch the emissary server.

IMO, the commands in this section would be much more clear if we make two changes:

1. within each `docker run` command, specify the image using explicit `emissary:latest` syntax.  For Docker newbies, this will make it obvious that instance of "emissary" refers to the *image*; experienced Docker users should know the tag we're specifying could be omitted, but it shouldn't bother them.
2. use a container name that's not "emissary" (for instance, `--name em_container`).  This makes the removes ambiguity in any of the below commands. 

Thoughts?

drivenflywheel (approved):
No blockers, some minor comments that can all be considered optional suggestions.

dev-mlb:
> When talking to some of the newer developers on the project at the time (@ggambrell and @cfkoehler) I was under the impression that this separation made it easier for them to parse the information and find what they needed. Since new developers is generally our target demographic for READMEs would we want to keep what seemed easier for them? Both you (@dev-mlb) and @jpdahlke seem to be in favor of keeping the docs together. If there is split opinion it would make more sense to default to your opinions as the more senior devs.
> 
> With this added information let me know if you still want these combined or not and I will proceed as requested.

I think the split between README and DEVELOPING is good, but I don't like the separate docs for commands and docker. I'd combine the commands -> README, and docker -> DEVELOPING.

In the README I'd like to see is the intro section, reqs list (maven, java, etc), and then a quick start -- a small section on how to build and run and possibly a quick docker example as well. I'd like the user to be able to get up and running pretty quickly w/o having to dig too deep into the docs. Then I think we can deep dive into the details of Emissary and commands -- I think there are some commands that are not documented that should be added. This doc should be everything about running and using Emissary. Also, the ToC section is huge - maybe there is a better way to handle? 

I'm thinking in DEVELOPING we can have sections like (not in order) reqs, code organization, threads, building/distributions, ide, etc. Anything that leans more towards dev stuff, put it in there. If we want a section to be in both docs, I'd rather have a quick read in the README and more detailed version in DEVELOPING - if it makes sense. 

Thoughts?

LightSeekerSC:
According to github's best-practices recommendation, they recommend a wiki for in-depth content: https://guides.github.com/features/wikis/

If we had a wiki, it acts as a separate repository that could also be migrated/cloned wherever we need in the future: https://stackoverflow.com/questions/21992151/migrate-or-export-github-wiki-to-gitlab

I personally like keeping information divided into smaller pieces that feel more manageable to work with. If we go with the wiki option we could do this well. If you don't want to do that, then I will work on the changes you requested

burritoqueen:
We ensured these updates were all addressed in #194 which has a different file layout.  Closing this PR.

PR #154, opened 2021-05-26
Update 'fatal' log message to be a single call

arp-0984:


src/main/java/emissary/pool/AgentThreadGroup.java
```diff
@@ -20,13 +20,6 @@ public void uncaughtException(Thread thread, Throwable throwable) {
...
             throw (java.lang.ThreadDeath) throwable;
         }
-        logger.error("#######    #    #######    #    #      ");
-        logger.error("#         # #      #      # #   #      ");
-        logger.error("#        #   #     #     #   #  #      ");
-        logger.error("#####   #     #    #    #     # #      ");
-        logger.error("#       #######    #    ####### #      ");
-        logger.error("#       #     #    #    #     # #      ");
-        logger.error("#       #     #    #    #     # #######");
         logger.error("Fatal Thread Error on " + thread.getName() + " had an uncaught exception", throwable);
```
drivenflywheel, line 23: If we want this message to have greater cosmetic visibility, this could still use a multi-line message and ("i.e., "####################\nFatal Thread Error on ..."), but otherwise this looks good.

jpdahlke: That would certainly avoid the multiple logger calls which is the main issue, but the secondary objectives are to make it human readable and make it consumable downstream w/ log analysis tools. Changing the level to FATAL could achieve this as well, but I'm ok with the simplicity here and leaving it ERROR.

This should be convertible to parameterized logging. That would be my only ask. Bonus points for fixing both log statements.

jpdahlke (changes requested):

src/main/java/emissary/pool/AgentThreadGroup.java
arp-0984: Added parameterized logging for both log statements in the most recent change

src/main/java/emissary/pool/AgentThreadGroup.java
```diff
@@ -17,16 +17,9 @@ public AgentThreadGroup(String s) {
...
         }
-        logger.error("#######    #    #######    #    #      ");
-        logger.error("#         # #      #      # #   #      ");
-        logger.error("#        #   #     #     #   #  #      ");
-        logger.error("#####   #     #    #    #     # #      ");
-        logger.error("#       #######    #    ####### #      ");
-        logger.error("#       #     #    #    #     # #      ");
-        logger.error("#       #     #    #    #     # #######");
-        logger.error("Fatal Thread Error on " + thread.getName() + " had an uncaught exception", throwable);
+        logger.error("Fatal Thread Error on {} had an uncaught exception {}", thread.getName(), throwable);
```
jpdahlke, line 23: The second set of {}'s can be dropped. The exception will be handled naturally w/o replacement as long as it's the last parameter provided.

jpdahlke (changes requested):

src/main/java/emissary/pool/AgentThreadGroup.java
arp-0984: Removed the second set of {}'s

jpdahlke (approved):

jdcove2 (approved):

PR #156, opened 2021-06-09
filter known peers based on configuration

jpdahlke:


drewfarris (approved):
There's much more cleanup to be done, but this is a good start, lets get it merged.

FineAndDandy (approved):

PR #157, opened 2021-06-14
Add max of 50 to computed agents. 

cfkoehler:
The configurable property can override this.

### DIFF
--- a/src/main/java/emissary/pool/AgentPool.java
+++ b/src/main/java/emissary/pool/AgentPool.java
@@ -11,6 +11,10 @@ import org.slf4j.LoggerFactory;
  * Extends the GenericObjectPool to hold MobileAgents, each on it's own thread.
  */
 public class AgentPool extends GenericObjectPool<IMobileAgent> {
+
+    private static final int MAX_CALCULATED_AGENT_COUNT = 50;
+    private static final int BYTES_IN_GIGABYTES = 1073741824;
+
     /**
      * The default name by which we register into the namespace
      */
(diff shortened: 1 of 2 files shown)


jpdahlke (changes requested):
Please add a unit test. Recommend possibly creating a getMaxMemory() method to leverage/override in test to provide maxMem value rather than messing with the Runtime.

cfkoehler:
Added a test class and test's for the computePoolSize. Change it to take in a parameter of max memory.

jdcove2 (approved):

src/main/java/emissary/pool/AgentPool.java
```diff
@@ -35,35 +39,40 @@
...
-        // sizePerAgent = conf.findIntEntry("agent.average_size_kb", 1024) * 1024;
-        // }
-        // catch (IOException ex) {
-        // logger.info("Cannot read config file " + ex.getMessage() + ", using default agent size");
-        // }
-
-        long maxMem = Runtime.getRuntime().maxMemory();
-        // UNUSED
-        // float headRoom = 0.40f; // space for places
+    protected static int computePoolSize(final long maxMemoryInBytes, final Integer propertyOverride) {
```
jpdahlke, line 46: Does poolSizeOverride make more sense?

src/test/java/emissary/pool/AgentPoolTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Test
+    public void testComputePoolSize() throws IllegalArgumentException {
+        // setup expected exception
+        if (expectedException != null) {
+            thrown.expect(expectedException);
+            thrown.expectMessage(expectedExceptionMsg);
+        }
+
+        assertEquals(expectedPoolSize, AgentPool.computePoolSize(maxMemoryInBytes, propertyOverride), "Pool Size Calculation is not as");
```
jpdahlke, line 54: This looks like:
`
Pool Size Calculation is not as ==> expected: <expected val> but was: <actual val>
`
Is this what was desired? "Pool Size Calculation Error"?
`
Pool Size Calculation Error ==> expected: <expected val> but was: <actual val>
`

jpdahlke (approved):

PR #158, opened 2021-06-19
Error Log For Form Containing KeyManipulator.SEPARATOR

LightSeekerSC:


src/main/java/emissary/core/BaseDataObject.java
```diff
@@ -448,6 +451,9 @@ public int enqueueCurrentForm(final String newForm) {
     public int pushCurrentForm(final String newForm) {
         if (newForm == null) {
             throw new IllegalArgumentException("caller attempted to push a null form value");
+        } else if (newForm.indexOf(KeyManipulator.SEPARATOR) > 0) {
```
jpdahlke, line 454: Prefer to use contains for this check

jpdahlke (changes requested):

src/main/java/emissary/core/BaseDataObject.java
```diff
@@ -448,6 +451,9 @@ public int enqueueCurrentForm(final String newForm) {
...
+            // If there is a key separator in the form, then throw an error log as this will cause issues in routing
```
fbruton, line 455: Recommend updating language from ```throw an error log``` to ```log an error```. Also consider adding the potential routing issue to the log statement.

With an updated log description, the comment might not be necessary.

src/main/java/emissary/core/BaseDataObject.java
fbruton, line 454: Should this case also throw an ```IllegalArgumentException```?

fbruton: This issue will more than likely be discovered in a unit test with proper code coverage if an exception is thrown.

drivenflywheel: > Should this case also throw an `IllegalArgumentException`?

The upstream ticket indicates a plan to only log these in the short term, in order _(I'm assuming)_ to quantity the frequency of occurrence and to determine if an exception here would cause more harm than good.

fbruton: Is the intention to not log if ```newForm``` start with a separator? If not, then maybe consider ```StringUtils.contains(newForm, KeyManipulator.SEPARATOR)```

drivenflywheel:
I'd like this PR to get revisited.   I think we should throw an exception if proposed new form contains a period, since we know that case can cause a routing loop.  If we prevent forms that contain periods, we can also dismiss [Security Alert 3](https://github.com/NationalSecurityAgency/emissary/security/code-scanning/3) and [Security Alert 4](https://github.com/NationalSecurityAgency/emissary/security/code-scanning/4), both of which stem from the use of the form to construct a FileSystem path.  By preventing periods, we mitigate the ability to perform directory traversal.

jdcove2:
After a discussion with @jpdahlke, @fbruton and @drivenflywheel, it was decided that the check should be for a "valid" form. A "valid" form is defined as only upper-case letters, lower-case letters, numbers, underscore and dash.

LightSeekerSC:
The new changes are a work in progress to demonstrate two possible solutions for checking form validity

Here are images of the results of testing the different methods
image (3)
image (2)
image (1)

LightSeekerSC:
@jdcove2 The requested changes have been made

drivenflywheel (changes requested):
A few minor changes requested, but none that require much effort.  I like this approach a lot.

src/test/java/emissary/util/PayloadUtilTest.java
```diff
@@ -151,4 +162,64 @@ public void testXmlSerizliaztion() throws Exception {
...
     }
+
+    @Test
+    public void testIsValidForm() {
+        // Check that all expected valid characters are valid
+        String alphaLow = "abcdefghijklmnopqrstuvwxyz";
+        assertTrue("Lower case alpha characters are not considered valid", PayloadUtil.isValidForm(alphaLow));
+        assertTrue("Upper case alpha characters are not considered valid", PayloadUtil.isValidForm(alphaLow.toUpperCase()));
+        assertTrue("Numeric characters are not considered valid", PayloadUtil.isValidForm("0123456789"));
+        assertTrue("Dash and underscore aren't considered valid", PayloadUtil.isValidForm("-_"));
```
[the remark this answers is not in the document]
LightSeekerSC: Added dot and space

LightSeekerSC:
> A few minor changes requested, but none that require much effort. I like this approach a lot.

Thanks! All recommended changes were made

jdcove2 (approved):

drivenflywheel (approved):
Looks good - ship it!

PR #160, opened 2021-06-22
Modified DropOffUtil to put all file extensions in FILEXT

### LINKED ISSUE #159: FILEXT should contain all file extensions, not just the last one.

dlmarion:
Closes #159


fbruton:
In the case of multiple file extensions, should the last file extension be preserved as a separate value?

dlmarion:
That was not mentioned in the internal ticket.

drivenflywheel:
> In the case of multiple file extensions, should the last file extension be preserved as a separate value?

That's a good question, and probably needs clarification.  My first instinct is that this (last extension) would be more aligned with how the data might be searched for.

dlmarion:
Re-reading [issue](https://github.com/NationalSecurityAgency/burrito-grande/issues/1474) it seems that maybe I misunderstood the original requirements. It seems "Original-Filename" may contain multiple file names. I was thinking the issue was with file extensions. I'll rework.

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -1019,12 +1019,12 @@ public void processMetadata(final List<IBaseDataObject> payloadList) {
             parentTypes.put("" + level, p.getFileType());
 
             final String fn = p.getStringParameter("Original-Filename");
```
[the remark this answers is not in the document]
dlmarion: I re-read the issue and modified the code in dedf4c9

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -1019,12 +1019,12 @@ public void processMetadata(final List<IBaseDataObject> payloadList) {
...
-            if (fn != null && fn.indexOf(".") > -1) {
-                final int pos = fn.lastIndexOf(".") + 1;
+            if (StringUtils.isNotEmpty(fn)) {
+                final int pos = fn.indexOf('.') + 1;
                 if (pos < fn.length()) {
-                    final String fext = fn.substring(pos);
+                    final String fext = fn.substring(pos).toLowerCase();
                     if (fext.length() > 0 && fext.length() <= this.maxFilextLen) {
-                        p.setParameter("FILEXT", fext.toLowerCase());
+                        p.setParameter("FILEXT", fext);
```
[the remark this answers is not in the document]
dlmarion: This has been addressed in  dedf4c9

dlmarion:
Note that there is a change in the preserved file extension in this change. Prior to this change the code used the last occurrence of `.` and this change uses the first. For a file `foo.tar.gz` the file extension prior to this change would have been `.gz` and this change makes it `tar.gz`.

drivenflywheel (changes requested):
> Note that there is a change in the preserved file extension in this change. Prior to this change the code used the last occurrence of `.` and this change uses the first. For a file `foo.tar.gz` the file extension prior to this change would have been `.gz` and this change makes it `tar.gz`.

Verbal feedback I've gotten from the original requestor is that this should continue to extract only after the last occurrence of  `.`.   Same extraction as pre-PR, but now extracting the "last file extension" from each of the values stored under the "Original-filename" key

dlmarion:
> Verbal feedback I've gotten from the original requestor is that this should continue to extract only after the last occurrence of .. Same extraction as pre-PR, but now extracting the "last file extension" from each of the values stored under the "Original-filename" key

Addressed in d80fff0e16d7e2382595abba648b2d23ca00e6c6

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -1018,15 +1018,23 @@ public void processMetadata(final List<IBaseDataObject> payloadList) {
...
 
-            final String fn = p.getStringParameter("Original-Filename");
-            if (fn != null && fn.indexOf(".") > -1) {
-                final int pos = fn.lastIndexOf(".") + 1;
-                if (pos < fn.length()) {
-                    final String fext = fn.substring(pos);
-                    if (fext.length() > 0 && fext.length() <= this.maxFilextLen) {
-                        p.setParameter("FILEXT", fext.toLowerCase());
+            if (p.hasParameter("Original-Filename")) {
+                final List<String> extensions = new ArrayList<>();
```
fbruton, line 1022: Recommend extracting lines 1022:1034 to a new method.

dlmarion: What's the rationale for the recommendation? Is this duplicated elsewhere? Where you looking for an associated test for this block of code?

drivenflywheel: Agreed that this method extraction would make the `processMetadata` method more coherent, but I don't consider that a blocker for the PR.  We see virtually the same pattern immediately after this FILEXT processing (https://github.com/NationalSecurityAgency/emissary/pull/160/files#diff-5e7393b5cf28de1fa844df12051fecba96c6962dcbf4aaa3be6b7c233677e410R1040-R1060 would also benefit from a similar method extraction).

fbruton: Just concerned about the overall length and complexity of ```processMetadata```. I believe extracting to method will make it  more readable.

dlmarion: Addressed in 78aa26094eeec7db567017c574de866820e6c0a0

drivenflywheel (approved):

drivenflywheel:
@dlmarion - can you rebase this onto the lastest master at your convenience?
@fbruton - can you gave this a final look after it's been rebased?  

No reason this little PR is still languishing.

dev-mlb (approved):

PR #122, opened 2021-03-06
WIP: Jetty/Jersey upgrade

fbruton:

PR #168, opened 2021-07-27
Make compatible with Java 11

fbruton:


drivenflywheel:
Which JDK11 has this been tested with?  Do we have a prescribed release?  (OpenJDK, Amazon Corretto, ...)

drivenflywheel:
What's the relationship between this PR and #122?  This PR was opened the same day that #122 was renamed from "Java 11 upgrade".   Is there a dependency between the two?

drivenflywheel:
(Closed accidentally because this UI puts buttons with very different intents directly next to each other.  Now Reopened.)

drivenflywheel:
I'll create a separate PR to address CodeQL findings.  The only outstanding issue I see is for this PR is that we need to update *contrib/docker/Dockerfile* to install `java-11-openjdk` instead of `java-1.8.0-openjdk`.  (I've tested this on my own workstation and Emissary works correctly.)

drivenflywheel (changes requested):
Need to update the Dockerfile to specify `java-11-openjdk` instead of `java-1.8.0-openjdk`

jpdahlke:
Closing in favor of #252

PR #169, opened 2021-08-26
Allow a single child and preserve parents data in MultiFileUnixCommandPlace

cfkoehler:


jdcove2 (approved):

src/test/java/emissary/place/MultiFileUnixCommandPlaceTest.java
```diff
@@ -137,6 +138,20 @@ public void testMultiFileUnixCommandPlaceFileWithSingleChildHandling() throws Ex
...
+        assertNotNull("Place must be created", place);
+        place.setFileOutputCommand();
+        createScript(Executrix.OUTPUT_TYPE.FILE, 1);
+
+        place.preserveParentData = true;
+        place.singleOutputAsChild = true;
+
+        List<IBaseDataObject> att = place.processHeavyDuty(payload);
+        assertEquals("One Attachment should be created", 1, att.size());
+        assertEquals("Parent data should be preserved", PAYLOAD_STRING, new String(payload.data()));
```
drivenflywheel, line 152: (optional)
```suggestion
        assertEquals("Parent data should be preserved", PAYLOAD_STRING, new String(payload.data()));
        assertEquals("Child payload should match script output", W, new String(att.get(0).data()).trim());
```

drivenflywheel (approved):

src/test/java/emissary/place/MultiFileUnixCommandPlaceTest.java
cfkoehler: Added check. Thanks

fbruton (approved):

jpdahlke:
Need to confirm how the different child creation scenarios handle copy-down of parameters

cfkoehler:
> Need to confirm how the different child creation scenarios handle copy-down of parameters
The parent Information is copied down to children: 
https://github.com/NationalSecurityAgency/emissary/blob/a3030eed3fc64fbec3197ad24b643a85e1150acd/src/main/java/emissary/place/MultiFileUnixCommandPlace.java#L544

PR #170, opened 2021-08-26
Fix "null" Filetype in Transform History

pechrista:


LightSeekerSC:
Another solution that could work and I think may be preferable (although we should confirm with others) is if we create an overload of DropOffUtil.getFileType that takes an IBDO as the input parameter, then it can call the Map<> metadata version. If a filetype is found and a filetype doesn't currently exist on the IBDO (the parameter doesn't exist), then we can set it. This will improve the efficiency of future calls to the getFileType method (since the filetype will now exist) and will effectively accomplish the same change

drivenflywheel:
> Another solution that could work and I think may be preferable (although we should confirm with others) is if we create an overload of DropOffUtil.getFileType that takes an IBDO as the input parameter, then it can call the Map<> metadata version. If a filetype is found and a filetype doesn't currently exist on the IBDO (the parameter doesn't exist), then we can set it. This will improve the efficiency of future calls to the getFileType method (since the filetype will now exist) and will effectively accomplish the same change

See approved [PR #139](https://github.com/NationalSecurityAgency/emissary/pull/139), which does something very similar.

PR #184, opened 2021-11-22
Guard paths to which output can be written

drivenflywheel:
Addresses CodeQL feedback and resolves three CodeQL security alerts


src/main/java/emissary/server/mvc/internal/RegisterPeerAction.java
```diff
@@ -77,7 +77,7 @@ private Response processRegisterPeer(String peerKey, String dirName) {
         }
 
         final Set<String> set = new HashSet<>();
-        set.add(KeyManipulator.getDefaultDirectoryKey(peerKey));
+        set.add(KeyManipulator.getDefaultDirectoryKey("#" + peerKey));
```
drivenflywheel, line 80: We have existing and sufficient SSRF mitigation on lines 70-73, which restricts the hosts that can be registered through this endpoint.  The "#" prefix added here is used to trigger the CodeQL [RequestForgery.HostnameSanitizer](https://github.com/github/codeql/blob/main/java/ql/lib/semmle/code/java/security/RequestForgery.qll#L54-L77) flag, which indicates that we've sanitized the hostname to prevent a SSRF attack. This flag is considered when evaluating dataflow paths for SSRF / [cwe-918](https://github.com/github/codeql/blob/main/java/ql/src/Security/CWE/CWE-918/RequestForgery.ql) vulnerability, and we trip the flag to signal that we've performed hostname sanitation.  This code change has no functional impact.

src/main/java/emissary/place/Main.java
```diff
@@ -1040,33 +1042,71 @@ public void printPayload(IBaseDataObject payload) {
...
+     *
+     * @param requiredBase required base directory
+     * @param filePath file path to be tested
+     * @return the normalized file path is within the required base directory
+     * @throws IllegalArgumentException if the filePath contains illegal characters or is outside the required base
+     *         directory
+     */
+    static String filePathIsWithinBaseDirectory(String requiredBase, String filePath) throws IllegalArgumentException {
+        // probably an overly simplistic test
+        if (filePath.contains("..")) {
```
drivenflywheel, line 1096: The check on this line is a simple guard that satisfies a CodeQL [path sanitizer](https://github.com/github/codeql/blob/main/java/ql/src/Security/CWE/CWE-022/TaintedPath.ql#L22-L31) check, our sanization is considered (by CodeQL scans) sufficient to defeat [Uncontrolled data used in path expression](https://codeql.github.com/codeql-query-help/java/java-path-injection/)" vulnerability.

Lines 1099 to 1103 perform more thorough validation to ensure that we consider the `filePath` resolves to a location we trust (and doesn't try to perform path traversal to an unexpected location), but we need the guard on line 1096 so line 1100 isn't mistakenly flagged as vulnerable code. 

This change has no production impact:  no existing production code paths reach this method, but it's reachable by legacy `public static main(String[] args)` methods in numerous Place implementations that call the `ServiceProviderPlace.mainRunner` method for ad hoc testing.

src/main/java/emissary/place/Main.java
```diff
@@ -1040,33 +1042,71 @@ public void printPayload(IBaseDataObject payload) {
...
+    /**
+     * Normalizes the provided paths and tests to see whether the file path is within the required base directory.
+     *
+     * @param requiredBase required base directory
+     * @param filePath file path to be tested
+     * @return the normalized file path is within the required base directory
+     * @throws IllegalArgumentException if the filePath contains illegal characters or is outside the required base
+     *         directory
+     */
+    static String filePathIsWithinBaseDirectory(String requiredBase, String filePath) throws IllegalArgumentException {
```
jpdahlke, line 1094: Worth protecting against null/empty inputs?

src/main/java/emissary/place/Main.java
```diff
@@ -1040,33 +1042,71 @@ public void printPayload(IBaseDataObject payload) {
...
             }
         }
     }
 
+    /**
+     * Normalizes the provided paths and tests to see whether the file path is within the required base directory.
+     *
+     * @param requiredBase required base directory
+     * @param filePath file path to be tested
+     * @return the normalized file path is within the required base directory
```
jpdahlke, line 1090: @return the normalized file path ***if*** within the required base directory

When I read it initially, my mind went to a boolean return.

src/main/java/emissary/server/mvc/internal/RegisterPeerAction.java
jpdahlke: Agreed. 

Would it be worth adding a quick test case for the KeyManipulator that proves this behavior jic functionality gets changes somewhere down the road?

jpdahlke (changes requested):
Good this is getting cleaned up.

Just a few small things.

drivenflywheel:
Incorporated feedback

dev-mlb (approved):

jpdahlke (approved):

Issue #180, opened 2021-11-12
Do not allow duplicate forms in the form stack on IBDO


andrewbp:
"I believe the idea here is that there should never be cause to add a form that already exists on the form stack a second time. Once a form is handled it would probably be inappropriate or erroneous to revisit that form.

confirm this is possible
prevent forms matching a form that already exists in the list of forms from being added (enqueueForm method does this?)
add a log (warn?) message that alerts us to an attempt to add a duplicate form so we could investigate the origin

ibdo.pushCurrentForm allows for duplicate forms to be placed in the list."
-justin-c-4

drewfarris:
> I believe the idea here is that there should never be cause to add a form that already exists on the form stack a second time. 

What about nested archive files, e.g.: a zip file within a zip file?

drivenflywheel:
> What about nested archive files, e.g.: a zip file within a zip file?

Wouldn't those files map to separate IBDO instances?

PR #185, opened 2021-11-23
#180: WIP reworked process of adding forms to prevent dupes

andrewbp:
This work aims to prevent IBDOs from having duplicate forms in the form stack. This seems to break some tests that currently exist and I'm unsure if the tests should be reworked or removed, or what the new outcome of those cases should be.

I've also reworked a lot of the methods that are used to change the form stack: they all seem to return an int value strictly for testing purposes. They now all return void and those tests have been changed to be more obvious about what is being tested. This work has also seemed to cause some methods to be obsolete/redundant, and some of these methods just seem to be wrappers to methods that Collections already provides. Would like another pair of eyes to let me know if I'm going down the right path before I proceed.

### DIFF
--- a/src/test/java/emissary/directory/RoutingAlgorithmTest.java
+++ b/src/test/java/emissary/directory/RoutingAlgorithmTest.java
@@ -260,11 +260,11 @@ public class RoutingAlgorithmTest extends UnitTest {
         loadAllTestEntries();
         final DirectoryEntry eplace = new DirectoryEntry("ERROR.e1.IO.http://example.com:8001/E$5050");
         this.dir.addTestEntry(eplace);
-        this.payload.pushCurrentForm(emissary.core.Form.ERROR);
-        this.payload.pushCurrentForm(emissary.core.Form.ERROR);
+        this.payload.pushCurrentForm("OTHER1");
+        this.payload.pushCurrentForm("OTHER2");
         this.payload.pushCurrentForm(emissary.core.Form.ERROR);
         final DirectoryEntry result = this.agent.getNextKeyAccess(this.dir, this.payload);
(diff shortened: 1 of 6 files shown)


src/test/java/emissary/directory/RoutingAlgorithmTest.java
```diff
@@ -260,11 +260,11 @@ public void testErrorHandlingErrorPopsAllCurrentForms() {
...
         final DirectoryEntry eplace = new DirectoryEntry("ERROR.e1.IO.http://example.com:8001/E$5050");
         this.dir.addTestEntry(eplace);
-        this.payload.pushCurrentForm(emissary.core.Form.ERROR);
-        this.payload.pushCurrentForm(emissary.core.Form.ERROR);
+        this.payload.pushCurrentForm("OTHER1");
+        this.payload.pushCurrentForm("OTHER2");
         this.payload.pushCurrentForm(emissary.core.Form.ERROR);
         final DirectoryEntry result = this.agent.getNextKeyAccess(this.dir, this.payload);
-        assertEquals("Error in error handling place removes all forms", 0, this.payload.currentFormSize());
+        assertEquals("Error in error handling place removes all other forms", 1, this.payload.currentFormSize());
```
andrewbp, line 267: What should happen here? Should all the forms be cleared out, or should it just be left with the Error form?

jpdahlke:
@dev-mlb chime in if you feel different, but I think the goal here was to not impact the BDO api, but just prevent a form from being added to the currentForm list via the enqueueCurrentForm method and just log a message when such an attempt occurs.

andrewbp:
I am going to close this PR and start over. I'm going to keep the scope of the resolution to just the issue that needs to be fixed.

PR #186, opened 2021-11-26
#180: do not allow dupe forms in the IBDO stack

andrewbp:
This work simply adds a "remove(newForm)" call before any attempt to add a new form to the IBDO's form stack. This will remove any already existing matching forms (or do nothing if it doesn't already exist) and then continue as before. The test I wrote adds the same form multiple times in different ways and then checks to see that only one of that form exists in the stack.

As an aside, it may be worth revisiting BaseDataObject for a little bit of refactoring, but that's outside the scope of this ticket.

### DIFF
--- a/src/main/java/emissary/core/BaseDataObject.java
+++ b/src/main/java/emissary/core/BaseDataObject.java
@@ -426,7 +426,7 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
             throw new IllegalArgumentException("caller attempted to add a null form value at position " + i);
         }
 
-        this.currentForm.remove(newForm); // this will remove the form if already in the stack, preventing duplicates
+        checkForDuplicates(newForm);
         if (i < this.currentForm.size()) {
             this.currentForm.add(i, newForm);
         } else {
@@ -441,9 +441,15 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
             throw new IllegalArgumentException("caller attempted to enqueue a null form value");
         }
 
-        this.currentForm.remove(newForm); // this will remove the form if already in the stack, preventing duplicates
+        checkForDuplicates(newForm);
         this.currentForm.add(newForm);
         return this.currentForm.size();
+    }
+
+    private void checkForDuplicates(String newForm) {
+        if (!newForm.equals(Form.ERROR.intern())) {
+            this.currentForm.remove(newForm); // this will remove the form if already in the stack, preventing duplicates
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


andrewbp:
In RoutingAlgorithmTest, testErrorHandlingErrorPopsAllCurrentForms() is currently failing after this change because it is expecting to see at least two ERRORs in the form stack, which is now impossible. There are a few ways to fix this:

1) Allow for multiple ERRORs to exist in the stack
2) Change the test conditions/answers
3) Change the check in MobileAgent, line 454

For now, I will implement allowing multiple ERRORs in the stack as it is the easiest to implement and still keep with current functionality.

src/main/java/emissary/core/BaseDataObject.java
```diff
@@ -440,10 +441,17 @@ public int enqueueCurrentForm(final String newForm) {
             throw new IllegalArgumentException("caller attempted to enqueue a null form value");
         }
 
+        checkForDuplicates(newForm);
```
dev-mlb, line 444: Thinking out loud here, is it better to remove the form or not add it? Essentially, you could be kicking a form to the back of the line - I don't know if that is good or bad. Possibly unintended consequences? Something to think about.

andrewbp: The intention was to push the form to the top if for whatever reason it was encountered again, but I may have gone about it incorrectly.

jpdahlke:
After looking over the current code and talking it over with @dev-mlb a bit, I think if we wanted to do anything now we should start enumerating these instances with logs so we can see how often it happens and under which situations before we take actions to prevent. The ticket does ask for logging updates and confirmation and that might be a better first step.

I'm concerned that there are natural situations we need to know about and then there is the noted ERROR form handling scenario. It may be the case that we are safe from running into multiple ERROR forms with changes we've made to routing and removing moveTo, but the kind of code I would have replaced https://github.com/NationalSecurityAgency/emissary/blob/7.7.0/src/main/java/emissary/core/MobileAgent.java#L454 with was actually bug fixed to be the current state which makes me wonder what I'm missing.

jpdahlke:
> > After looking over the current code and talking it over with @dev-mlb a bit, I think if we wanted to do anything now we should start enumerating these instances with logs so we can see how often it happens and under which situations before we take actions to prevent. The ticket does ask for logging updates and confirmation and that might be a better first step.
> > I'm concerned that there are natural situations we need to know about and then there is the noted ERROR form handling scenario. It may be the case that we are safe from running into multiple ERROR forms with changes we've made to routing and removing moveTo, but the kind of code I would have replaced https://github.com/NationalSecurityAgency/emissary/blob/7.7.0/src/main/java/emissary/core/MobileAgent.java#L454 with was actually bug fixed to be the current state which makes me wonder what I'm missing.
> 
> Just to make sure I'm interpreting correctly, we should scratch this pull request and instead just log instances of duplicate forms?

Yes, let's go that route for now to better understand this behavior.

andrewbp:
Closing this and opening a new pull request: https://github.com/NationalSecurityAgency/emissary/pull/304

PR #188, opened 2021-12-16
Improved ExtractionTest Validation of Attachment and Extracted Children Expected vs Found Count

sambish5:
This is in reference to NationalSecurityAgency/burrito-grande#884.
Improved the ExtractionTest to further checks on Attachment and Extracted Children counts. Includes new tests in TestExtractionTest to verify count check works.


src/main/java/emissary/test/core/ExtractionTest.java
```diff
@@ -155,10 +153,44 @@ protected void checkAnswers(Document answers, IBaseDataObject payload, List<IBas
     }
 
     protected void checkAnswers(Element el, IBaseDataObject payload, List<IBaseDataObject> attachments, String tname) throws DataConversionException {
+        // check to see if payload has extracted records and see if attachments is not null
+        // if true, implement check on payload and attachments to compare expected v found counts
+        // else, implement checkAttachmentAndExtractedRecordCount
+        if ((attachments != null) || payload.hasExtractedRecords()) {
```
dev-mlb, line 159: One case that does not work is if the <numAttachments>0</numAttachments> and you add <att1>...</att1>, this test still passes.

dev-mlb (changes requested):

src/main/java/emissary/test/core/ExtractionTest.java
sambish5: Fixed the issue in the newest commit. Test now fails when numAttachments=0 and if att1 (or more) is present.

src/main/java/emissary/test/core/ExtractionTest.java
```diff
@@ -256,7 +257,27 @@ protected void checkAnswers(Element el, IBaseDataObject payload, List<IBaseDataO
             byte viewData[] = payload.getAlternateView(viewName);
             assertTrue(String.format("Alternate View '%s' is present, but should not be, in %s", viewName, tname), viewData == null);
         }
+    }
 
+    public void checkAttachmentAndExtractedRecordCount(Element el, IBaseDataObject payload, List<IBaseDataObject> attachments, String tname)
+            throws DataConversionException {
+        int numAtt = JDOMUtil.getChildIntValue(el, "numAttachments");
```
dev-mlb, line 264: If I don't define numAttachments, then I can still have att elements in the xml and the test do not fail.

src/main/java/emissary/test/core/ExtractionTest.java
```diff
@@ -256,7 +257,27 @@ protected void checkAnswers(Element el, IBaseDataObject payload, List<IBaseDataO
...
+        int attFoundCount = 0;
+        if (numAtt > -1) {
+            int attCount = 1;
```
dev-mlb, line 267: You should not assume that the att count starts at one. Test will pass if you add att2 element.

src/main/java/emissary/test/core/ExtractionTest.java
dev-mlb: Same issue with extracted records.

dev-mlb: If I don't define numAttachments and/or extractCount, tests pass.

dev-mlb (changes requested):

src/main/java/emissary/test/core/ExtractionTest.java
sambish5: Yep, after I stopped working last night I thought of a lot of these cases. Will push a new commit today to look out for these use cases.

src/main/java/emissary/test/core/ExtractionTest.java
```diff
@@ -1,8 +1,6 @@
 package emissary.test.core;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
```
dev-mlb, line 3: Fix import

dev-mlb (changes requested):
lgtm - just need to fix the import statements

src/test/java/emissary/test/core/TestExtractionTest.java
```diff
@@ -1,5 +1,7 @@
 package emissary.test.core;
 
+import static org.junit.Assert.*;
```
dev-mlb, line 3: import here too

dev-mlb (dismissed):

drivenflywheel:
This seems way more burdensome than the original ask, which was "don't allow the XML to have more child attachment or extractedRecord elements than specified in the `numAttachments` / `extractionCount` elements". 

As currently implemented, we can no longer assert the numAttachments or extractionCount without creating a child element for every one of those children.  That could cause an enormous burden for downstream integration. and it will actively hinder extraction tests for some packaged file processing. If nothing else, the ExtractionTest class should provide the ability to disable this check on a class-by-class or test-by-test basis so downstream projects can integrate this version of emissary without requiring a huge integration PR with thousands of lines of new XML.

sambish5:
Some updates:
- Added all changes into junit5/ExtractionTest file to pair with core/ExtractionTest
- Added check for answer file child `skipAttExtCountCheck`, which if set to true, allows validation to be skipped.

sambish5:
@drivenflywheel @dev-mlb  some major updates within this. Instead of having it fail if `<numAttachments>` and `<extractCount>` do not match `<att#>` and `<extract#>`, it will now just log the non-matching counts. I currently have it written so even if numAttachments/extractCount are not specified, the check still occurs (this can easily be moved to within the `if > -1` parts if we only want it to check when those are specified, but I figured we'd want it to always be specified if there are att or extract in the xml).

If you guys would prefer that method however, I can move it inside of those ifs, and in the future if we want to build back out the functionality to fail if the counts do not match, we can.

dev-mlb (dismissed):

src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -171,8 +171,12 @@ protected void checkAnswers(Document answers, IBaseDataObject payload, List<IBas
     }
 
     protected void checkAnswers(Element el, IBaseDataObject payload, List<IBaseDataObject> attachments, String tname) throws DataConversionException {
-
         int numAtt = JDOMUtil.getChildIntValue(el, "numAttachments");
+        long numAttElements = el.getChildren().stream().filter(c -> c.getName().startsWith("att")).count();
+        if (Math.toIntExact(numAttElements) != numAtt && !(numAttElements <= 0 && numAtt == -1)) {
+            logger.warn("Expected numAttachments and actual <att#> count in {} not equal. Expected: {} Actual: {}", tname, Math.max(numAtt, 0),
```
drivenflywheel, line 177: We have plenty of downstream answer files that only spot check among numerous attachments.  I should be worried if the number of `<attN>` elements exceeds the `<numAttachments>` value, but I'd hate to add hundreds of `<attN>` elements just to suppress this warning for a given test.  

Maybe we can add another element that allows "attN count validation" behavior to be overridden for the current answer file.  Something like `<attCountCheck>minimal</attCountCheck>` ?   (Naming things is hard.)

sambish5: I think that is something that could be added if that was the desire. I don't know the best way to go about it/if we'd want different logging levels if counts were excessively different, but the basic idea I agree could be helpful in those downstream environments.

drivenflywheel: Another option (maybe better than the new answer file element) would be to add a private member to ExtractionTest that governs this behavior.  If desired, extending test classes could disable/alter the behavior by calling a setting that modifies the value of that private member -  e.g., `setAttachmentCountValidation(newValue);`

sambish5: That I think may end up being a cleaner/easier option in the long run. Would we only want to do this for attachments, or extracts as well?

src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -419,4 +433,13 @@ protected void setupPayload(IBaseDataObject payload, Document doc) {
...
     }
+
+    // allow the validation of att and ext counts to be skipped/not logged in tests
+    protected void setAttachmentCountValidation(boolean skipAttCountValidation) {
+        this.skipAttCountValidation = skipAttCountValidation;
+    }
+
+    protected void setExtractCountValidation(boolean skipExtractCountValidation) {
+        this.skipExtractCountValidation = skipExtractCountValidation;
+    }
```
drivenflywheel, line 444: These parameter names are unexpected, since they're the inversion of what the method name would suggest.  Should rename to clarify
```suggestion
    protected void setAttachmentCountValidation(boolean value) {
        this.skipAttCountValidation = !value;
    }

    protected void setExtractCountValidation(boolean value) {
        this.skipExtractCountValidation = !value;
    }
```

src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -171,8 +174,12 @@ protected void checkAnswers(Document answers, IBaseDataObject payload, List<IBas
     }
 
     protected void checkAnswers(Element el, IBaseDataObject payload, List<IBaseDataObject> attachments, String tname) throws DataConversionException {
-
         int numAtt = JDOMUtil.getChildIntValue(el, "numAttachments");
+        long numAttElements = el.getChildren().stream().filter(c -> c.getName().startsWith("att")).count();
+        if (Math.toIntExact(numAttElements) != numAtt && !(numAttElements <= 0 && numAtt == -1) && !skipAttCountValidation) {
+            logger.warn("Expected numAttachments and actual <att#> count in {} not equal. Expected: {} Actual: {}", tname, Math.max(numAtt, 0),
+                    numAttElements);
```
drivenflywheel, line 181: Now that we've provided a way to disable validation, we should make this an assertion by default.

```suggestion
            if (numAtt >= -1 || numAttElements > 0) {
                if (!skipAttCountValidation) {
                    assertEquals(numAtt, Math.toIntExact(numAttElements), "Expected numAttachments not equal to number of <att#> elements");
                } else {
                    if (Math.toIntExact(numAttElements) != numAtt) {
                        logger.warn("Expected numAttachments and actual <att#> count in {} not equal. Expected: {} Actual: {}", tname,
                            Math.max(numAtt, 0), numAttElements);
                    }
                }
            }
```

drivenflywheel (changes requested):
Last changes? 🤞

src/test/java/emissary/test/core/junit5/ExtractionTest.java
sambish5: so you still want it to log the mismatched counts even with validation skipped? From the previous discussion, it seemed like we wanted the logging to only happen when validation was desired, so that in files where the numbers were super far off, it could be disabled and the warning would not constantly be yelled about.

sambish5:
@drivenflywheel made the suggested changes, had to make some slight changes to the assertion things to make it work  like how it did when it was just logging

drivenflywheel (dismissed):
Looks good.  Sorry for the high number cycles.

sambish5:
rebased and made some _minor_ updates to just include the test name when validation finds expected vs found count not equal on assertion, hopefully this can close this PR 🤞

src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -280,15 +295,29 @@ protected void checkAnswers(Element el, IBaseDataObject payload, @Nullable List<
...
 
         if (payload.hasExtractedRecords()) {
             List<IBaseDataObject> extractedChildren = payload.getExtractedRecords();
             int foundCount = extractedChildren.size();
 
-            if (extractCountStr != null) {
-                assertEquals(Integer.parseInt(extractCountStr), foundCount,
-                        String.format("Number of extracted children in '%s' is %s, not expected %s", tname, foundCount, extractCountStr));
+            if (extractCount > -1) {
+                assertEquals(extractCount, foundCount,
```
drivenflywheel, line 318: Should this get the same `skipExtractCountValidation` treatment as above? (warn vs assert)

sambish5: I could, though that is kind of out of the scope of this PR. The check above this verifies answer files have matching counts, I'm not sure if we want to have the option to "skip validation" when actually checking the answerFile vs the explicit payload extracted records count

drivenflywheel: Ah - you're 100% right.  I lost track of which count the various variables were referencing.

drivenflywheel (dismissed):

dev-mlb (approved):

sambish5:
Changes made by request of @jpdahlke 
Overview of changes:
- no longer fails if `<numAttachments>` & `<extractCount>` DNE `<att#>` or `<extract#>` in answer file, now just logs differences.
- Has a new priority system to make sure that counts in answer file match counts in payload:
       - Checks `<numAttachments>`/`<extractCount>` if present in answer file against count from payload
       - If those are not present, check against `<att#>`/`<extract#>` if present
       - If neither are present, and there are atts or extracts in the payload, **FAIL** the test, saying `<numAttachments>`/`<extractCount>` matching payload count needs to be in answer file

src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -173,8 +173,26 @@ protected void checkAnswers(Element el, IBaseDataObject payload, @Nullable List<
...
+                fail(String.format("%d attachments in payload with no count in answer xml, add matching <numAttachments> count for %s",
+                        attachments.size(), tname));
+            }
+        }
+        // log if <numAttachments> and <att#> don't match in answer file
+        if (numAtt > -1 || numAttElements > 0) {
+            if (Math.toIntExact(numAttElements) != numAtt) {
+                logger.warn("Expected and actual attachment counts in {} not equal. <numAttachments>: {} <att#>: {}", tname,
+                        Math.max(numAtt, 0), numAttElements);
+            }
```
jpdahlke, line 195: I would just remove these log messages I think. Based on the behavior above we expect this scenario and accept it.

jpdahlke (commented):
I think this update makes sense given our integration experiences.

src/test/java/emissary/test/core/junit5/ExtractionTest.java
sambish5: You're right, I'll remove them

dev-mlb (approved):

jpdahlke (approved):

PR #190, opened 2021-12-22
Improved error message on config IMPORT_FILE fail

sambish5:
This is in reference to NationalSecurityAgency/burrito-grande#803.
Updated the error message that is output when an IMPORT_FILE within a config file cannot be found or DNE to make it more clear to Devs what is wrong.

### DIFF
--- a/src/main/java/emissary/config/ServiceConfigGuide.java
+++ b/src/main/java/emissary/config/ServiceConfigGuide.java
@@ -11,6 +11,7 @@ import java.io.Serializable;
 import java.io.StreamTokenizer;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -323,7 +324,10 @@ public class ServiceConfigGuide implements Configurator, Serializable {
                 } catch (IOException e) {
                     // Throw exception if it is an IMPORT_FILE and the base file is not found
                     if ("IMPORT_FILE".equals(parmName) && i == 0) {
-                        throw new IOException("IMPORT_FILE = " + sval + " : Directive failed. Called from " + filename, e);
+                        String importFileName = Paths.get(svalArg).getFileName().toString();
+                        throw new IOException("In " + filename + ", cannot find IMPORT_FILE: " + sval
+                                + " on the specified path. \nMake sure IMPORT_FILE (" + importFileName + ") exists, and the file path is correct.",
+                                e);
                     }
                 }
             }
--- a/src/main/java/emissary/config/ServiceConfigGuide.java
+++ b/src/main/java/emissary/config/ServiceConfigGuide.java
@@ -326,7 +326,7 @@ public class ServiceConfigGuide implements Configurator, Serializable {
                     if ("IMPORT_FILE".equals(parmName) && i == 0) {
                         String importFileName = Paths.get(svalArg).getFileName().toString();
                         throw new IOException("In " + filename + ", cannot find IMPORT_FILE: " + sval
-                                + " on the specified path. \nMake sure IMPORT_FILE (" + importFileName + ") exists, and the file path is correct.",
+                                + " on the specified path. Make sure IMPORT_FILE (" + importFileName + ") exists, and the file path is correct.",
                                 e);
                     }
                 }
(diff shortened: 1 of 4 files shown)


drivenflywheel:
@sambish5 - can you add a unit test to demonstrate/validate the changed behavior?

sambish5:
> 

Yeah can build out 1 or 2.

src/test/java/emissary/config/ServiceConfigGuideTest.java
```diff
@@ -516,6 +516,66 @@ public void testMergeRemoval() throws IOException {
...
+            // will catch as IMPORT_FILE is not created/found, String result will be thrown IO Exception Message
+            importFileFound = false;
+            result = iox.toString();
+        } finally {
+            Files.deleteIfExists(Paths.get(priname));
+            Files.deleteIfExists(Paths.get(impname));
+        }
+
+        // fails, because IMPORT_FILE is not found
+        assertTrue(result, importFileFound);
```
drivenflywheel, line 576: Update this so that it examines the exception message, which is the focus of this PR.  (As currently written, this test would pass even if ServiceConfigGuide hadn't been updated in this PR.)

Can probably get rid of the `importFileFound` boolean. The fact that an IOException was thrown doesn't demonstrate that condition as the cause unless the exception's message is examined, so it's probably redundant.

sambish5: Okay, just pushed a new commit that I think does better of examining the message.

drivenflywheel (approved):
Generally looks good.  A few optional suggestions to consider.

src/main/java/emissary/config/ServiceConfigGuide.java
```diff
@@ -323,7 +324,10 @@ protected ConfigEntry handleNewEntry(final String parmNameArg, final String sval
                 } catch (IOException e) {
                     // Throw exception if it is an IMPORT_FILE and the base file is not found
                     if ("IMPORT_FILE".equals(parmName) && i == 0) {
-                        throw new IOException("IMPORT_FILE = " + sval + " : Directive failed. Called from " + filename, e);
+                        String importFileName = Paths.get(svalArg).getFileName().toString();
+                        throw new IOException("In " + filename + ", cannot find IMPORT_FILE: " + sval
```
jpdahlke, line 328: Out of caution and without testing to be sure, I would shy away from using \n in any logs or exceptions due to the possibility of breaking any expected logstash grok-ing.

jpdahlke (changes requested):

src/main/java/emissary/config/ServiceConfigGuide.java
sambish5: Can quickly go in and remove it. Had it originally for readability ease.

dev-mlb: I don't think this is fixing the issue. In this class we are still throwing the IOException, but with a different message. The ticket say something about how we are not handling the exception correctly. I updated the JsonEscapePlace.cfg file and added IMPORT_FILE = "/does/not/exist" and got the following error:

> 2021-12-28 22:58:56,000 localhost-8001 ERROR emissary.core.Factory -   - Constructor failed
java.lang.reflect.InvocationTargetException: null
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
        at emissary.core.Factory.create(Factory.java:67)
        at emissary.core.Factory.create(Factory.java:122)
        at emissary.admin.PlaceStarter.createPlace(PlaceStarter.java:107)
        at emissary.admin.PlaceStarter.createPlace(PlaceStarter.java:72)
        at emissary.admin.Startup$1.run(Startup.java:371)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.io.IOException: None of the 2 preferences could be found: [emissary.transform.JsonEscapePlace.cfg, JsonEscapePlace.cfg]
        at emissary.config.ConfigUtil.getConfigInfo(ConfigUtil.java:333)
        at emissary.place.ServiceProviderPlace.loadConfigurator(ServiceProviderPlace.java:215)
        at emissary.place.ServiceProviderPlace.loadConfigurator(ServiceProviderPlace.java:193)
        at emissary.place.ServiceProviderPlace.<init>(ServiceProviderPlace.java:139)
        at emissary.transform.JsonEscapePlace.<init>(JsonEscapePlace.java:30)
        ... 10 common frames omitted

The import failed, but it looks like the JsonEscapePlace.cfg does not exist.

dev-mlb (changes requested):

dev-mlb (approved):

PR #202, opened 2022-02-25
Removed code that potentially logs sensitive content in JDOMUtil

andrewbp:


src/main/java/emissary/util/xml/JDOMUtil.java
```diff
@@ -267,7 +267,7 @@ public static int getChildIntValue(final Element el, final String childName) {
         try {
             x = Integer.parseInt(val);
         } catch (NumberFormatException ex) {
-            logger.debug("Unable to parse {} as an integer", val);
+            logger.debug("Unable to parse String as an integer");
```
cfkoehler, line 270: Should these be error since it is an exception?
Also is it worth logging the ex message?

dev-mlb: These were too noisy and changed to debug.

dev-mlb (approved):

arp-0984 (approved):

PR #203, opened 2022-03-07
Built out ability to validate meta field names from xml files against filters using ExtractionTest

sambish5:
This is in reference to NationalSecurityAgency/burrito-grande#968.
This change to ExtractionTest allows for xml meta fields to be compared to filters to validate that meta names exist within filters. 

**Instruction updated below.**
https://github.com/NationalSecurityAgency/emissary/pull/203#issuecomment-1081103992


sambish5:
Updated the validation of the paramaters against the passed in filters. The filters have to be passed through the answer .xml. An example of this is included within TestValidateFieldExtractionTest.xml, but I will describe it here as well.

Two things must be included in the .xml to indicate what filters and meta field names you want checked:
1. In the <answers> section of the .xml, you must include:
        <filters>
    	    <filter1>emissary.output.filter.JsonOutputFilter</filter1>
    	    <filter2>emissary.output.filter.DataFilter</filter2>
    	</filters>
(For Burrito filters, it needs to be formatted as <filter#>com.burrito.output.filters.LogFilter<filter#>)
These allow the filters to be found through an InputStream.
2. For the individual meta field, you must include valdiateField="true" within the meta tag. If this is set to false or not included, the name will not be validated.

src/test/java/emissary/test/core/TestExtractionTest.java
```diff
@@ -47,6 +48,116 @@ public void testCheckStringValueForCollectionFailure() throws JDOMException, IOE
...
+        boolean t2 = test.findFilter("emissary.output.filter.JsonOutputFilter");
+        boolean t3 = test.findFilter("this.should.not.be.found");
+        boolean t4 = test.findFilter("DataFilter");
+
+        // verify boolean "result" from findFilter() is returning correctly
+        Assert.assertTrue("DataFilter should be found.", t1);
+        Assert.assertTrue("JsonOutputFilter should be found.", t2);
+        Assert.assertFalse("This filter should not be found.", t3);
+        Assert.assertFalse("Should return false since path not provided.", t4);
+
```
andrewbp, line 69:
```
 // verify boolean "result" from findFilter() is returning correctly
 Assert.assertTrue("DataFilter should be found.", test.findFilter("emissary.output.filter.DataFilter");
 Assert.assertTrue("JsonOutputFilter should be found.", test.findFilter("emissary.output.filter.JsonOutputFilter"));
 Assert.assertFalse("This filter should not be found.", test.findFilter("this.should.not.be.found"));
 Assert.assertFalse("Should return false since path not provided.",test.findFilter("DataFilter"));
```
Would this be better? Too condensed?

src/main/java/emissary/test/core/ExtractionTest.java
```diff
@@ -214,6 +219,16 @@ protected void checkAnswers(Element el, IBaseDataObject payload, List<IBaseDataO
             assertEquals("Broken status in " + tname, broke, payload.isBroken() ? "true" : "false");
         }
 
+        // Check for Filter to validate against and see if it exists
+        Element filters = el.getChild("filters");
+        if (filters != null && filters.getChildren().size() > 0) {
```
cfkoehler, line 224: I would suggest using isEmpty as a better check.
`!filters.getChildren().size().isEmpty()`

src/main/java/emissary/test/core/ExtractionTest.java
```diff
@@ -290,7 +305,20 @@ protected void checkStringValue(Element meta, String data, String tname) {
         String key = meta.getChildTextTrim("name");
         String value = meta.getChildText("value");
         String matchMode = "equals";
+        String validateField = "false";
```
cfkoehler, line 308: Would it be better to have this as a boolean? 
Then when the vf.getValue() is called it can convert the value to boolean state.

src/main/java/emissary/test/core/ExtractionTest.java
```diff
@@ -330,6 +358,136 @@ protected void checkStringValue(Element meta, String data, String tname) {
...
+     * Method to validate a meta key name against fields/params within Filter .cfg files. Currently, this is set to NOT RUN
+     * unless the meta field specifies it in the .xml file like: meta validateField="true"
+     *
+     * Along with this, filter paths must be passed through the .xml within the answers section. This is demonstrated within
+     * the TestValidateFieldExtractionTest.xml file within resources/emissary/test/core
+     *
+     * @param key the current meta key name to validate against LogFilter
+     * @param tname the current test name
+     */
+    protected void validateFieldInFilter(String key, String tname) {
```
cfkoehler, line 371: Sonar recommends breaking this method up to reduce complexity. Parts can possible be abstracted out into helper methods.
Not a blocker but something to be aware of.

src/main/java/emissary/test/core/ExtractionTest.java
```diff
@@ -330,6 +358,136 @@ protected void checkStringValue(Element meta, String data, String tname) {
...
+                    if ((i + 1) < filterList.size()) {
+                        logger.warn("Error while validating {}", key, e);
+                    } else {
+                        logger.warn("Error while validating {}. Moving to next file.", key, e);
+                    }
+                }
+            }
+        }
+        // if loops through all files and cannot find matching field/param for key, then validation fails
+        if (filterList.size() == 0) {
```
cfkoehler, line 462: Same as above. I think it would be better to use isEmpty instead of checking the size

src/main/java/emissary/test/core/ExtractionTest.java
```diff
@@ -330,6 +358,136 @@ protected void checkStringValue(Element meta, String data, String tname) {
...
+     */
+    protected boolean findFilter(String filterPath) {
+        boolean result = false;
+
+        try {
+            Class.forName(filterPath).getResourceAsStream(filterPath + ".cfg");
+            result = true;
+            filterList.add(filterPath);
+        } catch (ClassNotFoundException e) {
+            // result = false;
```
cfkoehler, line 485: Need to finish the catch here?

src/main/java/emissary/test/core/ExtractionTest.java
```diff
@@ -330,6 +358,136 @@ protected void checkStringValue(Element meta, String data, String tname) {
...
+     * checkAnswers. If filter is found, add to filterList as InputStream
+     *
+     * @param filterPath - filter path passed from xml
+     * @return result - boolean for if filter is found
+     */
+    protected boolean findFilter(String filterPath) {
+        boolean result = false;
+
+        try {
+            Class.forName(filterPath).getResourceAsStream(filterPath + ".cfg");
```
cfkoehler, line 481: I am not very familiar with resource streams but should this try be a 'try-with-resources'

src/test/java/emissary/test/core/TestExtractionTest.java
```diff
@@ -47,6 +48,111 @@ public void testCheckStringValueForCollectionFailure() throws JDOMException, IOE
...
+        // verify only found filters paths are added to filterList, should be 2 in this case
+        Assert.assertEquals("filterList<InputStream> should have size 2.", 2, test.filterList.size());
+    }
+
+    /**
+     * No filter is added to filterList, so validation of meta names should FAIL.
+     */
+    @Test(expected = AssertionError.class)
+    public void testValidateFieldWithNoFilter() throws IOException, JDOMException {
```
cfkoehler, line 73: 'public' can be removed from all the tests

cfkoehler (changes requested):

PR #205, opened 2022-03-10
(Minor) Update GitHub actions workflow

ukengineer925:
Also adds VSCode IDE dir to gitignore, fixes a minor typo in readme

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,6 +1,3 @@
-# This workflow will build a Java project with Maven
-# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
-
 name: Java CI with Maven
 
 on:
@@ -12,19 +9,23 @@ on:
 jobs:
   build:
 
+    name: Standard build on Java ${{ matrix.java }}
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java: ['8', '11']
 
     steps:
-    - uses: actions/checkout@v2
-    - name: Set up JDK 1.8
-      uses: actions/setup-java@v1
+    - name: Checkout repo
+      uses: actions/checkout@v3
+
+    - name: Configure Java & Maven
+      uses: actions/setup-java@v3
       with:
-        java-version: 1.8
-    - name: Cache Maven Dependencies
-      uses: actions/cache@v1
-      with:
-        path: ~/.m2
-        key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
-        restore-keys: ${{ runner.os }}-m2
+        java-version: '${{ matrix.java }}'
+        cache: 'maven'
+        distribution: 'temurin'
+
     - name: Build with Maven
-      run: mvn -B package --file pom.xml
+      run: mvn -B --no-transfer-progress package --file pom.xml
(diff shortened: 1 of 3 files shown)


dev-mlb (approved):

.github/workflows/maven.yml
```diff
@@ -15,16 +12,15 @@ jobs:
...
+
+    - name: Configure Java
+      uses: actions/setup-java@v3
       with:
-        path: ~/.m2
-        key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
-        restore-keys: ${{ runner.os }}-m2
+        java-version: 8
+        cache: 'maven'
+        distribution: temurin
```
jpdahlke, line 23: I had to lookup the usage of single-quotes for these actions and it looks like it just comes down to basic yaml syntax rules of which (of course) there are many thoughts on what "right" is. I don't have a strong preference but I can say that the github action examples I looked through standardize on single quotes for all strings. Not knowing any of this, I would have probably picked no quotes for simplicity. 

Anyway, I got confused because it was done both ways in the PR and the numeric value made me curious if it would work, so I leave it to you to pick a way and we'll try to stick to that convention. This would apply to java-version, cache, and distribution.

jpdahlke (changes requested):

jpdahlke:
This is cool. I was trying to figure out how to determine what runner version we are using since it sounds like we need Actions Runners v2.285.0 or later. Do you know the best way to determine that?

Also, curious if (in a different effort) there is value in trying to add a matrix to the runners to include java 11 support since I don't think we have anything specifically preventing 11 from working right now. 

From the docs:
`
jobs:
  build:
    runs-on: ubuntu-20.04
    strategy:
      matrix:
        java: [ '8', '11', '13', '15' ]
    name: Java ${{ matrix.Java }} sample
    steps:
      - uses: actions/checkout@v2
      - name: Setup java
        uses: actions/setup-java@v2
        with:
          distribution: '<distribution>'
          java-version: ${{ matrix.java }}
      - run: java -cp java HelloWorldApp
`

.github/workflows/maven.yml
ukengineer925: Very fair point - I'll amend to be single quotes throughout.

ukengineer925:
Thanks Jason - see PR 1649 in the other repo too which includes a slightly alternate example for the matrix feature. You're right though in that compiling w/ Java 11 works fine for emissary, so I'll update this PR to include that (and perhaps other versions if they compile OK)

jpdahlke:
> Thanks Jason - see PR 1649 in the other repo too which includes a slightly alternate example for the matrix feature. You're right though in that compiling w/ Java 11 works fine for emissary, so I'll update this PR to include that (and perhaps other versions if they compile OK)

Ah, yes I see. I'd certainly be in favor (provided it works) of locking in 8 and 11 as versions to enforce to help the 11 migration and call that good for now.

I'm not as convinced 16 and 17 are necessary in the immediate, but if it works to combine them in a way to enforce 8/11 and make 16/17 informatory only (which I think is how you were going in 1649) then it might be fine. What sort of timing are we talking here though? I believe these would intend to run max parallel, but it would depend on availability of the runner VMs which I don't have insight into how many resources we have.  I'm not sure we'd want the extra builds to delay the action completion too much.

jpdahlke:
Thanks for making these updates!

PR #213, opened 2022-03-24
Allow '+' character in IBDO Form name

cfkoehler:


drivenflywheel (approved):

fbruton (commented):
Please update the PayloadUtil.isValidForm() JavaDoc to account for the additional valid character.

fbruton (approved):

jpdahlke (changes requested):
We are going to put a hold on this for a bit until we have more information as to handling proper forms. 

We may need to back-pedal and change the strategy to only log an error for forms that contain a "." which was the original issue anyway. 

There is probably a larger discussion needed on the usage of dynamic forms.

src/test/java/emissary/util/PayloadUtilTest.java
```diff
@@ -172,6 +172,7 @@ public void testIsValidForm() {
         assertTrue("Dash and underscore aren't considered valid", PayloadUtil.isValidForm("-_"));
         assertTrue("Parentheses aren't considered valid", PayloadUtil.isValidForm("formName-(suffixInParens)"));
         assertTrue("Slash aren't considered valid", PayloadUtil.isValidForm("formName-(application/xml)"));
+        assertTrue("Plus aren't considered valid", PayloadUtil.isValidForm("formName-(application+xml)"));
         assertFalse("Dot isn't considered valid", PayloadUtil.isValidForm("."));
```
drivenflywheel, line 176: LOL - given that the initial driver for this validation was catching forms that contain a embedded period, it would be nice for our "dot" testing to use a more representative form than `"."`.

cfkoehler:
Discussion points:
- [ ] Should form names be normalized when they are set. For example replace spaces with underscore
- [ ] Set this check to a a warning and not an error since all the separates work except for periods
- [ ] If a period is in the form then that should stay as an error.

PR #216, opened 2022-04-08
add SPILoader and remove other initialization calls

jpdahlke:


fbruton (approved):

dev-mlb (approved):

drivenflywheel (approved):

src/main/java/emissary/spi/SPILoader.java
```diff
@@ -0,0 +1,28 @@
...
+ * <pre>
+ * - Loading is delegated to the {@link java.util.ServiceLoader}.
+ * - Configured via src/main/resources/META-INF/services/emissary.spi.InitializationProvider
+ * </pre>
+ */
+public class SPILoader {
+
+    private static final Logger logger = LoggerFactory.getLogger(SPILoader.class);
+
+    public static void load() {
```
drivenflywheel, line 20: Is there any risk in this loading not being synchronized? I don't think so, but it's possible a future provider could perform work that isn't safe for concurrent execution.

PR #221, opened 2022-04-12
downgrade surefire plugin to 3.0.0-M3

jpdahlke:


jpdahlke:
@flbruto do you remember this issue from our upgrade attempt long ago?

ukengineer925 (approved):

ukengineer925:
FWIW - M4 works fine, it's just M5 that introduces the regression. I think it _could_ be related to this, I tried with both Legacy and Surefire ForkNodeFactory implementations but neither seemed to make a difference: https://maven.apache.org/surefire/maven-surefire-plugin/examples/process-communication.html

fbruton (approved):

PR #226, opened 2022-04-20
Removed content logging within MultiFileUnixCommandPlace

sambish5:
Removed lines with possible content logging within MultiFileUnixCommandPlace.

### DIFF
--- a/src/main/java/emissary/place/MultiFileUnixCommandPlace.java
+++ b/src/main/java/emissary/place/MultiFileUnixCommandPlace.java
@@ -537,7 +537,7 @@ public class MultiFileUnixCommandPlace extends MultiFileServerPlace implements I
         }
 
         if (entries == null || entries.size() == 0) {
-            logger.debug("no messages found in {}", tData.getFilename());
+            logger.debug("no messages found in file.");
             return Collections.emptyList();
         }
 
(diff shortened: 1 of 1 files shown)


cfkoehler (approved):

drivenflywheel (approved):
Would be good to parameterize the many `logger.debug(..)` statements that use String concatenation, but that's a bit outside the original ticket scope.

src/main/java/emissary/place/MultiFileUnixCommandPlace.java
```diff
@@ -685,9 +682,6 @@ protected int processCommand(String[] cmd) {
         StringBuilder outbuf = new StringBuilder();
         StringBuilder errbuf = new StringBuilder();
         int status = processCommand(cmd, outbuf, errbuf);
-        if (outbuf.length() > 0 && logger.isDebugEnabled()) {
-            logger.debug("Discarding process stdout " + outbuf.toString());
-        }
         if (errbuf.length() > 0 && logger.isDebugEnabled()) {
             logger.debug("Discarding process stderr " + errbuf.toString());
```
jpdahlke, line 686: Are we sure the errbuf NEVER has anything we don't want to log?

sambish5: I left it because in the Executrix, the err is defined as "the destination to capture the standard error". From my work with the executrix in PcapPlaceTest, it usually puts out a file name and then a message regarding file format/type or the error (for example, /bin/sh: 1: BadCommand: not found). That being said, if you feel I should remove it, I can.

jpdahlke: I did not know of a scenario where this would be a problem, only feared that it COULD be a problem and by that was ok with just removing it to be sure.

sambish5: Just removed the errbuf logging.

PR #229, opened 2022-04-22
B63 - Remove redundant sorting from output filters

andrewbp:
It was noticed that when data goes through DropOffPlace, it can be sorted twice, which is redundant. This work aims to remove the possibility of that happening.

### DIFF
--- a/src/main/java/emissary/output/filter/AbstractFilter.java
+++ b/src/main/java/emissary/output/filter/AbstractFilter.java
@@ -234,10 +234,6 @@ public abstract class AbstractFilter implements IDropOffFilter {
     @Override
     public int filter(final List<IBaseDataObject> list, final Map<String, Object> params) {
         // Important to process them in order, if not already sorted
-        if (params.get(PRE_SORTED) == null) {
-            Collections.sort(list, new emissary.util.ShortNameComparator()); // unsafe?
-            params.put(PRE_SORTED, Boolean.TRUE);
-        }
 
         int status = 0;
         for (final IBaseDataObject d : list) {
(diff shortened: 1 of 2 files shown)


src/main/java/emissary/output/filter/AbstractFilter.java
```diff
@@ -269,16 +260,6 @@ public int filter(final List<IBaseDataObject> list, final Map<String, Object> pa
...
 
-    /**
-     * The method that all filter have to provide
-     * 
-     * @param payload the payload to run the filter on
-     * @param params map of params
-     * @return status value
-     */
-    @Override
-    public abstract int filter(IBaseDataObject payload, Map<String, Object> params);
```
andrewbp, line 280: This was removed because it is already defined in IDropOffFilter.

src/main/java/emissary/output/filter/AbstractRollableFilter.java
```diff
@@ -157,13 +157,11 @@ protected void manageRoller() {
 
     @Override
     public int filter(final IBaseDataObject payload, final Map<String, Object> params) {
-        params.put(PRE_SORTED, "true");
```
andrewbp, line 160: The map should already have "PRE_SORTED, true" by this point.

andrewbp:
https://github.com/NationalSecurityAgency/burrito-grande/issues/63

arp-0984:
Just a question about where runOutputFilters is called on DropOffPlace.processData line 365. Is the data pre-sorted at this point?If not we should add a call to preFilterHook before runOutputFilters.
Looks good other than that.

arp-0984 (commented):
Just a question about where runOutputFilters is called on DropOffPlace.processData line 365. Is the data pre-sorted at this point?If not we should add a call to preFilterHook before runOutputFilters.
Looks good other than that.

dev-mlb (approved):
My main concerns were the calls to processData, line 305. It's called on line 193, but haveList is set to true and runOutputFilters is never called. The other two calls set haveList to false on lines 249 and 252, but that deals with a single bdo from a non-hdmobileagent and sorting is not needed.

cfkoehler:
I had to rebase. Not sure if that was a mistake in my local environment though. GitHub seems happy so probably just on my end. Just something to check.

cfkoehler (approved):

PR #243, opened 2022-05-11
Removed content logging in DropOffUtil

sambish5:
Removed logging of content and unnecessary logging. Also cleaned up remaining log lines.

### DIFF
--- a/src/main/java/emissary/output/DropOffUtil.java
+++ b/src/main/java/emissary/output/DropOffUtil.java
@@ -705,6 +705,7 @@ public class DropOffUtil {
             fileType = metaData.get(FileTypeCheckParameter.FILETYPE.getFieldName());
         } else if (metaData.containsKey(FileTypeCheckParameter.FINAL_ID.getFieldName())) {
             fileType = metaData.get(FileTypeCheckParameter.FINAL_ID.getFieldName());
+            logger.debug("FINAL_ID FileType is ({})", fileType);
             metaData.put(FileTypeCheckParameter.FILETYPE.getFieldName(), fileType);
         } else {
             if (forms.contains(" ")) {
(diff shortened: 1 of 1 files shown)


arp-0984 (approved):

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -708,7 +705,6 @@ public String getFileType(final Map<String, String> metaData, final String forms
             fileType = metaData.get(FileTypeCheckParameter.FILETYPE.getFieldName());
         } else if (metaData.containsKey(FileTypeCheckParameter.FINAL_ID.getFieldName())) {
             fileType = metaData.get(FileTypeCheckParameter.FINAL_ID.getFieldName());
-            logger.debug("FINAL_ID FileType is ({})", fileType);
```
jpdahlke, line 711: Recommend keeping these

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -978,10 +973,9 @@ public Date extractEventDateFrom(final IBaseDataObject d, final boolean lastReso
             if (value != null) {
                 try {
                     date = emissary.util.TimeUtil.getDateFromISO8601(value);
-                    logger.debug("Successfully parsed event date from " + paramName);
                     return date;
                 } catch (DateTimeParseException ex) {
-                    logger.debug("Cannot parse EventDate " + value);
+                    logger.debug("Cannot parse EventDate", ex);
```
jpdahlke, line 978: Might need to get back to you on this one.

jpdahlke (changes requested):
Awaiting some guidance, but most of these look like good updates to me.

src/main/java/emissary/output/DropOffUtil.java
sambish5: For this one I assume just using the exception message in the log would work rather than calling the exact "value"

jpdahlke (approved):

PR #261, opened 2022-06-15
add git properties file to build dir #879

dev-mlb:


ukengineer925 (approved):
LGTM. I don't think the phase is needed since it's the default, but it does no harm leaving it in. I used this plugin on a different project previously and it was really helpful.

cfkoehler (approved):

PR #265, opened 2022-06-26
io :: updated JsonOutputFilter ignorelist

dev-mlb:


src/main/java/emissary/output/filter/JsonOutputFilter.java
```diff
@@ -130,41 +130,41 @@ protected boolean includeParameter(String key) {
             }
 
             // check the white/black list first
```
andrewbp, line 132: This line seems to have (understandably) escaped the search/replace.

andrewbp (changes requested):

andrewbp (dismissed):

drivenflywheel (dismissed):

src/main/java/emissary/output/filter/JsonOutputFilter.java
```diff
@@ -92,19 +92,19 @@ class IbdoParameterFilter extends SimpleBeanPropertyFilter implements Serializab
         private static final long serialVersionUID = 1L;
```
jpdahlke, line 92: Should we make this inner class static while we are in here?

jpdahlke (approved):

PR #266, opened 2022-06-26
updates based on sonarlint report - admin & analyze

dev-mlb:


src/main/java/emissary/admin/PlaceStarter.java
```diff
@@ -107,48 +108,52 @@ public static IServiceProviderPlace createPlace(final String theLocation, final
...
+                for (final DirectoryEntry entry : entries) {
+                    keys.add(entry.getKey());
+                }
+                logger.info("Forcing removal of {} keys due to failed {}", keys.size(), loc);
+                localDir.removePlaces(keys);
+            } else {
+                logger.debug("Failed {} did not have any directory keys registered", loc);
+            }
+        } catch (EmissaryException ee) {
+            logger.debug("NO local directory, cannot force key dereg for {}", loc);
```
cfkoehler, line 154: Should this and the debug message above be a higher log level? Something we would want to know about in production environment?

src/main/java/emissary/admin/PlaceStarter.java
```diff
@@ -161,7 +166,7 @@ public static IServiceProviderPlace alreadyExists(final String theLocation) {
         final String luStr = theLocation.substring(theLocation.indexOf("//"));
         try {
             final IServiceProviderPlace thePlace = (IServiceProviderPlace) Namespace.lookup(luStr);
-            logger.debug(theLocation + " already running on " + thePlaceHost);
+            logger.debug("{} already running on {}", theLocation, thePlaceHost);
             return thePlace;
         } catch (NamespaceException nse) {
```
cfkoehler, line 171: Should these two catches be combined. Or should we log something?

src/main/java/emissary/admin/PlaceStarter.java
```diff
@@ -82,7 +83,7 @@ public static IServiceProviderPlace createPlace(final String theLocation, final
      * @return the place that was found or created, or null if it can't be done
      */
     public static IServiceProviderPlace createPlace(final String theLocation, final Object[] constructorArgs, final String theClassStr) {
-        logger.debug("Ready to createPlace " + theLocation + " as " + theClassStr);
+        logger.debug("Ready to createPlace {} as {}", theLocation, theClassStr);
 
         final long t1 = System.currentTimeMillis();
```
cfkoehler, line 89: Since this PR is for Sonar and code cleanup should the TODO comment bellow on line 91 be addressed or removed? Another TODO on line 165

cfkoehler (approved):

src/main/java/emissary/admin/PlaceStarter.java
dev-mlb: This is a failure scenario, so the place the failed to start probably logged a stacktrace already. My guess is that this was turned down to debug because it is just overly noisy. We are trying to deregister the place, so we most likely don't need a bunch of stacktraces that we are having issues with deregistering the place that failed to start up.  We can set this to error if you want, but my suggestion would be to submit a separate pr for it. Thoughts?

src/main/java/emissary/admin/PlaceStarter.java
dev-mlb: I was going to handle the TODOs in a separate pr. Since it might change some functionality, maybe a good target for 8??

src/main/java/emissary/admin/PlaceStarter.java
dev-mlb: I don't know why we're catching Throwable here. It's on my list to look into. Agreed, we either need to combine them or remove the Throwable catch. I'm in favor of doing the latter, but I thought it may be a little out of scope for this pr. It shows up in the sonarlint report, so it'll be on my next iteration to cleanup.

PR #267, opened 2022-06-28
updates based on sonarlint report - client

dev-mlb:


src/main/java/emissary/client/response/PlacesResponseEntity.java
```diff
@@ -62,6 +67,6 @@ public void dumpToConsole() {
         for (PlaceList placeList : getCluster()) {
             placeList.dumpToConsole();
         }
-        System.out.println();
+        logger.info("");
```
cfkoehler, line 70: Do we still want an empty message in the logger?

cfkoehler (approved):

src/main/java/emissary/client/HTTPConnectionFactory.java
```diff
@@ -163,7 +162,7 @@ private static char[] loadPW(final String pazz) throws IOException {
         if (pazz == null) {
             return null;
```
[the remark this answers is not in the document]
dev-mlb: I traced the code to a specific implementation and it specifically checked for null. So yeah, for this effort, I did not want to do a full refactor and complicate this pr.

src/main/java/emissary/client/response/PlacesResponseEntity.java
dev-mlb: I don't think we use this output anywhere, but just in case, I wanted the output to be the same. I'm happy to remove it if you can confirm we don't use it anywhere.

PR #278, opened 2022-07-08
tests :: updated fail method calls

dev-mlb:
In junit 4, we needed to move fail calls to throw AssertionError because the stacktrace was not printed. As we've updated to junit 5, we can change the methods back to fail because there is an overloaded method to accept the throwable.

### DIFF
--- a/src/test/java/emissary/place/ServiceProviderPlaceTest.java
+++ b/src/test/java/emissary/place/ServiceProviderPlaceTest.java
@@ -194,7 +194,7 @@ class ServiceProviderPlaceTest extends UnitTest {
 
             mtp.shutDown();
         } catch (IOException iox) {
-            fail("Could not initialize MyStreamConfigedTestPlace with one arg constructor: " + iox);
+            fail("Could not initialize MyStreamConfigedTestPlace with one arg constructor", iox);
         }
     }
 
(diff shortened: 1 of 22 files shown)


andrewbp:
Is this strictly to replace all instances of `org.junit.jupiter.api.Assertions.fail` and not `org.junit.jupiter.api.Assertions.fail`?

andrewbp:
Some classes that may have been missed:

SeviceConfigGuideTest
FactoryTest
WorkBundleTest
MainTest
MultiFileServerPlaceTest
ServiceProviderPlaceTest
EnvTest
HealthCheckActionTest
MetricsActionTest
PoolTest
VersionTest

src/test/java/emissary/place/ServiceProviderPlaceTest.java
```diff
@@ -387,7 +387,7 @@ void testKeyConfigAddKey() {
             assertTrue(proxies.contains("TPROXY"), "Proxies contains original in set");
             assertTrue(proxies.contains("TPROXY2"), "Proxies contains original in set");
         } catch (IOException iox) {
-            fail("Place should have configured with SERVICE_KEY: " + iox.getMessage());
+            fail("Place should have configured with SERVICE_KEY: ", iox);
```
jpdahlke, line 390: This message could be updated like the other 3 to remove the colon.

jpdahlke (approved):

jpdahlke:
> Is this strictly to replace all instances of `org.junit.jupiter.api.Assertions.fail` and not `org.junit.Assert.fail`?

I believe the intent was to update any `fail()` calls that used a throwable in the signature such that:
 - `throw new AssertionError()` calls were converted to use `fail()`
 - Calls to `.getMessage()` are removed and the throwable object used instead

cfkoehler (approved):

PR #285, opened 2022-07-18
elk :: add json output for agents endpoint

dev-mlb:


contrib/elasticsearch/readme.md
```diff
@@ -0,0 +1,56 @@
...
+Helpful setup guide: https://www.elastic.co/guide/en/kibana/8.3/docker.html
+
+Note: These commands are meant to be run from the root of the Emissary project 
+
+## Elasticsearch and Kibana 
+
+Run Elasticsearch:
+```shell
+docker network create elastic
+docker run --name elasticsearch --hostname elasticsearch --net elastic -p 9200:9200 -p 9300:9300 -t docker.elastic.co/elasticsearch/elasticsearch:8.3.2
```
cfkoehler, line 13: It might be best to specify -e discovery.type=single-node for development use. 
```suggestion
docker run --name elasticsearch --hostname elasticsearch --net elastic -p 9200:9200 -p 9300:9300 -e discovery.type=single-node -t docker.elastic.co/elasticsearch/elasticsearch:8.3.2
```

Otherwise many systems will need to have the vm.max_map_count increased from the default system setting.

contrib/elasticsearch/readme.md
```diff
@@ -0,0 +1,56 @@
...
+docker run --name kibana --hostname kibana --net elastic -p 5601:5601 docker.elastic.co/kibana/kibana:8.3.2
+```
+
+Get certs:
+```shell
+mkdir target/tls
+docker cp elasticsearch:/usr/share/elasticsearch/config/certs/http_ca.crt target/tls/
+docker cp elasticsearch:/usr/share/elasticsearch/config/certs/http.p12 target/tls/
+docker exec -it elasticsearch bin/elasticsearch-keystore show xpack.security.http.ssl.keystore.secure_password
+openssl pkcs12 -in target/tls/http.p12 -out target/tls/http.key.pem -nocerts -nodes
```
cfkoehler, line 33: I might be missing something very simple.
What is the password for http.p12? It is not the elastic user password.
Or did you generate your own p12 crts to use?

dev-mlb: Should be the output of the previous command:
```
docker exec -it elasticsearch bin/elasticsearch-keystore show xpack.security.http.ssl.keystore.secure_password
```

dev-mlb: I updated it to use an env var. I was having some trouble with the docker output containing a carriage return (it's probably why I did not do it this way). Let me know if that works.

cfkoehler: Yep it works. Thanks. I missed the previous command.

cfkoehler (approved):

src/test/java/emissary/client/response/AgentsFormatterTest.java
```diff
@@ -0,0 +1,23 @@
+package emissary.client.response;
+
+import static org.junit.jupiter.api.Assertions.*;
```
jpdahlke, line 3: nit: import *

src/main/java/emissary/server/api/Agents.java
```diff
@@ -41,6 +45,18 @@ public Response agents() {
...
+        AgentsFormatter formatter = AgentsFormatter.builder().withHost(entity.getLocal().getHost()).build();
+        StringJoiner joiner = new StringJoiner("\n", "", "\n");
+        entity.getLocal().getAgents().forEach(agent -> joiner.add("{\"agents\":" + formatter.json(agent) + "}"));
+        entity.getErrors().forEach(err -> joiner.add("{\"agents\":" + formatter.json("error", err) + "}"));
+        return Response.ok().entity(joiner.toString()).build();
+    }
+
     @GET
     @Path("/cluster/agents")
     @Produces(MediaType.APPLICATION_JSON)
```
jpdahlke, line 62: I think we might want to put `@Deprecated` on this

src/main/java/emissary/server/api/Agents.java
```diff
@@ -41,6 +45,18 @@ public Response agents() {
...
     }
 
+    @GET
+    @Path("/agents/log")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response agentsLog() {
+        AgentsResponseEntity entity = lookupAgents();
+        AgentsFormatter formatter = AgentsFormatter.builder().withHost(entity.getLocal().getHost()).build();
+        StringJoiner joiner = new StringJoiner("\n", "", "\n");
+        entity.getLocal().getAgents().forEach(agent -> joiner.add("{\"agents\":" + formatter.json(agent) + "}"));
```
jpdahlke, line 55:
```suggestion
        entity.getLocal().getAgents().forEach(agent -> joiner.add("{\"agent\":" + formatter.json(agent) + "}"));
```
The generated output is a list like:
```
{"agents": {"timestamp":..."host":..."name":..."status":...}}
{"agents": {"timestamp":..."host":..."name":..."status":...}} 
{"agents": {"timestamp":..."host":..."name":..."status":...}} 
```
the "agents" structure intuitively to me would expect:
```
{"agents":
{"agent": {"timestamp":..."host":..."name":..."status":...}}
{"agent": {"timestamp":..."host":..."name":..."status":...}} 
{"agent": {"timestamp":..."host":..."name":..."status":...}}
}
```
which i think isn't the "format" we were going for

jpdahlke (changes requested):

jpdahlke:
I see `[object Object`] in the web ui instead of the agent values

src/main/java/emissary/server/api/Agents.java
dev-mlb: Added `@Deprecated` on the old methods. I labeled the new methods as `v2` to distinguish them from the older ones. Let me know if you want it named something else. I didn't want to name it `agents/ng` or something like that. Eventually we'll just remove the v1 methods and use the v2 ones.

jpdahlke: We don't have a /v1/agents/log so I don't know we need a /v2/agents/log for distinction.

seems like:
/agents (old style which calls /v1/agents)
/agents/log (which is new and can just be what's in /v2/agents/log)
/v1/agents
/v2/agents

is all we need?

jpdahlke (changes requested):

jpdahlke (approved):

PR #288, opened 2022-07-27
executrix :: updated to accept data stream

dev-mlb:


src/test/java/emissary/util/shell/ExecutrixTest.java
```diff
@@ -337,6 +337,76 @@ void testExecute() throws IOException {
...
+
+        final byte[] data = "bbb".getBytes();
+
+        final String cmd = "/bin/cat";
+
+        int pstat;
+        final StringBuilder sout = new StringBuilder();
+        final StringBuilder serr = new StringBuilder();
+
+        pstat = this.e.execute(cmd, data);
```
DavidAnDavis2021, line 351: Overall the unit tests look great! You may want to also add a test case when the data object is empty since there is a check for empty data objects.

src/main/java/emissary/util/shell/Executrix.java
```diff
@@ -748,6 +857,14 @@ public int execute(final String[] cmd, final StringBuilder out, final StringBuil
         return exitValue;
     }
 
+    private void streamData(Process p, byte[] data) throws IOException {
+        try (OutputStream os = new BufferedOutputStream(new DataOutputStream(p.getOutputStream()))) {
+            if (ArrayUtils.isNotEmpty(data)) {
+                os.write(data);
```
DavidAnDavis2021, line 863: Will we want to add an os.flush command here? Or will that be automatically handled since this is inside a try-with-object block?

src/main/java/emissary/place/UnixCommandPlace.java
```diff
@@ -171,7 +171,7 @@ protected void logMessages(String tempDir) {
     public byte[] fileProcess(String[] cmd, String outputFile) {
         logger.debug("fileProcess({})", Arrays.asList(cmd));
         StringBuilder errbuf = new StringBuilder();
-        int result = executrix.execute(cmd, null, errbuf);
+        int result = executrix.execute(cmd, (StringBuilder) null, errbuf);
```
DavidAnDavis2021, line 174: Any particular reason for the casting of null to a StringBuilder?

In particular I'm wondering other places or Places may need this change and I want to check this doesn't break compatibility anywhere.

dev-mlb: There are too many overloaded methods now, it wasn't sure which one this was trying to use.

src/main/java/emissary/util/shell/Executrix.java
dev-mlb: I added it

src/test/java/emissary/util/shell/ExecutrixTest.java
dev-mlb: Added an empty data object to the tests

DavidAnDavis2021 (commented):
Tested Emissary code against my sample code with byte input. Nice work!

drewfarris (approved):

jpdahlke (approved):

PR #289, opened 2022-07-28
sonarlint :: fix issues for jdomutil

dev-mlb:


arp-0984 (approved):

src/main/java/emissary/util/xml/AbstractJDOMUtil.java
```diff
@@ -0,0 +1,233 @@
...
+     * Create a JDOM element with some CDATA
+     */
+    public static Element cdataElement(final String name, final String text) {
+        final Element e = new Element(name);
+        e.addContent(new CDATA(text));
+        return e;
+    }
+
+    /**
+     * Create a JDOM element, protectign the data with encoding if needed
```
jpdahlke, line 166:
```suggestion
     * Create a JDOM element, protecting the data with encoding if needed
```

jpdahlke (changes requested):
Kind of a nit, but is there a chance we can re-order/sort the methods in the extending util classes so that it's easier to find the method signatures we want? It was troublesome to diff the two side by side.

jpdahlke (approved):

PR #294, opened 2022-08-08
(Minor) Use native git with commit-id plugin

ukengineer925:
Since JGit doesn't support git worktrees, the commit-id plugin for Maven also doesn't support them. This is a small tweak to tell it to use native git instead. There wasn't any significant performance degradation from what I saw locally.

See issue 215 on that plugin's GH repo for more details.


cfkoehler (approved):
Is there any concern about having different git versions in our build environments? Maybe that is something we should check and normalize. 
Otherwise it looks good to me.

ukengineer925:
> Is there any concern about having different git versions in our build environments? Maybe that is something we should check and normalize. Otherwise it looks good to me.

I don't think so - this really just affects local development when you are trying to build within a worktree. Since we're only really pulling basic info from git, a native instance of any version should suffice.

dev-mlb (approved):

PR #303, opened 2022-08-27
Add configuration of HTTPConnectionFactory with PEM certs

fbruton:


src/main/java/emissary/client/HTTPConnectionFactory.java
```diff
@@ -183,16 +195,48 @@ private static KeyStore buildStore(final String path, final char[] pazz, final S
...
+        } else {
+            try (final InputStream is = Files.newInputStream(filePath)) {
+                keyStore.load(is, pazz);
+            }
         }
         return keyStore;
     }
 
+    protected static boolean isPemCert(Path filePath) throws IOException {
+        DataInputStream reader = new DataInputStream(Files.newInputStream(filePath));
```
dev-mlb, line 215: try-with-resources?

dev-mlb (changes requested):

src/main/java/emissary/client/HTTPConnectionFactory.java
fbruton: Good catch

dev-mlb (approved):

src/main/java/emissary/client/HTTPConnectionFactory.java
```diff
@@ -183,16 +195,50 @@ private static KeyStore buildStore(final String path, final char[] pazz, final S
...
 
+    protected static boolean isPemCert(Path filePath) throws IOException {
+        byte[] beginBytes;
+        try (DataInputStream reader = new DataInputStream(Files.newInputStream(filePath))) {
+            int totalNumBytes = reader.available();
+            if (totalNumBytes <= BEGIN_CERT.length() + END_CERT.length()) {
+                return false;
+            }
+            beginBytes = new byte[BEGIN_CERT.length()];
+            reader.read(beginBytes);
```
cfkoehler, line 222: nit:
Not sure how much this matters in the real world so will defer to you more senior devs. Sonar complains about not checking the return value of the `read` call.
Would a `reader.readFully` make more sense here?
```suggestion
            reader.readFully(beginBytes);
```

cfkoehler (approved):

src/main/java/emissary/client/HTTPConnectionFactory.java
```diff
@@ -183,16 +195,50 @@ private static KeyStore buildStore(final String path, final char[] pazz, final S
...
+        }
```
cfkoehler, line 223: I need to familiarize myself with `try with resources` more but should it still have a catch? Or just leave it up to the calling function to handle?

src/main/java/emissary/client/HTTPConnectionFactory.java
jpdahlke: I think this makes sense, unlikely to matter in practice but simpler code.

src/main/java/emissary/client/HTTPConnectionFactory.java
jpdahlke: Is this in reference to converting to readFully when it throws an IOException or in general? I believe that the calling method would just handle the exception in either case. Once the reader is out of scope of the try it closes (so before any catch/finally).

src/main/java/emissary/client/HTTPConnectionFactory.java
fbruton: Updated

PR #310, opened 2022-09-07
Created GitRepositoryState struct for storing git properties and log Git version during server startup

cfkoehler:
Knowing what version of emissary was actually deployed and running would have been helpful in troubleshooting an issue today.
The goal of this PR was to use the git.properties file that is created by the git-commit-id plugin and store the values in a way that is accessible by the running application. 

I did not create tests yet for GitRepositoryState() If that is something that is deemed necessary then I will do that. 

Sample log message:
```
2022-09-07 23:10:28,537 localhost-8001  INFO emissary.command.ServerCommand -   - Starting Emissary Server - version: 7.9.0-SNAPSHOT - built on 2022-09-07T23:08:38+0000 - git hash: 68d40dc
```

### DIFF
--- a/src/main/java/emissary/Emissary.java
+++ b/src/main/java/emissary/Emissary.java
@@ -30,6 +30,7 @@ import emissary.command.StopCommand;
 import emissary.command.TopologyCommand;
 import emissary.command.VersionCommand;
 import emissary.command.WhatCommand;
+import emissary.util.GitRepositoryState;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -105,6 +106,7 @@ public class Emissary {
             }
             EmissaryCommand cmd = commands.get(commandName);
             dumpBanner(cmd);
+            dumpVersionInfo();
             cmd.run(jc);
             // don't exit(0) here or things like server will not continue to run
         } catch (MissingCommandException e) {
@@ -133,6 +135,10 @@ public class Emissary {
 
     private void dumpBanner() {
         dumpBanner(null);
+    }
+
+    protected void dumpVersionInfo() {
+        LOG.info(GitRepositoryState.dumpVersionInfo(GitRepositoryState.getRepositoryState()));
     }
 
     @VisibleForTesting
--- a/src/main/java/emissary/command/ServerCommand.java
+++ b/src/main/java/emissary/command/ServerCommand.java
@@ -12,6 +12,7 @@ import emissary.command.validator.ServerModeValidator;
 import emissary.core.EmissaryException;
 import emissary.server.EmissaryServer;
 import emissary.server.api.Pause;
+import emissary.util.GitRepositoryState;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -99,7 +100,8 @@ public class ServerCommand extends ServiceCommand {
     @Override
     protected void startService() {
         try {
-            LOG.info("Running Emissary Server");
+            LOG.info("Starting Emissary Server - version: {} - built on {} - git hash: {}", GitRepositoryState.getRepositoryState().getBuildVersion(),
+                    GitRepositoryState.getRepositoryState().getBuildTime(), GitRepositoryState.getRepositoryState().getCommitIdAbbrev());
             new EmissaryServer(this).startServer();
         } catch (EmissaryException e) {
             LOG.error("Unable to start server", e);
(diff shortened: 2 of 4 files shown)


src/main/java/emissary/util/GitRepositoryState.java
```diff
@@ -0,0 +1,147 @@
+package emissary.util;
+
+import java.io.IOException;
+import java.util.Properties;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class GitRepositoryState {
```
cfkoehler, line 9: TODO: Add a javadoc description

src/main/java/emissary/util/GitRepositoryState.java
```diff
@@ -0,0 +1,147 @@
...
+        this.closestTagCommitCount = properties.get("git.closest.tag.commit.count").toString();
+
+        this.buildUserName = properties.get("git.build.user.name").toString();
+        this.buildUserEmail = properties.get("git.build.user.email").toString();
+        this.buildTime = properties.get("git.build.time").toString();
+        this.buildHost = properties.get("git.build.host").toString();
+        this.buildVersion = properties.get("git.build.version").toString();
+    }
+
+    public static GitRepositoryState getRepositoryState() {
```
dev-mlb, line 59: Maybe allow the properties file to be passed in.
```suggestion
    public static GitRepositoryState getRepositoryState(String gitProperties) {
```

dev-mlb (changes requested):

src/main/java/emissary/command/ServerCommand.java
```diff
@@ -99,7 +100,8 @@ public void setupServer() throws EmissaryException {
     @Override
     protected void startService() {
         try {
-            LOG.info("Running Emissary Server");
+            LOG.info("Starting Emissary Server - version: {} - built on {} - git hash: {}", GitRepositoryState.getRepositoryState().getBuildVersion(),
```
dev-mlb, line 103: One last thought - I'm thinking we should move this to a location that is easier to override/update. For instance, we could add a method like dumpVersionInfo and add it after dumpBanner in Emissary.java to get:
```
|  ___|  \/  |_   _/  ___/  ___|/ _ \ | ___ \ \ / /
| |__ | .  . | | | \ `--.\ `--./ /_\ \| |_/ /\ V / 
|  __|| |\/| | | |  `--. \`--. \  _  ||    /  \ /  
| |___| |  | |_| |_/\__/ /\__/ / | | || |\ \  | |  
\____/\_|  |_/\___/\____/\____/\_| |_/\_| \_| \_/  

Emissary version: 7.9.0-SNAPSHOT built on 2022-09-10T10:39:15+0000 git hash: edaa856
```
This way we could output the version info for an implementation pretty easy - something like:
```
\___ /_,___/__,_/_/ /_/ /_/_/____/ ___/

Emissary version: 7.9.0-SNAPSHOT built on 2022-09-10T10:39:15+0000 git hash: edaa856
Implementation version: 0.0.1-SNAPSHOT built on 2022-09-10T11:12:34+0000 git hash: 1234567
``` 
Thoughts?

src/main/java/emissary/Emissary.java
```diff
@@ -105,6 +106,7 @@ protected void execute(String[] args) {
             }
             EmissaryCommand cmd = commands.get(commandName);
             dumpBanner(cmd);
+            LOG.info(GitRepositoryState.dumpVersionInfo(GitRepositoryState.getRepositoryState()));
```
dev-mlb, line 109: Make this line a method, and I think we're good to go. That way in an implementation we can call:
```
@Override
protected void dumpVersionInfo() {
    super.dumpVersionInfo();
    LOG.info(
       GitRepositoryState.dumpVersionInfo(
           GitRepositoryState.getRepositoryState("<implementation>.git.properties")
        )
    );
}
```

dev-mlb (approved):

fbruton (approved):

jpdahlke:
I will say, it's a bit jarring to see my user name come up in the logs. We might want to see how a release goes or maybe remove that bit.

cfkoehler:
> I will say, it's a bit jarring to see my user name come up in the logs. We might want to see how a release goes or maybe remove that bit.

We can over ride the output in different emissary implementations thanks to @dev-mlb feedback.
Though it is handy when deploying a custom build to validate the version running is what is expected as compared to another SNAPSHOT build with the same version number.

PR #313, opened 2022-09-13
Improve HTTPConnectionFactory PEM reader

fbruton:


src/main/java/emissary/client/HTTPConnectionFactory.java
```diff
@@ -86,16 +90,17 @@ public class HTTPConnectionFactory {
     private static final String HTTPS = "https";
     private static final String FILE_PRE = "file://";
     private static final Pattern ENV_VARIABLE_PATTERN = Pattern.compile("\\$\\{(\\w+)}");
+    private static final Pattern CERT_PATTERN = Pattern.compile(
```
drivenflywheel, line 93: CodeQL flags this Pattern as risky, although this is largely mitigated since we control the input files that this will be matched against.  A bounded Pattern such as this would be cleaner.  I chose the minimum and maximum lengths somewhat arbitrarily, but they're conservatively broad while preventing runaway scans.

```
private static final Pattern CERT_PATTERN = Pattern.compile(
"(?:^-----BEGIN CERTIFICATE-----$)([^-]{64,4096})(?:^-----END CERTIFICATE-----$)", Pattern.MULTILINE);
```

fbruton: Updated again to included a named capture group.

drivenflywheel: Ideally, yes.  My snippet was lazier and deferred validation until trying to construct the actual X509Certficate, but it's better to constrain the characters in the Pattern itself.   I think I was just annoyed that I couldn't successfully constrain the length of individual Base64-encoded lines within a multi-line capture group, so I overcorrected when simplifying.   Your change is better.

drivenflywheel (approved):
Looks good.  I've suggested some minor changes, but they're not blockers and can be ignored without hurting my ego.  😄

src/main/java/emissary/client/HTTPConnectionFactory.java
```diff
@@ -203,45 +207,36 @@ private static KeyStore buildStore(final String path, final char[] pazz, final S
...
-        CertificateFactory factory = CertificateFactory.getInstance("X.509");
-        return (X509Certificate) factory.generateCertificate(new ByteArrayInputStream(certBytes));
+        int certCount = 0;
+        int matcherPosition = 0;
+        Matcher matcher = CERT_PATTERN.matcher(pemData);
+        while (matcher.find(matcherPosition)) {
+            byte[] derBytes = DatatypeConverter.parseBase64Binary(matcher.group(BASE_64_ENCODED_DATA).trim());
+            CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
+            X509Certificate x509Certificate = (X509Certificate) certificateFactory.generateCertificate(new ByteArrayInputStream(derBytes));
+            keyStore.setCertificateEntry("cert_" + certCount++, x509Certificate);
```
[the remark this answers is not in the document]
fbruton: There is currently no use-case that requires a more identifiable alias and I don't believe there will be need in the near future.

src/test/java/emissary/client/HTTPConnectionFactoryTest.java
```diff
@@ -90,6 +90,21 @@ void loadPemCertFile() throws Exception {
...
+        addTrustStoreProps(this.cfg);
+
+        this.cfg.removeEntry(CFG_TRUST_STORE, "*");
+        this.cfg.addEntry(CFG_TRUST_STORE, projectBase + "/test-classes/certs/testcertwithcomments.pem");
+
+        final HTTPConnectionFactory instance = new HTTPConnectionFactory(this.cfg);
+
+        final SSLContext fromConfig = instance.build(this.cfg);
+
+        assertNotSame(SSLContext.getDefault(), fromConfig);
```
drivenflywheel, line 105: I'll admit, it wouldn't be awful if we could more directly validate the certificate loading, but that would require the transient TrustStore to be accessible or require us to refactor the KeyStore management into a separate utility class that we could test separately.

fbruton: The long term plan is to move this the KeyStore methods to a utility class. I have no issue with including that in this merge request.

fbruton: Moved PKI functions to a util class and created additional unit-tests.

drivenflywheel (approved):
`PkiUtil.java` could benefit from some cleanup (replace ancient log4j import with slf4j, improve truncated method and variable names, add javadoc comments), but those can be done in a follow up PR.  This works correctly "as is".

dev-mlb (approved):

cfkoehler (approved):

PR #318, opened 2022-09-23
Add github action for auto rebase

cfkoehler:


cfkoehler:
Does not work with forks :(

PR #328, opened 2022-10-07
Fix for routing bug when coordination place adds to transform history

dev-mlb:


dev-mlb:
Definitely far from perfect, but I'm trying to ensure backwards-compatibility. I'm thinking more of these methods are possible candidates for deprecation, but I held off on that for now. I deprecated the setHistory method that takes a list of strings because you lose context of the keys and whether or not they are coordinated.

src/main/java/emissary/core/IBaseDataObject.java
```diff
@@ -653,15 +653,39 @@ static enum MergePolicy {
      *
```
drivenflywheel, line 653: Whenever we add a `@Deprecated` annotation, we should let devs know the preferred alternative so they have a path towards addressing any new warnings they'll see.
```suggestion
     *  @deprecated See {@link #setHistory(TransformHistory)}
     * 
```

src/main/java/emissary/core/BaseDataObject.java
```diff
@@ -565,9 +574,14 @@ public void appendTransformHistory(final String key) {
      * @param newHistory list of new history strings to use
      */
     @Override
```
drivenflywheel, line 576: My preference would be to remove the javadoc comment here so it doesn't override/conflict with the javadoc from the (IBDO) method it implements.  No sense trying to keep 2 copies in sync.

src/test/java/emissary/core/BaseDataObjectTest.java
```diff
@@ -938,6 +938,34 @@ void testVisitHistory() {
...
+        assertNotNull(pen, "Penultimate place");
+        assertEquals("UNKNOWN.FOO.ID.http://host:1234/FooPlace$1010", pen.getFullKey(), "Pen place key");
+
+        assertTrue(this.b.hasVisited("*.FOO.*.*"), "Has visited");
+        assertTrue(this.b.hasVisited("*.BAR_COORDINATION.*.*"), "Has visited");
+        assertFalse(this.b.hasVisited("*.BAZ.*.*"), "Not visited");
+        assertFalse(this.b.hasVisited("*.BAM.*.*"), "Not visited");
+
+        this.b.clearTransformHistory();
+        assertFalse(this.b.hasVisited("*.FOO.*.*"), "Has visited");
```
drivenflywheel, line 965:
```suggestion
        assertFalse(this.b.hasVisited("*.FOO.*.*"), "Has not visited after clear");
```

src/test/java/emissary/core/BaseDataObjectTest.java
```diff
@@ -938,6 +938,34 @@ void testVisitHistory() {
...
+        assertFalse(this.b.hasVisited("*.BAR_COORDINATION.*.*"), "Has no visited after clear");
```
drivenflywheel, line 966:
```suggestion
        assertFalse(this.b.hasVisited("*.BAR_COORDINATION.*.*"), "Has not visited after clear");
```

drivenflywheel (changes requested):
I think this implementation looks great - I added some minor suggestions but they're purely cosmetic.

Could you also add at least one test to `RoutingAlgorithmTest` to demonstrate that coordinated visits don't affect the value returned by the MobileAgent's `getNextKey(place, payload)` call?   My biggest concern with missing that coverage is the risk of a future regression.

dev-mlb:
I still need to add the test to RoutingAlgorithmTest

dev-mlb:
> RoutingAlgorithmTest

Done!

src/test/java/emissary/core/BaseDataObjectTest.java
```diff
@@ -940,30 +940,30 @@ void testVisitHistory() {
...
-        assertNotNull(pen, "Penultimate place");
-        assertEquals("UNKNOWN.FOO.ID.http://host:1234/FooPlace$1010", pen.getFullKey(), "Pen place key");
+        assertNotNull(pen, "Penultimate place directory entry should exist");
+        assertEquals("UNKNOWN.FOO.ID.http://host:1234/FooPlace$1010", pen.getFullKey(), "Penultimate place returned the wrong key");
 
-        assertTrue(this.b.hasVisited("*.FOO.*.*"), "Has visited");
-        assertTrue(this.b.hasVisited("*.BAR_COORDINATION.*.*"), "Has visited");
-        assertFalse(this.b.hasVisited("*.BAZ.*.*"), "Not visited");
-        assertFalse(this.b.hasVisited("*.BAM.*.*"), "Not visited");
+        assertTrue(this.b.hasVisited("*.FOO.*.*"), "Has visited should havev matched for pattern");
```
drivenflywheel, line 959: Almost too small to mention
```suggestion
        assertTrue(this.b.hasVisited("*.FOO.*.*"), "Has visited should have matched for pattern");
```

drivenflywheel (approved):
This looks great.  There's a minor typo in an assertion comment - could probably fix before squashing but it's not enough to merit a separate commit or round of review.

src/test/java/emissary/core/BaseDataObjectTest.java
dev-mlb: haha, figures. fixed!

arp-0984 (approved):

PR #329, opened 2022-10-11
WIP Adding DateTimeFormats class to load date/time formats used in the application

arp-0984:
Initial commit as part of effort to declare date/time formats used by DateTimeFormatter or SimpleDateFormat all in one place. Going forward we can reference the config file instead of re-typing these formats in multiple places.
DateTimeFormats.cfg can be overridden to load other formats.

### DIFF
--- a/src/main/resources/emissary/util/DateTimeFormats.cfg
+++ b/src/main/resources/emissary/util/DateTimeFormats.cfg
@@ -1,0 +1,7 @@
+
+DATE_FORMAT_YEAR = "yyyy"
+DATE_FORMAT_YEAR_DAY = "yyyy-DD"
+
+DATE_TIME_1 = "yyyy MMM dd HH ss mm"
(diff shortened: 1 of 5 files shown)


src/main/resources/emissary/util/DateTimeFormats.cfg
```diff
@@ -0,0 +1,10 @@
+
+
+DATE_1 = "yyyy"
```
andrewbp, line 3: Are we expecting a large list of DateTimeFormats? I was wondering if maybe more descriptive names would be helpful. 

Not really a suggestion, just a thought I had.

arp-0984: I think we could have a large list. For some of these formats for the time being I agree it might make sense to have a more descriptive name. I'll add this.

PR #338, opened 2022-10-25
WIP 1877: Refactored Stage to no longer be a list

andrewbp:
This work aims to restructure Stage to take advantage of Java enums. I'm putting this up as WIP to get initial feedback before continuing down this path.

### DIFF
--- a/src/main/java/emissary/core/Stage.java
+++ b/src/main/java/emissary/core/Stage.java
@@ -1,136 +1,240 @@
 package emissary.core;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 /**
  * Information class about processing stages in the workflow
  */
 public class Stage {
-    // List of stages to use. Use non-default constructors or extend class to change
-    protected static List<String> stages;
+    private StageEnum stageEnum;
 
-    // Each stage must be marked parallel or not
-    protected static List<Boolean> parallel;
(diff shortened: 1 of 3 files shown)


ldhardy:
Everything seems straight forward to me.  I think we should proceed using the enums rather than strings.

ldhardy (approved):

src/main/java/emissary/core/Stage.java
```diff
@@ -1,136 +1,240 @@
...
+                return StageEnum.UNDEFINED;
+        }
+    }
 
-    /**
-     * Return a list of stages
-     */
     public List<String> getStages() {
-        return new ArrayList<String>(stages);
+        List<String> stages = new ArrayList<>();
```
DavidAnDavis2021, line 224: It feels like this list could be calculated in an init function and just returned when called since this list shouldn't change during runtime.

DavidAnDavis2021 (commented):
Overall this looks good. It removes the iterations through the isParallel loop, makes it clearer which index corresponds to which stage.

Only additional suggestion I have beyond the above points is to keep an eye out for unnecessary string comparisons in stages names now that you can just compare the enums.

andrewbp:
Closing this as https://github.com/NationalSecurityAgency/emissary/pull/369 replaces it.

PR #343, opened 2022-11-02
Correct medium security hotspots

cfkoehler:
Address two medium security hotspots. Not really vulnerabilities in the application with the way they are used but it get's flagged on scans. If the changes are not preferred then I can look into a way of suppressing the scans.

DropOffUtil random change: https://rules.sonarsource.com/java/RSPEC-2245
FilePickUp hash change: https://rules.sonarsource.com/java/RSPEC-4790


src/main/java/emissary/pickup/file/FilePickUpClient.java
```diff
@@ -89,7 +89,7 @@ public FilePickUpClient(InputStream configInfo) throws IOException {
      */
     protected void configurePlace() {
         try {
-            digest = MessageDigest.getInstance("MD5");
+            digest = MessageDigest.getInstance("SHA-256");
```
dev-mlb, line 92: Should we make this a configurable value?

cfkoehler: I don't know where we would need to have it different. But it is easy enough to add so will do that.

dev-mlb: If it is not needed, don't worry about it. I didn't know if you'd need to change it internally without having to do a code change.

cfkoehler:
- [ ] Made the most recent commit in Gitub's web interface. Need to pull down and test.

dev-mlb (approved):

jpdahlke (approved):

PR #344, opened 2022-11-02
1927 errorprone fix parameter missing nullable

ldhardy:


src/test/java/emissary/kff/KffMemcachedTest.java
```diff
@@ -8,6 +8,8 @@
 import static org.mockito.Mockito.validateMockitoUsage;
 import static org.mockito.Mockito.when;
 
+import emissary.kff.KffFilter.FilterType;
```
dev-mlb, line 11: Imports are out of order...building locally has changes for all the test files. You may have to import the [formatter.xml](https://github.com/NationalSecurityAgency/emissary/blob/master/contrib/formatter.xml) into your IDE.

dev-mlb (changes requested):

src/test/java/emissary/kff/KffMemcachedTest.java
ldhardy: Should it be the formatter.xml in the emissary contrib folder?  Because when I load that formatter into IntelliJ and reformat a clean copy of the master branch, it modifies a lot of files.

ldhardy: I know we've always had trouble getting all the different IDEs to play nicely together with one format file. I thought that whatever the IDEs were doing was being negated by the maven plugin which would reformat the files consistently.  Maybe I didn't run the maven goal before committing back to my branch.

ldhardy: I added another execution to the maven formatter plugin to handle the test sources.  I think it was already handling formatting the non-test source files.  That might be what was missing.  Hopefully what the IDEs do isn't relevant for the java files at least.  I'll have to figure out how to make IntelliJ match xml files since it seems to be ignoring the formatter.xml.

SemperFidelisTyrannosaurus (approved):

PR #345, opened 2022-11-03
Add environment check to enable object metrics appender

cfkoehler:
We need a way to control what logback appenders are enabled based on the environment that they are running in.
This will enable/disable the object metrics appender based on an environment variable.
I still need to test this more to see if this will work.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -523,7 +523,7 @@
     <dependency>
       <groupId>org.codehaus.janino</groupId>
       <artifactId>janino</artifactId>
-      <version>3.1.8</version>
+      <version>3.1.7</version>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -521,6 +521,11 @@
       <artifactId>xmlbeans</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.codehaus.janino</groupId>
+      <artifactId>janino</artifactId>
+      <version>3.1.8</version>
+    </dependency>
+    <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-plus</artifactId>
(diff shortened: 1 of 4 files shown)


src/main/config/logback.xml
```diff
@@ -20,27 +20,31 @@
...
-        <message>[ignore]</message>
-      </fieldNames>
-    </encoder>
-    <!--<file>logs/${emissary.node.name}-${emissary.node.port}-object-metrics.log</file>
-    <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
-      <fileNamePattern>logs/${emissary.node.name}-${emissary.node.port}-object-metrics.log.%d{yyyy-MM-dd-HH}</fileNamePattern>
-    </rollingPolicy>-->
-  </appender>
+  <!-- Enable output metrics logging by creating an environment variable of "LogObjectMetrics=true" -->
+  <if condition='${LogObjectMetrics} == true'>
```
dev-mlb, line 24: Should we use the env variable in the config to turn on/off the metrics as well?? Something like -
```
OUTPUT_OBJECT_METRICS = "@ENV{'LogObjectMetrics'}"
```

cfkoehler: Forgot that env's can be in configs! I like this idea. No need for custom flavors as well. I will make the change and validate

pom.xml
```diff
@@ -520,6 +520,11 @@
       <groupId>org.apache.xmlbeans</groupId>
       <artifactId>xmlbeans</artifactId>
     </dependency>
+    <dependency>
```
cfkoehler, line 523: Janino is needed: https://logback.qos.ch/manual/configuration.html#conditional
- [ ] Need to check availability

dev-mlb (approved):

jpdahlke (approved):

PR #347, opened 2022-11-04
1938 errorprone fix DefaultCharset

ldhardy:


jpdahlke:
Recommend breaking this (and other prs like this) into parts. One suggestion would be test files, code that is internally facing (configs, commands, pickups), and then distributing other pieces like output filters, input parsers, etc as appropriate. I would not anticipate there being issues, but adding other changes like FileWriter -> Writer would be nice to isolate more.

ldhardy:
> Recommend breaking this (and other prs like this) into parts. 

I will try to do this going forward.

> I would not anticipate there being issues, but adding other changes like FileWriter -> Writer would be nice to isolate more.

Agreed.

ldhardy:
I'm going to create a new branch and PR for these changes because the import sort merge is too tedious to correct manually.

PR #348, opened 2022-11-08
1944 Sorting imports, enforce import sort for checkstyle

ldhardy:
This PR does nothing but the following

add the maven import sorter plugin
sort all the imports
Update the checkstyle config and enforce it on builds for import sorting
Added a maven plugin that automatically sorts the imports during the build, regardless of IDE settings. (thanks Tubbs)

Runs the checkstyle verification during the build by default (without an additional flag like before) to make sure that the impsort is as expected.

Right now the only "failing" violation for checkstyle is import ordering. As you fix other categories that are being flagged by checkstyle, increase their severity to error so that new instances of those violations will fail the build.

Several of my pull requests are being dinged for import sorting issues - but every file in our code base is inconsistent on import sorting so lets get ourselves together here.

### DIFF
--- a/contrib/checkstyle.xml
+++ b/contrib/checkstyle.xml
@@ -140,12 +140,16 @@
     </module>
     <module name="OverloadMethodsDeclarationOrder"/>
     <module name="VariableDeclarationUsageDistance"/>
-    <module name="CustomImportOrder">
-      <property name="severity" value="ignore"/>
-      <property name="customImportOrderRules" value="STATIC###SPECIAL_IMPORTS###THIRD_PARTY_PACKAGE###STANDARD_JAVA_PACKAGE"/>
-      <property name="specialImportsRegExp" value="com.google"/>
-      <property name="sortImportsInGroupAlphabetically" value="true"/>
-      <metadata name="net.sf.eclipsecs.core.lastEnabledSeverity" value="inherit"/>
+    <module name="ImportOrder">
+      <!-- entering an invalid value for severity essential disables this check -->
+      <!-- valid values are ignore, info, warning, error -->
+      <property name="severity" value="error"/>
+      <property name="option" value="bottom"/> <!-- static imports at bottom-->
+      <property name="groups" value="emissary,*,/java.*/"/>
+      <property name="ordered" value="true"/>
+      <property name="separated" value="true"/>
+      <property name="separatedStaticGroups" value="false"/>
+      <property name="sortStaticImportsAlphabetically" value="true"/>
     </module>
     <module name="MethodParamPad"/>
     <module name="OperatorWrap">
--- a/pom.xml
+++ b/pom.xml
@@ -687,10 +687,11 @@
           </configuration>
           <executions>
             <execution>
-              <id>sort-imports</id>
+              <id>sort-java-imports</id>
               <goals>
                 <goal>sort</goal>
               </goals>
+              <phase>generate-sources</phase>
             </execution>
           </executions>
         </plugin>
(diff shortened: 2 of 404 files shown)


pom.xml
```diff
@@ -1343,8 +1359,33 @@
             <artifactId>sortpom-maven-plugin</artifactId>
           </plugin>
           <plugin>
+            <!-- Sort java imports -->
             <groupId>net.revelc.code</groupId>
             <artifactId>impsort-maven-plugin</artifactId>
+            <version>1.6.2</version>
+            <!--TODO update to latest once we move past JDK11 -->
+            <configuration>
+              <groups>*,java</groups>
```
fbruton, line 1368: Recommend adding an import group for emissary:
`<groups>emissary,*,java</groups>`

contrib/checkstyle.xml
```diff
@@ -141,9 +141,10 @@
     <module name="OverloadMethodsDeclarationOrder"/>
     <module name="VariableDeclarationUsageDistance"/>
     <module name="CustomImportOrder">
-      <property name="severity" value="ignore"/>
-      <property name="customImportOrderRules" value="STATIC###SPECIAL_IMPORTS###THIRD_PARTY_PACKAGE###STANDARD_JAVA_PACKAGE"/>
-      <property name="specialImportsRegExp" value="com.google"/>
+      <property name="severity" value="error"/>
+      <property name="customImportOrderRules" value="THIRD_PARTY_PACKAGE###STANDARD_JAVA_PACKAGE###STATIC"/>
```
fbruton, line 145: If we decide to move forward with an `emissary` import goup, suggest the following update:
```      
      <property name="customImportOrderRules" value="SPECIAL_IMPORTS###THIRD_PARTY_PACKAGE###STANDARD_JAVA_PACKAGE###STATIC"/>
      <property name="specialImportsRegExp" value="emissary"/>
      ```

**Review by fbruton — COMMENTED** [review]
Looks good to me. Please squash commits.

**Review by fbruton — CHANGES_REQUESTED** [review]
Looks good to me. Please squash commits.

**Comment by ldhardy** [comment]
I've only ever squashed during the merge.   Isn't the upstream Github project set to squash on merge?

**Review by fbruton — APPROVED** [review]

**Review by dev-mlb — APPROVED** [review]

File: contrib/checkstyle.xml [hunk 3]
```diff
@@ -141,9 +141,11 @@
     <module name="OverloadMethodsDeclarationOrder"/>
     <module name="VariableDeclarationUsageDistance"/>
     <module name="CustomImportOrder">
-      <property name="severity" value="ignore"/>
-      <property name="customImportOrderRules" value="STATIC###SPECIAL_IMPORTS###THIRD_PARTY_PACKAGE###STANDARD_JAVA_PACKAGE"/>
-      <property name="specialImportsRegExp" value="com.google"/>
+      <property name="severity" value="error"/>
```
dev-mlb, line 144: Should we align all projects to have similar properties?? i.e.:
```
      <property name="option" value="above"/>
      <property name="groups" value="emissary,*,java"/>
      <property name="ordered" value="true"/>
      <property name="separated" value="true"/>
      <property name="separatedStaticGroups" value="false"/>
      <property name="sortStaticImportsAlphabetically" value="true"/>
```

dev-mlb (changes requested):

contrib/checkstyle.xml
ldhardy: When you say "all projects" I think you mean just this one and one other project, but I'm not sure that there might be more projects.  What inconsistency are you seeing?  Between the two pull requests for this issue, there should be consistency.

dev-mlb (approved):

ldhardy:
Per the team voting, we've decided to adopt the method implemented here for ordering of imports. The decision has been documented, it can be revisited at some later date if we want to change it in the future.

fbruton (approved):

PR #351, opened 2022-11-13
Create lightweight alpine Dockerfile

cfkoehler:
I needed a lighter-weight Emissary image for a project. I figured I would share.
This is something the team will probably never use. So if we don't want it as part of the main repo I understand and it can just live in my fork :)

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -592,7 +592,7 @@ mvn clean install -Pdist
 docker build -f contrib/docker/Dockerfile . -t emissary
 ```
 
-For a lightweight emissary darker image the Dockerfile.alpine can be used. 
+For a lightweight emissary docker image the Dockerfile.alpine can be used. 
 This image is 1/3 of the size of the normal centOS 7 based image
 ```
 mvn clean install -Pdist
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -595,6 +595,7 @@ docker build -f contrib/docker/Dockerfile . -t emissary
 For a lightweight emissary darker image the Dockerfile.alpine can be used. 
 This image is 1/3 of the size of the normal centOS 7 based image
 ```
+mvn clean install -Pdist
 docker build -f contrib/docker/Dockerfile.alpine . -t emissary-light
 ```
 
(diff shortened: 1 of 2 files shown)


dev-mlb:
I have no issues including it, but we should update the DEVELOPING.md on how to use it (it can be as simple as adding it to the Docker build section as an alternative build).

dev-mlb (approved):

andrewbp:
I ran into this when I tried to test it:
image
Guessing I'd have to try it in a different environment?

andrewbp:
I ran into this when trying to test the image:
image
Guessing I need to try it in a different environment?

cfkoehler:
> I ran into this when trying to test the image: image Guessing I need to try it in a different environment?

Wow. I have never seen that. Do you have a process in the background that is pulling a lot of images from docker hub?
The limit for an anonymous account is 100 in a six hour period (https://www.docker.com/increase-rate-limits/) 

I guess another possible reason is that all docker pulls from our dev environments are going out as if they were from a single user?

DEVELOPING.md
```diff
@@ -592,6 +592,12 @@ mvn clean install -Pdist
 docker build -f contrib/docker/Dockerfile . -t emissary
 ```
 
+For a lightweight emissary darker image the Dockerfile.alpine can be used. 
+This image is 1/3 of the size of the normal centOS 7 based image
+```
+docker build -f contrib/docker/Dockerfile.alpine . -t emissary-light
```
drivenflywheel, line 598: Should include the following line to clarify that this step is still needed regardless of base Dockerfile.  (This just mirrors the instruction steps that use the original Dockerfile):
```suggestion
mvn clean install -Pdist
docker build -f contrib/docker/Dockerfile.alpine . -t emissary-light
```

dev-mlb (approved):

DEVELOPING.md
```diff
@@ -592,6 +592,13 @@ mvn clean install -Pdist
 docker build -f contrib/docker/Dockerfile . -t emissary
 ```
 
+For a lightweight emissary darker image the Dockerfile.alpine can be used.
```
jpdahlke, line 595:
```suggestion
For a lightweight emissary docker image the Dockerfile.alpine can be used. 
```

jpdahlke (approved):

dev-mlb (approved):

PR #355, opened 2022-11-23
Creating a separate legacy class for the DropOffUtil time parsing so it can be replaced in the future

arp-0984:
First change for effort in #1969. Extracts the parsing of date/time formats from DropOffUtil and puts them in their own class. 
Following this change, I will create a new class for the proposed changes to DropOffUtil date/time parsing and I will make it possible to switch between the two classes easily through a config file.


arp-0984:
Closing until analysis is complete or the new class is added.

PR #357, opened 2022-11-25
Stop showing duplicate version with version command

cfkoehler:


dev-mlb (approved):

fbruton (changes requested):
The version is no longer being displayed for other commands, such as `./emissary server`

cfkoehler:
> The version is no longer being displayed for other commands, such as `./emissary server`

I corrected this. I was checking the list of all commands if a command existed and not the args that were given.

PR #295, opened 2022-08-09
Changing SimpleDateFormat usages in DropOffUtil to DateTimeFormatter

arp-0984:
Upgrading to newer date library in DropOffUtil in order to resolve issue.

The one thing I noticed with going from SimpleDateFormat -> DateTimeFormatter is that we lose lenient pattern matching; SimpleDateFormat matches a pattern even if the commas and spacing are not exactly what we expect. With DateTimeFormatter, surrounding a pattern in '[ ]' makes it an optional part of the pattern, so I added optional commas and made spaces optional in many portions of the patterns in order to address this.

PR #358, opened 2022-12-01
Creating a separate legacy class for the DropOffUtil time parsing so it can be replaced in the future

arp-0984:
This pull request has no functional impact. The goals of this PR are the following:

- I have extracted the existing date parsing functionality from DropOffUtil and put it in its own class DateTimeFormatParserLegacy
- I put my proposed date formatting changes from PR #295 in its own class DateTimeFormatParser. This class is disabled at this time, but I have added it so I can easily test with it.
- I made configurable the date parsing class to use in DropOffUtil with options between DateTimeFormatParserLegacy, DateTimeFormatParser, and FlexibleDateTimeParser.


src/main/java/emissary/util/DateTimeFormatParser.java
```diff
@@ -0,0 +1,211 @@
...
+
+        Configurator configG;
+        try {
+            configG = emissary.config.ConfigUtil.getConfigInfo(DateTimeFormatParser.class);
+        } catch (IOException e) {
+            logger.error("Cannot open default config file", e);
+            return;
+        }
+
+        for (final String dfentry : configG.findEntries("DATE_TIME_ZONE_FORMAT")) {
```
andrewbp, line 44: I made this change to lessen the amount of code, seems to work. Could maybe even be further lessened with a Map<String, List>?

```suggestion
protected static void configure() {

        Configurator configG;
        try {
            configG = emissary.config.ConfigUtil.getConfigInfo(DateTimeFormatParser.class);
        } catch (IOException e) {
            logger.error("Cannot open default config file", e);
            return;
        }

        loadDateTimeEntries(configG, "DATE_TIME_ZONE_FORMAT", dateTimeZoneFormats);
        loadDateTimeEntries(configG, "DATE_TIME_OFFSET_FORMAT", dateTimeOffsetFormats);
        loadDateTimeEntries(configG, "DATE_TIME_FORMAT", dateTimeFormats);
        loadDateTimeEntries(configG, "DATE_FORMAT", dateFormats);

    }

    private static void loadDateTimeEntries(Configurator configG, String entryType, List<DateTimeFormatter> dateFormats) {
        for (final String dfentry : configG.findEntries(entryType)) {
            try {
                DateTimeFormatter initialSdf =
                        new DateTimeFormatterBuilder().parseCaseInsensitive().appendPattern(dfentry).toFormatter();
                if (entryType.equals("DATE_TIME_ZONE_FORMAT")) {
                    initialSdf = initialSdf.withZone(GMT);
                }
                final DateTimeFormatter sdf = initialSdf;
                dateFormats.add(sdf);
            } catch (Exception ex) {
                logger.debug("{} entry '{}' cannot be parsed", entryType, dfentry, ex);
            }
        }
        logger.debug("Loaded {} {} entries", dateTimeZoneFormats.size(), entryType);
    }        
```

src/main/java/emissary/util/DateTimeFormatParser.java
```diff
@@ -0,0 +1,211 @@
...
+
+    private static final ZoneId GMT = ZoneId.of("GMT");
+
+    protected static final Logger logger = LoggerFactory.getLogger(DateTimeFormatParserLegacy.class);
+
+    static {
+        configure();
+    }
+
+    protected static void configure() {
```
andrewbp, line 34: I made this change to lessen the amount of code, seems to work. Could maybe even be further lessened with a Map<String, List>?

```suggestion
protected static void configure() {

        Configurator configG;
        try {
            configG = emissary.config.ConfigUtil.getConfigInfo(DateTimeFormatParser.class);
        } catch (IOException e) {
            logger.error("Cannot open default config file", e);
            return;
        }

        loadDateTimeEntries(configG, "DATE_TIME_ZONE_FORMAT", dateTimeZoneFormats);
        loadDateTimeEntries(configG, "DATE_TIME_OFFSET_FORMAT", dateTimeOffsetFormats);
        loadDateTimeEntries(configG, "DATE_TIME_FORMAT", dateTimeFormats);
        loadDateTimeEntries(configG, "DATE_FORMAT", dateFormats);

    }

    private static void loadDateTimeEntries(Configurator configG, String entryType, List<DateTimeFormatter> dateFormats) {
        for (final String dfentry : configG.findEntries(entryType)) {
            try {
                DateTimeFormatter initialSdf =
                        new DateTimeFormatterBuilder().parseCaseInsensitive().appendPattern(dfentry).toFormatter();
                if (entryType.equals("DATE_TIME_ZONE_FORMAT")) {
                    initialSdf = initialSdf.withZone(GMT);
                }
                final DateTimeFormatter sdf = initialSdf;
                dateFormats.add(sdf);
            } catch (Exception ex) {
                logger.debug("{} entry '{}' cannot be parsed", entryType, dfentry, ex);
            }
        }
        logger.debug("Loaded {} {} entries", dateTimeZoneFormats.size(), entryType);
    }        
```

src/main/java/emissary/util/DateTimeFormatParser.java
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * Parse an RFC-822 Date or one of the thousands of variants make a quick attempt to normalize the timezone information
+     * and get the timestamp in GMT. Should change to pass in a default from the U124 header
+     *
+     * @param dateString the string date from the RFC 822 Date header
+     * @param supplyDefaultOnBad when true use current date if sentDate is unparseable
+     * @return the GMT time of the event or NOW if unparseable, or null if supplyDefaultOnBad is false
+     */
+    public static LocalDateTime parseDate(final String dateString, final boolean supplyDefaultOnBad) {
```
andrewbp, line 95: I was trying to avoid having to check for date == null more times than necessary and maybe add some readability. I don't know if this does anything for readability, but I'll share it anyway.
```suggestion
    public static LocalDateTime parseDate(final String dateString, final boolean supplyDefaultOnBad) {
```
    public static LocalDateTime parseDate(final String dateString, final boolean supplyDefaultOnBad) {
        LocalDateTime date = null;

        if (dateString != null && dateString.length() > 0) {
            // Take it apart and stick it back together to
            // get rid of multiple contiguous spaces
            String instr = cleanDateString(dateString);

            for (String entry : entrytypes) {
                switch (entry) {
                    case "DATE_TIME_ZONE_FORMAT":
                        date = tryParseWithDateTimeZoneFormats(instr);
                        break;
                    case "DATE_TIME_OFFSET_FORMAT":
                        date = tryParseWithDateTimeOffsetFormats(instr);
                        break;
                    case "DATE_TIME_FORMAT":
                        date = tryParseWithDateTimeFormats(instr);
                        break;
                    case "DATE_FORMAT":
                        date = tryParseWithDateFormats(instr);
                        break;
                    default:
                        logger.error("Invalid DateTimeFormat entry: {}", entry);
                }
                if (date != null) {
                    break;
                }
            }
            if (date == null) {
                try {
                    date = Instant.from(DateTimeFormatter.ISO_INSTANT.parse(instr)).atZone(GMT).toLocalDateTime();
                } catch (DateTimeParseException e) {
                    // ignore
                }
            }
            // Use the default if required
        }
        if (date == null && supplyDefaultOnBad) {
            date = LocalDateTime.now();
        }
        return date;
    }

    private static String cleanDateString(String dateString) {
        String instr = dateString.replaceAll("\t+", " "); // tabs
        instr = instr.replaceAll("[ ]+", " "); // multiple spaces
        instr = instr.replaceAll("=0D$", ""); // common qp'ified ending
        return instr;
    }

andrewbp: I understand the Legacy class will ultimately go away, but should it still have its own tests in the meantime?

arp-0984: I can add the same tests that I added for the new class. Now that I think about it, it's probably a good idea to help assess the differences in results between the three classes.

arp-0984: Attempted that and everything failed lol. Which makes sense because the number of formats the legacy class was designed can handle is much more limited. Tests the for legacy class will be handled elsewhere.

andrewbp: > Attempted that and everything failed lol. Which makes sense because the number of formats the legacy class was designed can handle is much more limited. Tests the for legacy class will be handled elsewhere.

Sounds good!

src/main/java/emissary/util/DateTimeFormatParser.java
arp-0984: Added this helper function

src/main/java/emissary/util/DateTimeFormatParser.java
arp-0984: For the comment about potentially refactoring wrt the null checks, I spent a little time considering solutions. I refactored so some of the null checks were no longer necessary, and tried using return statements in some places. Let me know what you think, I was struggling to find the cleanest way of doing this.

andrewbp: I like this!

andrewbp (approved):

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -973,6 +962,15 @@ public Date getEventDate(final IBaseDataObject d, @Nullable final IBaseDataObjec
         return eventDate;
     }
 
+    /**
+     * Getter for dateParser
+     * 
+     * @return the date parser to use
+     */
+    public String getDateParser() {
```
dev-mlb, line 970: Not sure I understand the return type of this method being a string. If I want a date parser, then I'm going to get something like `LEGACY`, so what do I do with it? Is there a method somewhere to translate that to an actual date parser?

dev-mlb (changes requested):

src/main/java/emissary/util/DateTimeFormatParser.java
```diff
@@ -0,0 +1,206 @@
...
+     *
+     * @param instr the string to attempt to format
+     * @return the LocalDateTime object if a formatter worked, or null otherwise
+     */
+    private static LocalDateTime tryParseWithDateTimeZoneFormats(final String instr) {
+        // formats with a time zone
+        for (final DateTimeFormatter dtf : dateTimeZoneFormats) {
+            try {
+                ZonedDateTime zdt = ZonedDateTime.parse(instr, dtf);
+                return ZonedDateTime.ofInstant(zdt.toInstant(), GMT).toLocalDateTime();
```
jpdahlke, line 143: It seems like this normalizes to a local time zone? Would we want to add a default time zone for the cases where there is not one? We could assume local or we could assume GMT or really any other depending on the configuration?

jpdahlke (changes requested):
Test coverage looks good on the DateTimeFormatParser class. Curious if we need anything quick and dirty to cover the legacy version.

We talked through some things wrt to DropOffUtil class and how to handle future updates.

I might have a chance to try my hand to clean up the "if !null return date" logic in parseDate if you don't get to it first. It might be good enough for now.

src/main/java/emissary/util/DateTimeFormatParser.java
```diff
@@ -0,0 +1,206 @@
...
+     */
+    private static void loadDateTimeEntries(Configurator configG, String entryType, List<DateTimeFormatter> dateFormats) {
+        for (final String dfentry : configG.findEntries(entryType)) {
+            try {
+                DateTimeFormatter initialSdf =
+                        new DateTimeFormatterBuilder().parseCaseInsensitive().appendPattern(dfentry).toFormatter();
+                if (entryType.equals(DATE_TIME_ZONE_FORMAT)) {
+                    initialSdf = initialSdf.withZone(GMT);
+                }
+                final DateTimeFormatter sdf = initialSdf;
```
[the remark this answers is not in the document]
arp-0984: I think that's exactly where it came from. I'll rename it to something better.

src/test/java/emissary/util/DateTimeFormatParserTest.java
```diff
@@ -0,0 +1,387 @@
...
+        test("2016004", EXPECTED_NO_TIME);
+    }
+
+    @Test
+    void parse_yyyy_DDD() {
+        test("2016-004", EXPECTED_NO_TIME);
+    }
+
+    /*
+     * @Test void testCleanDateString() { test("2016-01-04 18:20<br>", EXPECTED_NO_SECS, "HTML");
```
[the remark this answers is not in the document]
arp-0984: Since all the tests in this class are copied over from FlexibleDateTimeParser, this was one of the copied over tests that did not pass with the new code. I'll mark it as disabled.

src/main/java/emissary/util/DateTimeFormatParser.java
arp-0984: I can add a configurable option to pick a time zone to assume or use local time.

src/main/java/emissary/util/DateTimeFormatParser.java
jpdahlke: Here's a suggestion. I'm sure there is a way to condense this via lambda, but this is fairly readable.
```
            List<Function<String, LocalDateTime>> methodList = new ArrayList();
            methodList.add(date -> tryParseWithDateTimeZoneFormats(date));
            methodList.add(date -> tryParseWithDateTimeOffsetFormats(date));
            methodList.add(date -> tryParseWithDateTimeFormats(date));
            methodList.add(date -> tryParseWithDateFormats(date));

            for (Function<String, LocalDateTime> method : methodList) {
                LocalDateTime date = method.apply(instr);
                if (date != null) {
                    return date;
                }
            }
```

src/main/java/emissary/util/DateTimeFormatParser.java
arp-0984: I checked the java documentation and tested myself to confirm: toLocalDateTime should just convert it into a LocalDateTime object and not convert to a local time zone. 
https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html#toLocalDateTime--
But I will add the option to choose a time zone to be configurable.

arp-0984:
Updated to address review comments.

dev-mlb (approved):

jpdahlke (approved):

PR #362, opened 2022-12-07
KFFHashPlace should prefer byte[] to SeekableByteChannelFactory

drivenflywheel:
KffHashPlace hashing should attempt to use the BDO's byte array before trying to use its SeekableByteChannelFactory, given that the vast majority of BDOs will have a byte array but no channel factory for the foreseeable future


cfkoehler (approved):

fbruton (changes requested):
I recommend `KffHashPlace.useSbc` default to  false as well.

jdcove2 (dismissed):
I believe this change requires discussion.

cfkoehler:
Yeah. I support moving this to false in the code. Right now if someone goes to integrate this version of emissary then behavior is changing. The large file support was supposed to be a no op change to start with. Follow on software is going to need to add this config to disable.

Generally new/changing functionality is defaulted to off or false.

drivenflywheel:
> I recommend `KffHashPlace.useSbc` default to false as well.

Done

src/main/java/emissary/place/KffHashPlace.java
```diff
@@ -22,7 +22,7 @@ public class KffHashPlace extends ServiceProviderPlace {
 
     public static final String SKIP_KFF_HASH = "SKIP_KFF_HASH";
 
-    private boolean useSbc = true;
+    private boolean useSbc = false;
```
cfkoehler, line 25: Won't that get overwritten by line 54 if the entry is not in the config?

drivenflywheel: Addressed

cfkoehler (approved):

fbruton (approved):

drivenflywheel:
(Sorry - accidentally re-requested review because I hadn't refreshed a browser tab.)

PR #365, opened 2022-12-08
Fix flaky tests in PlaceComparisonHelperTest

ukengineer925:
These tests turned out to be slightly flaky in rare occurrences, believe the underlying cause has now been addressed.

### DIFF
--- a/src/test/java/emissary/util/PlaceComparisonHelperTest.java
+++ b/src/test/java/emissary/util/PlaceComparisonHelperTest.java
@@ -1,10 +1,12 @@
 package emissary.util;
 
+import emissary.config.ConfigUtil;
+import emissary.config.Configurator;
 import emissary.core.BaseDataObject;
 import emissary.core.IBaseDataObject;
 import emissary.place.ServiceProviderPlace;
+import emissary.test.core.junit5.UnitTest;
 
-import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 import java.io.ByteArrayInputStream;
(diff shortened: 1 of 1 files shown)


cfkoehler:
Since the place comparison is not needed or used. Should all classes relating to that be removed until it is ready? We should not add code that does not have testing.

jpdahlke:
Would just adding a TODO and removing the `@Test` help? I'm not sure how close these tests are to working or if you really just wanted to clean the slate and start with a new approach. Either way is fine.

ukengineer925:
Thanks both - I've just pushed up a new revision which I believe fully addresses the causes of the flakiness (it was to do with config during place initialisation, we weren't properly extending UnitTest which takes care of this).

src/test/java/emissary/util/PlaceComparisonHelperTest.java
```diff
@@ -13,56 +15,65 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+class PlaceComparisonHelperTest extends UnitTest {
+
+    private static final Class<NullPointerException> NPE = NullPointerException.class;
+
+    private static final byte[] configurationBytes = new StringBuilder()
+            .append("PLACE_NAME = \"TesterPlace\"")
```
drivenflywheel, line 27: Don't these need newlines in between the various entries? This is currently generating one long line of muck.

drivenflywheel: > Don't these need newlines in between the various entries? This is currently generating one long line of muck:
> 
> ```
> PLACE_NAME = "TesterPlace"SERVICE_NAME = "TESTER"SERVICE_TYPE = "TRANSFORM"SERVICE_DESCRIPTION = "Place defined for testing"SERVICE_COST = 50SERVICE_QUALITY = 50SERVICE_PROXY = "UNKNOWN"
> ```

ConfigUtil parses the inputStream into separate `ConfigEntry` values just fine as is.

drivenflywheel (approved):
Would be good to have a unit test demonstrating that `PlaceComparisonHelper.getPlaceToCompare` can return a valid place, but I don't consider that a blocker

src/test/java/emissary/util/PlaceComparisonHelperTest.java
```diff
@@ -13,56 +15,65 @@
...
+            .append("SERVICE_NAME = \"TESTER\"")
+            .append("SERVICE_TYPE = \"TRANSFORM\"")
+            .append("SERVICE_DESCRIPTION = \"Place defined for testing\"")
+            .append("SERVICE_COST = 50")
+            .append("SERVICE_QUALITY = 50")
+            .append("SERVICE_PROXY = \"UNKNOWN\"").toString().getBytes(StandardCharsets.UTF_8);
+
+    @Test
+    void testGetPlaceToCompareArguments() throws Exception {
+        assertNull(PlaceComparisonHelper.getPlaceToCompare(null));
```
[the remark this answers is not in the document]
ukengineer925: Agreed, will add this in a future PR.

jpdahlke (approved):

PR #370, opened 2022-12-13
Publish multipass

cfkoehler:


pom.xml
```diff
@@ -22,7 +22,20 @@
     <url>https://github.com/NationalSecurityAgency/emissary/issues</url>
   </issueManagement>
   <ciManagement />
-  <distributionManagement />
+  <distributionManagement>
+    <repository>
+      <uniqueVersion>false</uniqueVersion>
+      <id>multipass</id>
+      <name>Multipass Release</name>
+      <url>https://multipass.revelc.net/nexus/content/repositories/releases/</url>
```
dev-mlb, line 30: Internally, would specifying this screw us up in any way? Should we make this configurable, as in something like ${distrib.release.url}? Could the job inject this somehow?

cfkoehler: I might need to rethink the approach here. I think ideally we might just publish to OSS Sonar type and might add that change to this pr: https://github.com/NationalSecurityAgency/emissary/pull/372

Ideally I think we want to get to the point where we don't need to build and release internally. We just pull in the published artifact.

PR #374, opened 2022-12-15
1938 ErrorProne fix DefaultCharset

ldhardy:
Apply error prone patches for [DefaultCharset](https://errorprone.info/bugpattern/DefaultCharset).

Jenkins build for this commit hash: https://multipass.revelc.net/jenkins/job/ldh-emissary/19/

### DIFF
--- a/src/main/java/emissary/command/Banner.java
+++ b/src/main/java/emissary/command/Banner.java
@@ -5,6 +5,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 public class Banner {
 
     static String DEFAULT_TEXT = " ________  ________ _____ _____  ___  ________   __\n"
@@ -21,7 +23,7 @@ public class Banner {
         } else {
             String bText = DEFAULT_TEXT;
             try {
-                BufferedReader reader = new BufferedReader(new InputStreamReader(in));
+                BufferedReader reader = new BufferedReader(new InputStreamReader(in, UTF_8));
                 StringBuilder result = new StringBuilder();
                 String line;
                 while ((line = reader.readLine()) != null) {
--- a/src/main/java/emissary/command/WhatCommand.java
+++ b/src/main/java/emissary/command/WhatCommand.java
@@ -39,6 +39,7 @@ import javax.annotation.Nullable;
 
 import static emissary.core.Form.HTML;
 import static emissary.core.Form.TEXT;
+import static java.nio.charset.StandardCharsets.UTF_8;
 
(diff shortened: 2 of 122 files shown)


cfkoehler (changes requested):
Looks good except for needing a rebase.
Will go through on a second pass after that.

ldhardy:
The jenkins build for the rebased commit:  https://multipass.revelc.net/jenkins/me/my-views/view/Lisa's%20View/job/ldh-emissary/23/

cfkoehler (approved):
I am good with this. I change the checks for errorProne on both the master and this branch and the diff shows that all problems were resolved.

With that being said is there an intent to now set `-Xep:DefaultCharset:ERROR` ? If so should that be done in this PR? Not sure about the order of changes here. Another thing is are we expecting CI jobs to run with the errprProne profile eventually? I think that would be ideal because otherwise we can get the drift and have to repeat these efforts again. 

Otherwise thank you @ldhardy for pushing forward on this work! We have exciting days ahead with clean and repeatable builds.

jpdahlke (changes requested):
Looking through the usages of charsets it seems like this could get us into a bind since today we are leveraging the platform default charset which just happens to be UTF-8 and also specifying inputs and outputs to be UTF-8. So I think it's true that forcing everything to UTF-8 would have no impact with the way things are currently configured and expected, but if the platform config was changed to a different encoding, we could not enforce everything to be UTF-8. I think it would have to properly utilize the platform default. 

It seems like there are cases where it would absolutely make sense to force UTF-8 (inside of a .toUtf8() call or when outputting something specifically in utf-8 format) but for the most part I think we would want to leverage the platform default instead for most of these which should be what's happening today.

If we just convert all of these into using `Charset.defaultCharset()`, that theoretically just codifies current behavior and probably won't break our integration testing, but doesn't actually address the issue of correctly applying a charset. To do that we would need a more targeted approach to pick which scenarios to force and which to leave up to the platform.

Of course, flipping the platform from utf-8 to utf-16 today seems to be a bit of a non-starter which indicates to me we certainly have a problem to solve wrt encodings, I just think we should be careful not to mask the behavior and create a sneaky issue further down the road.

A few steps to get there could be:
- move the FileWriter -> BufferedWriter changes to a separate effort
- identify which places we can justify forcing UTF-8 and correct them as one effort
- convert platform default usages to explicitly utilize defaultEncoding() in another effort
- ensure that previous StandardCharsets.UTF-8 calls are properly applied (they might need an effort to convert to defaultEncoding() instead)
- migrate StandardCharsets.UTF-8 calls to the static variety for consistency (this could be intertwined with other changes as we go)
- identify how to correctly configure the platform default and propagate that through

PR #378, opened 2022-12-20
Initial pass at adding generics to the SCBF Framework

drivenflywheel:
Trial approach for review/consideration.  Much like the initial PR that introduced the SCBF framework, the potential value of the changes won't really be evident until integrated in a downstream code base.  Right now this looks like a lot of angle brackets and ceremony with no obvious benefit. That may also prove to be the case once a trial integration is evaluated, but I'm hopeful these changes will enable some more explicitly constrained interfaces without getting in the way.

### DIFF
--- a/src/main/java/emissary/core/BaseDataObject.java
+++ b/src/main/java/emissary/core/BaseDataObject.java
@@ -180,7 +180,7 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
     /**
      * A factory to create channels for the referenced data.
      */
-    protected SeekableByteChannelFactory seekableByteChannelFactory;
+    protected SeekableByteChannelFactory<?> seekableByteChannelFactory;
 
     protected enum DataState {
         NO_DATA, CHANNEL_ONLY, BYTE_ARRAY_ONLY, BYTE_ARRAY_AND_CHANNEL
@@ -319,11 +319,13 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
 
     /**
      * Set the byte channel factory using whichever implementation is providing access to the data.
-     * 
+     *
      * Setting this will null out {@link #theData}
+     * 
+     * @param sbcf the new channel factory to set on this object
      */
     @Override
-    public void setChannelFactory(final SeekableByteChannelFactory sbcf) {
+    public void setChannelFactory(final SeekableByteChannelFactory<?> sbcf) {
         Validate.notNull(sbcf, "Required: SeekableByteChannelFactory not null");
         this.theData = null;
         this.seekableByteChannelFactory = sbcf;
@@ -336,7 +338,7 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
      * @return the factory containing the data reference or the data wrapped in a new factory
      */
     @Override
-    public SeekableByteChannelFactory getChannelFactory() {
+    public SeekableByteChannelFactory<?> getChannelFactory() {
         switch (getDataState()) {
             case BYTE_ARRAY_AND_CHANNEL:
                 throw new IllegalStateException(String.format(INVALID_STATE_MSG, shortName()));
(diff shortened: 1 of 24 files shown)


src/main/java/emissary/core/channels/ImmutableChannel.java
```diff
@@ -0,0 +1,72 @@
...
+import java.nio.ByteBuffer;
+import java.nio.channels.NonWritableChannelException;
+import java.nio.channels.SeekableByteChannel;
+
+/**
+ * Decorator class that wraps an existing {@link SeekableByteChannel} instance and makes it immutable
+ * 
+ * @param <T> Type of SeekableByteChannel to wrap
+ */
+public final class ImmutableChannel<T extends SeekableByteChannel> implements SeekableByteChannel {
```
drivenflywheel, line 15: This file demonstrates one of the cases for which I broke some of the existing encapsulation by taking a previously `private static` inner class and making it public class.

src/main/java/emissary/core/BaseDataObject.java
```diff
@@ -180,7 +180,7 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
     /**
      * A factory to create channels for the referenced data.
      */
-    protected SeekableByteChannelFactory seekableByteChannelFactory;
+    protected SeekableByteChannelFactory<?> seekableByteChannelFactory;
```
drivenflywheel, line 183: SCBF references in the BaseDataObject class and the IBDO interface use the unbounded wildcard type specifier (`<?>`), since the BDO itself should have no need to require a specific `SeekableByteChannel` implementation, exactly as it had been before the `SeekableByteChannelFactory` interface was made generic.

src/main/java/emissary/core/channels/FileChannelFactory.java
```diff
@@ -28,14 +22,14 @@ private FileChannelFactory() {}
      * @return a new instance
      * @see SeekableByteChannelHelper#file(Path)
      */
-    public static SeekableByteChannelFactory create(final Path path) {
-        return ImmutableChannelFactory.create(new FileChannelFactoryImpl(path));
+    public static ImmutableChannelFactory<?> createFactory(final Path path) {
```
drivenflywheel, line 25: This method rename is one I repeated across the other `XXXChannelFactory` classes as well, for the selfish reason that I makes it easier for me to read some chained calls:
 - instance methods that return a channel instance are still named `create`
 - static methods that return a channel _factory_ instance are now named `createFactory`, rather than `create`

src/main/java/emissary/core/channels/FileChannelFactory.java
```diff
@@ -49,47 +43,9 @@ private FileChannelFactoryImpl(final Path path) {
...
          */
         @Override
-        public SeekableByteChannel create() {
-            return new LazyFileChannelImpl(path);
+        public LazyOpeningFileChannel create() {
+            return new LazyOpeningFileChannel(path);
         }
     }
 
-    private static final class LazyFileChannelImpl extends AbstractSeekableByteChannel {
```
drivenflywheel, line 57: Another example when I pulled back the curtain some and broke some of the encapsulation.

src/main/java/emissary/core/channels/FillChannelFactory.java
```diff
@@ -70,18 +69,18 @@ private FillChannel(final long size, final byte value) {
...
         @Override
-        protected void closeImpl() throws IOException {
+        protected void closeImpl() {
             // Nothing to do.
         }
 
         @Override
         protected int readImpl(final ByteBuffer byteBuffer, final int maxBytesToRead) throws IOException {
-            final int bytesToFill = Math.min(maxBytesToRead, Integer.MAX_VALUE);
+            final int bytesToFill = Math.min(maxBytesToRead, byteBuffer.remaining());
```
drivenflywheel, line 83: Actual bug fix (with `Integer.MAX_VALUE` there this method is a no-op by definition), but I'm not sure it matters since I don't know any actual use case for `FillChannel` instances besides unit tests?

drivenflywheel:
Meh, not getting the return on decoration that I was optimistically hoping for.

PR #379, opened 2022-12-29
move new FileInputStream to Files.newInputStream

dev-mlb:


andrewbp:
Is this strictly meant to happen only in these files? I do see other places where this change can be made.

dev-mlb:
> Is this strictly meant to happen only in these files? I do see other places where this change can be made.

Yep. All the others are in `main` methods, so I was not concerned with them.

andrewbp:
> > Is this strictly meant to happen only in these files? I do see other places where this change can be made.
> 
> Yep. All the others are in `main` methods, so I was not concerned with them.

I see. Unrelated to the PR's approval, what's the reasoning for that?

andrewbp (approved):

jpdahlke (approved):

PR #380, opened 2022-12-30
Added functionality to allow ExtractionTest to verify that a value is not indexed

andrewbp:


dev-mlb (changes requested):
Can you update the Extraction test that is in the junit5 folder as well? We have to keep both of them synced until we deprecate and remove the junit 4 version. Also, we should probably add a TestExtractionTest for the junit 5 version.

fbruton (changes requested):
This should be a JUnit 5 update. Also, the current unit-test does not fully exercise the change.

andrewbp:
PR has been updated.

fbruton:
@dev-mlb @andrewbp I think we should consider changing `notIndexed` to either`!index` or `noindex` to emphasize negation of the `index` assertion. The same pattern can be followed if we decide to negate other assertions.

andrewbp:
> @dev-mlb @andrewbp I think we should consider changing `notIndexed` to either`!index` or `noindex` to emphasize negation of the `index` assertion. The same pattern can be followed if we decide to negate other assertions.

In looking at the other assertions and in thinking about potential future assertions, I think using "!" makes the most sense. It'd be the clearest way to describe what's happening.

drivenflywheel:
> > @dev-mlb @andrewbp I think we should consider changing `notIndexed` to either`!index` or `noindex` to emphasize negation of the `index` assertion. The same pattern can be followed if we decide to negate other assertions.
> 
> In looking at the other assertions and in thinking about potential future assertions, I think using "!" makes the most sense. It'd be the clearest way to describe what's happening.

I think the most precise and unambiguous option would be "doesNotContain or "!contains".  I don't see a need for the mode name to include "index", given that we have 4 others that don't. 

(TBH, it wouldn't suck to also add support for "contains" as an alias for "index", since it makes the intent immediately obvious when reading an answer file entry.)

andrewbp:
> 

Makes sense to me. I did find myself wondering what "index" meant in this context. I wouldn't be opposed to outright changing "index" to "contains", but I imagine the concern is upstream stuff that might currently be using "index". I can add it as an alias if we decide to do that as part of this ticket.

andrewbp:
None of these other matchMode assertions have tests in TestExtractionTest. If we're future proofing for other potential matchModes, then it's probably best to make test for the matchModes that already exists. It shouldn't be done as part of this ticket, but the tests being created here should be designed with that in mind.

fbruton:
> !contains
I'm in favor of adding an alias for index and supporting the following:
* `contains` and `index`
* `!contains` and `!index`

We can potentially deprecate and then remove`index` in the future.

drivenflywheel:
> > !contains
> > I'm in favor of adding an alias for index and supporting the following:
> 
> * `contains` and `index`
> * `!contains` and `!index`
> 
> We can potentially deprecate and then remove `index` in the future.

I think this is a good approach (adding the alias and using the "!" prefix.) 

I don't think deprecation should be anywhere on our radar at this point, given the huge number of warnings it would cause.

andrewbp:
PR updated based on the various feedback provided.

src/test/java/emissary/test/core/TestExtractionTest.java
```diff
@@ -51,6 +54,56 @@ void testCheckStringValueForCollectionFailure() throws JDOMException, IOExceptio
...
+        List<Element> dataList = answerDoc.getRootElement().getChild("answers").getChildren("data");
+        Element data = getAttributeFromDataChild(dataList, matchMode);
+
+        test.checkStringValue(data, "time:20221229", "testCheckStringValue!ContainsTrue");
+        assertThrows(AssertionError.class, () -> test.checkStringValue(data, "timestamp:20221229", "testCheckStringValue!ContainsFalse"));
+    }
+
+    private Element getAttributeFromDataChild(List<Element> dataList, String matchMode) {
+        Element data = null;
+        // Having different matchModes in the same data necessitates having to go through each child and filter for the correct one
```
andrewbp, line 98: To add to this comment, the other approach would be to make a different file for each matchMode we want to test, but that seemed cumbersome. If that's the preferred way to do it, then I'll change what I'm doing here.

fbruton: I recommend updating an existing answer file to exercise the that alias works should be sufficient. Maybe consider `sample.xml`.

andrewbp: Removed `contains` tests, updated `HtmlEscapePlaceTest/sample.xml`.

fbruton (changes requested):

src/test/java/emissary/test/core/junit5/TestExtractionTest.java
```diff
@@ -0,0 +1,98 @@
...
+        inputStream.close();
+        String matchMode = "!index";
+
+        WhyDoYouMakeMeDoThisExtractionTest test = new WhyDoYouMakeMeDoThisExtractionTest();
+
+        List<Element> dataList = answerDoc.getRootElement().getChild("answers").getChildren("data");
+        Element data = getAttributeFromDataChild(dataList, matchMode);
+
+        test.checkStringValue(data, "time:20221229", "testCheckStringValue!IndexTrue");
+        assertThrows(AssertionError.class, () -> test.checkStringValue(data, "timestamp:20221229", "testCheckStringValue!IndexFalse"));
```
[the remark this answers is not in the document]
andrewbp: Added that in, let me know if that works.

fbruton (approved):

dev-mlb (approved):

PR #382, opened 2023-01-08
Add super-linter github action

cfkoehler:
Github action that will lint:
- Github actions
- Jave
- Docker
- Markdown
- Xml
- yaml

It will run only on changed files for PR's and on all files when running on the master branch.
This action will not fail a PR but will provide a warning if it has a recommendation


cfkoehler:
Can't publish results from forks :( 
https://github.com/github/super-linter/issues/3419

I still think this could be useful. Just have to look in the logs for the results

.github/workflows/super-linter.yaml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
+        with:
+          # Full git history is needed to get a proper list of changed files within super-linter
+          fetch-depth: 0
+
+      - name: Lint Code Base
+        uses: github/super-linter/slim@v4.10.0
+        env:
```
jpdahlke, line 27:
```suggestion
        env:
          LOG_LEVEL: NOTICE
```
This seems to quiet down the very verbose (pun intended) printing that I'm not sure is helpful.

pom.xml
```diff
@@ -27,7 +27,7 @@
     <BUILD_TIMESTAMP>${maven.build.timestamp}</BUILD_TIMESTAMP>
     <EMISSARY_VERSION>${project.version}</EMISSARY_VERSION>
     <argLine />
-    <checkstyleFormatter>${project.basedir}/contrib/checkstyle.xml</checkstyleFormatter>
+    <checkstyleFormatter>${project.basedir}/.github/linters/checkstyle.xml</checkstyleFormatter>
```
jpdahlke, line 30:
```suggestion
    <checkstyleFormatter>${project.basedir}/contrib/checkstyle.xml</checkstyleFormatter>
```
If you use relative paths or symlinks, this can go back to the original.

jpdahlke (changes requested):
I didn't comb through the lint rules for yaml/markdown. I assume if we have an issue with those we can always update the rules.

.github/linters/.markdown-lint.yml
```diff
@@ -0,0 +1,7 @@
+---
+# MD013/line-length - Line length
+MD013:
+  # Number of characters, default is 80
+  line_length: 9999
```
ukengineer925, line 5: Should this be a bit lower, maybe 256 or less?

ukengineer925 (dismissed):

jpdahlke:
Agree with disabling the java checks for speed while we mess with this as well as markdown because of the shear volume of updates needed. I think we can tackle the other ones easily if we want.

jpdahlke (approved):

pechrista (approved):

PR #383, opened 2023-01-14
Remove hasConfigErrors from ConfigUtil

dev-mlb:
The method, hasConfigErrors, was needed in Jetty 6 to halt startup when there were config issues. The class that called this method was refactored/removed, so it appears that this is no longer needed.


cfkoehler (approved):
Everything still starts as expected

jpdahlke (approved):

PR #385, opened 2023-01-16
Add helper method to SeekableByteChannelHelper

ukengineer925:
Just a move of a helper method from elsewhere to tidy up some particular use cases, also renamed existing method to be more accurate.


drivenflywheel (approved):
LGTM

jpdahlke (approved):
I don't see current references to these methods that would be impacted by this change. I assume those changes are relevant to subsequent PRs that are awaiting this change.

PR #387, opened 2023-01-19
sonarlint :: cast long operands before assignment

jpdahlke:
Attempting to solve sonar bug reports for casting operands before assignment: 

- in-line cast to make as little impact as possible
- promote numeric values

### DIFF
--- a/src/main/java/emissary/kff/KffFile.java
+++ b/src/main/java/emissary/kff/KffFile.java
@@ -164,7 +164,7 @@ public class KffFile implements KffFilter {
             // mappedBuf.position(record.length * mid);
             // mappedBuf.get(record);
             try {
-                knownFile.seek(record.length * mid);
+                knownFile.seek(record.length * (long) mid);
                 int count = knownFile.read(record);
                 if (count != record.length) {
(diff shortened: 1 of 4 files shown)


src/main/java/emissary/kff/KffFile.java
```diff
@@ -164,7 +164,7 @@ private boolean binaryFileSearch(@Nullable byte[] hash, long crc) {
             // mappedBuf.position(record.length * mid);
             // mappedBuf.get(record);
             try {
-                knownFile.seek(record.length * mid);
```
jpdahlke, line 167: I opted to make this change instead of converting all the variables involved to long.

src/main/java/emissary/util/WindowedSeekableByteChannel.java
```diff
@@ -91,7 +91,7 @@ private void realignBuffers() throws IOException {
             return;
         }
         // keep track of our position
-        final int offset = this.buff1.position() + this.buff2.position();
```
jpdahlke, line 94: This is not the original line that was impacted, but converting an addition variable to a long made more sense to me than casting in-line during a subtraction operation.

fbruton (approved):

PR #389, opened 2023-01-23
WIP Answers XML Validator

andrewbp:
This work in progress aims to validate that answers xml files are using proper schema. The challenge is that a proper schema has not been determined, so for now I just want to create a base that accepts most existing test xml files.

### DIFF
--- a/src/test/java/emissary/test/core/junit5/AnswersXMLValidator.java
+++ b/src/test/java/emissary/test/core/junit5/AnswersXMLValidator.java
@@ -1,0 +1,70 @@
+package emissary.test.core.junit5;
+
+import emissary.config.Configurator;
+
+import org.jdom2.Document;
+import org.jdom2.Element;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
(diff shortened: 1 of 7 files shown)


src/test/java/emissary/test/core/junit5/AnswersXMLValidator.java
```diff
@@ -0,0 +1,70 @@
...
+                    "fileType", "meta", "nometa",
+                    "noview", "numAttachments", "shortName", "view");
+    private static final List<String> validInnerInnerFields = Arrays.asList("name", "value");
+    private static final List<String> validDataAttributes = Arrays.asList("matchMode");
+    Configurator configG;
+
+
+    public static void validate(Document answerDoc) {
+        Element root = answerDoc.getRootElement();
+        assertEquals("result", root.getName());
```
drivenflywheel, line 29: I think you'll find that downstream projects have many answer files with a root element named "root".   If we care about the root element name _(I'm not 100% sure we do)_, we might want to start with some sort of warning rather than something that will fail builds as soon downstream projects update to this version of emissary.

Actually, that point might be true for any of these assertions. Maybe these should start as warnings?  (Without integration testing, it's hard to quantify the number of downstream changes this will require for initial integration.)

At least one open downstream PR has proposed changes that would add new elements AND new attributes.

andrewbp: Updated to only log warnings instead of fail

PR #391, opened 2023-01-27
update actions checkout to 3.3.0

cfkoehler:
This updates the maven and codeql action to the latest checkout version.
Previous version gives a warning due to node.js actions being deprecated

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -39,7 +39,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3.3.0
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -27,7 +27,7 @@ jobs:
 
     steps:
     - name: Checkout repo
-      uses: actions/checkout@v3
+      uses: actions/checkout@v3.3.0
 
     - name: Configure Java & Maven
       uses: actions/setup-java@v3


jpdahlke:
```
[Analyze (javascript)] Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
```

jpdahlke (approved):

.github/workflows/codeql-analysis.yml
```diff
@@ -39,7 +39,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3.3.0
```
ukengineer925, line 42: Do we want to be so specific about the version? v3 == v3.3.0 currently, I doubt we care much about minor/patch version bumps.

cfkoehler: I did not realize v3==v3.3.3
From some of my security training it always talks about locking in specific releases but also this action has no permissions since it is a public repo

ukengineer925: I normally agree with specific versions too, but I think for these types of actions and dependencies, the overhead of maintaining updates for them feels like it outweighs the benefits of taking updates more regularly and easily. I have no strong preference either way though :smile:

ukengineer925 (approved):

.github/workflows/codeql-analysis.yml
cfkoehler: I agree on the added overhead. @jpdahlke do you have an opinion?

PR #393, opened 2023-01-27
Move position update during read up to AbstractSeekableByteChannel

ukengineer925:
We spotted this earlier as a potential source of bugs (indeed, `FileChannelFactory` would likely exhibit this behaviour), and would like to resolve it sooner rather than later before more implementations get created downstream (we're not aware of any that would result in having to co-ordinate this release with integration work though).

The old behaviour relied on a child class of `AbstractSeekableByteChannel` to update the position of the channel during the read, but this could easily be forgotten. This change moves this into Abstract to avoid the implementer having to do this in every `readImpl` override.

### DIFF
--- a/src/main/java/emissary/core/channels/FileChannelFactory.java
+++ b/src/main/java/emissary/core/channels/FileChannelFactory.java
@@ -82,6 +82,8 @@ public final class FileChannelFactory {
         @Override
         protected int readImpl(ByteBuffer byteBuffer, int maxBytesToRead) throws IOException {
             initialiseChannel();
+            // Set position for underlying channel, otherwise this won't match the outer channel position
+            channel.position(position());
             return channel.read(byteBuffer);
         }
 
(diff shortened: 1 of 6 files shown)


ukengineer925:
I've also included a few related fixes and improvements as part of this PR - particularly the InputStreamChannelFactory change which is still inherently related to positioning.

src/main/java/emissary/core/channels/FileChannelFactory.java
```diff
@@ -82,6 +82,8 @@ protected void closeImpl() throws IOException {
         @Override
         protected int readImpl(ByteBuffer byteBuffer, int maxBytesToRead) throws IOException {
             initialiseChannel();
+            // Set position for underlying channel, otherwise this won't match the outer channel position
```
drivenflywheel, line 85: It seems odd to be setting the FileChannel's position here, rather than setting it as part of the `position(final long position)` call.

ukengineer925: Do you mean as in overriding the `position(final long position)` method?

drivenflywheel: Well, that the `position(final long position)` method in `AbstractSeekableByteChannel` is marked as `final`, but it might be reasonable to have that method to call an optional `positionImpl(final long position)` method that channel implementations could override as needed.

ukengineer925: Gotcha. I think given this is the only implementation currently, I'll hang fire until we have another one and do the refactoring at that point. Agree that having a more extensible/override-able base class would be desirable in the future.

drivenflywheel (approved):

fbruton (approved):

PR #394, opened 2023-01-27
add dependency review action

cfkoehler:
Got this as a recommended action for PR's and since I am going through and updating actions might as well add another one!

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,0 +1,20 @@
+# Dependency Review Action
+#
+# This Action will scan dependency manifest files that change as part of a Pull Request, surfacing known-vulnerable versions of the packages declared or updated in the PR. Once installed, if the workflow run is marked as required, PRs introducing known-vulnerable packages will be blocked from merging.
+#
+# Source repository: https://github.com/actions/dependency-review-action
+# Public documentation: https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review#dependency-review-enforcement
+name: 'Dependency Review'
+on: [pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,20 @@
...
+  contents: read
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@v3
+      - name: 'Dependency Review'
+        uses: actions/dependency-review-action@v2
```
dev-mlb, line 20: Looks like there is a v3, is there a reason to stick with v2?

cfkoehler, line 20:
```suggestion
        uses: actions/dependency-review-action@v3
```

dev-mlb (approved):

ldhardy:
I think it will be necessary to modify default configs for this action.

The severity defaults to low, I'm not sure we're going to want that?  Or maybe we do?
```
with:
          fail-on-severity: moderate
```

Is there  a way to see an example of this before it gets merged?  Can I branch your personal branch and create a PR back to your branch to see it run?

cfkoehler:
> Is there a way to see an example of this before it gets merged? Can I branch your personal branch and create a PR back to your branch to see it run?

This PR has the action run showing no dependency changes: https://github.com/NationalSecurityAgency/emissary/actions/runs/4638563872

PR showing new vulnerable dependencies: https://github.com/NationalSecurityAgency/emissary/actions/runs/4775533594/jobs/8489903135?pr=467
Screenshot 2023-04-22 at 8 43 22 PM

cfkoehler:
> I think it will be necessary to modify default configs for this action.
> 
> The severity defaults to low, I'm not sure we're going to want that? Or maybe we do?
> 
> ```
> with:
>           fail-on-severity: moderate
> ```

I think it is ok to have it set to low. This action will not block a PR if it is from a fork. Either way we can override if it does block. I think we should need to take action to allow even a low severity vulnerability to be merged.

ldhardy (approved):

PR #411, opened 2023-02-02
Simplify AbstractSeekableByteChannel by right-sizing the provided byteBuffer

ukengineer925:
This builds on #393 by also right-sizing the provided byteBuffer to avoid reading more than we should, and abstracting this away from implementations. Now, the provided byteBuffer to implement against will have the right amount of remaining bytes to read into based on its position and capacity.

This PR is blocked until #393 is merged.


drivenflywheel (approved):
LGTM

ukengineer925:
> LGTM

Thanks - that was quick! I've pushed up an amendment to the way the new limit is calculated based on advice here, but otherwise I'm not anticipating making any further changes at my end unless there's review comments.

drivenflywheel (approved):

fbruton (approved):

PR #417, opened 2023-02-21
Adding ability to configure Transform History for explicit files/tests

sambish5:
https://github.com/NationalSecurityAgency/burrito-grande/issues/1990
https://github.com/NationalSecurityAgency/burrito-grande/issues/1991

Allows for the creation of a `reduced-transform-itinerary-list.csv` to allow the configuration of the logged transform history.
As of right now, I built out the functionality to configure to two different outputs compared to the full itinerary:
1. REDUCED - This outputs just the dropoff place of the transform history for the matching tests/files.
2. NOURL - Removes the url from each place in the given test/file history

More can be added/these can be altered, but I figured this was a good starting point to be able to build out the base implementation


src/main/java/emissary/util/PayloadUtil.java
```diff
@@ -65,18 +65,87 @@ public static String getPayloadDisplayString(final IBaseDataObject payload) {
...
     }
 
+    /**
+     * Generate a string about the payload object. This checks reduced-transform-itinerary-list.csv against current
+     * filenames to see if a reduced history is desired on output
+     *
+     * @param payload the payload to describe
+     * @param reducedTransformItineraryList string list of tests/files that should have reduced transform history
+     */
+    public static String getPayloadDisplayString(final IBaseDataObject payload, List<String> reducedTransformItineraryList) {
```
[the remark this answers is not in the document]
sambish5: The original ask was to allow the transform inventory to be configurable, explicitly the transform history. It was mentioned more pieces may want to be configured in the future, so no that was not the plan to deprecate that method, at least as of right now.

src/test/resources/emissary/util/PayloadUtilTest/reduced-transform-itinerary-list-test.csv
```diff
@@ -0,0 +1,9 @@
+Test file and example of the reducedTransformItineraryList,
```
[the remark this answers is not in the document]
sambish5: I thought about it and thought something similar to the lists that we have for the known-file-sets may be the simplest thing to do. However, after thinking on it more this weekend after reading this comment, I'm going to go through today and do some testing around to see if a Payload.cfg would be more beneficial.

src/main/java/emissary/core/MobileAgent.java
```diff
@@ -760,7 +765,22 @@ protected void logAgentCompletion(final IBaseDataObject payloadArg) {
         final Logger dest = (isProbe == null) ? logger : probeLogger;
 
         if (dest.isInfoEnabled()) {
-            dest.info(PayloadUtil.getPayloadDisplayString(payloadArg));
+            // grab reduced-transform-itinerary-list if it exists and add lines to reducedTransformItineraryList
+            List<String> reducedTransformItineraryList;
+            try (Stream<String> lines =
+                    Files.lines(Paths.get(System.getenv("PROJECT_BASE") + "/../src/test/resources/reduced-transform-itinerary-list.csv"))) {
```
cfkoehler, line 771: Related to @andrewbp comment. I think it is a little odd to load what is basically a 'configuration' file from the test resources. I don't believe the test/resources directory is available in a normally running emissary. 
So following the standard .cfg file pattern and keeping it in the config directory might be best. If you need a custom .cfg for unit tests and such that should go in test/resources

cfkoehler, line 771: Another thing is I think this is loading the configuration each time for every object. We should probably just load it once at startup.

sambish5:
@cfkoehler @andrewbp reworking currently, should have an update for it up hopefully tomorrow.

andrewbp (approved):
Looks good to me.

cfkoehler (changes requested):
Sorry for the delay in the review Sam. This looks good now.
Only thing is that I am getting a build failure from the import-sorter that recently changed.
I did a rebase and it fixed the errors

andrewbp:
Getting test failures after pulling in latest changes:
```
[ERROR] Failures: 
[ERROR]   ClassLocationVerificationProviderTest.testVerifyClassInWorkingDirectory:15 expected: <true> but was: <false>
```
master rebase did not solve the issue.

sambish5:
@andrewbp hmm odd, that test isn't failing for me. I'll look into it and see if I can find anything

andrewbp:
> @andrewbp hmm odd, that test isn't failing for me. I'll look into it and see if I can find anything

The test passes if I just run the test, but fails consistently when I try to build Emissary. Could be something on my end, but I'm not sure what it could be.

cfkoehler:
> > @andrewbp hmm odd, that test isn't failing for me. I'll look into it and see if I can find anything
> 
> The test passes if I just run the test, but fails consistently when I try to build Emissary. Could be something on my end, but I'm not sure what it could be.

I am not having any build errors with both centOS and macOS. Running maven 3.9.1 and 3.8.1.
@andrewbp What is the failure that you are getting?

cfkoehler (dismissed):

arp-0984:
I'm also seeing the same error as Andrew, but only when I build it on the docker container. When I build it locally it builds successfully.
`[ERROR] Failures: 
[ERROR]   ClassLocationVerificationProviderTest.testVerifyClassInWorkingDirectory:15 expected: <true> but was: <false>`

arp-0984:
Just noting that I also saw the same issue Andrew mentioned when building on the docker container, but I also have the same issue when I build the master branch so it doesn't appear to be related to this PR.

src/test/java/emissary/util/PayloadUtilTest.java
```diff
@@ -98,10 +98,60 @@ void testMultiLineString() {
...
+        final String answer = PayloadUtil.getPayloadDisplayString(d, false);
+        PayloadUtil.noURLHistory.clear();
+
+        // verify
+        assertTrue(answer.contains("\n"), "Must be multi-line string");
+        assertTrue(answer.contains("filename: noUrlHistory"), "Answer did not contain the correct filename");
+        assertTrue(answer.contains("FOO.UNKNOWN.FOOPLACE"),
+                "Answer should not contain the URL");
+        assertTrue(answer.contains("BAR.UNKNOWN.BARPLACE"),
+                "Answer should not contain the URL");
```
arp-0984, line 154: In addition to the above two, you can also add these two statements to confirm that the full URL is not present
`assertFalse(answer.contains("FOO.UNKNOWN.FOOPLACE.http://example.com:1234/FooPlace"), ...`
`assertFalse(answer.contains("BAR.UNKNOWN.BARPLACE.http://example.com:1234/BarPlace"), ...`

arp-0984 (changes requested):

jpdahlke:
> Just noting that I also saw the same issue Andrew mentioned when building on the docker container, but I also have the same issue when I build the master branch so it doesn't appear to be related to this PR. I can build successfully when I do not use docker.

https://github.com/NationalSecurityAgency/emissary/pull/476 should resolve this problem

sambish5:
Rebased and added @arp-0984 code changes to PayloadUtilTest.

src/main/java/emissary/util/PayloadUtil.java
```diff
@@ -67,16 +96,60 @@ public static String getPayloadDisplayString(final IBaseDataObject payload) {
...
+
+        // transform history output
+        switch (configureHistoryCase(fileName)) {
+            case REDUCED_HISTORY:
+                // found reduced history match, output only dropoff
+                sb.append("   ** reduced transform history **").append("\n").append("     dropOff -> ")
+                        .append(payload.getLastPlaceVisited())
+                        .append("\n");
+                break;
+            case NO_URL:
```
dev-mlb, line 108: Cases `NO_URL` and `default` are almost identical, can we squash these together?

dev-mlb (changes requested):

src/main/java/emissary/util/PayloadUtil.java
sambish5: I could, then just move the check for the NO_URL inside the default to alter the h.getKey() or h.getKeyNoUrl(). Whichever way of checking that you think is better

sambish5: Squashed NO_URL and default, left as switch statement for now for possible future transform history altercations.

arp-0984 (dismissed):

sambish5:
Changes based on information from @jpdahlke. Changed the configuration to look at FileType rather than File/Test name. Still currently only built out Reduced to output only DropOff and No_Url to remove urls from itinerary listed places, but left the ability to add in different filters.

arp-0984 (dismissed):

src/main/java/emissary/core/TransformHistory.java
```diff
@@ -193,6 +193,11 @@ public String getKey() {
             return key;
         }
 
+        public String getKeyNoUrl() {
+            String[] splitKey = key.split(".http");
```
dev-mlb, line 197: return StringUtils.substringBefore(key, ".http");

src/main/java/emissary/util/PayloadUtil.java
```diff
@@ -61,22 +90,63 @@ public static String getPayloadDisplayString(final IBaseDataObject payload) {
...
-        for (final TransformHistory.History h : th) {
-            sb.append("     ");
-            if (h.wasCoordinated()) {
-                sb.append("  ");
-            }
-            sb.append(h.getKey()).append("\n");
+
+        // transform history output
+        String historyCase = configureHistoryCase(fileType);
+        switch (historyCase) { // leaving as switch for future transform history altercations
```
dev-mlb, line 103: this can be simplified to an if statement:
`if (REDUCED_HISTORY.equals(historyCase)) {`

src/main/java/emissary/util/PayloadUtil.java
```diff
@@ -61,22 +90,63 @@ public static String getPayloadDisplayString(final IBaseDataObject payload) {
...
+                break;
+            default:
+                for (final TransformHistory.History h : th) {
+                    sb.append(" ");
+                    if (h.wasCoordinated()) {
+                        sb.append(" ");
+                    }
+
+                    // check is NO_URL or not
+                    if (historyCase.equals(NO_URL)) {
```
dev-mlb, line 118: `sb.append("    ").append(NO_URL.equals(historyCase) ? h.getKeyNoUrl() : h.getKey()).append("\n");`

src/main/java/emissary/util/PayloadUtil.java
```diff
@@ -61,22 +90,63 @@ public static String getPayloadDisplayString(final IBaseDataObject payload) {
...
+    /**
+     * Check if fileType in PayloadUtil.cfg matches current payload fileType
+     *
+     * @param fileType current payload fileType
+     * @return string for output case
+     */
+    public static String configureHistoryCase(String fileType) {
+        for (String currentReducedItem : reducedTransformHistory) {
+            if (!currentReducedItem.equals("") && fileType.equalsIgnoreCase(currentReducedItem)) {
+                return REDUCED_HISTORY;
```
dev-mlb, line 137: What happens if the filetype is in both REDUCED_HISTORY and NO_URL? REDUCED_HISTORY takes precedence, but is that expected? Would we want both? If we don't want both at the same time, I'd update the PayloadUtil.cfg with some explanation on the expected behavior if filetype is in both lists.

dev-mlb (changes requested):

src/main/java/emissary/util/PayloadUtil.java
sambish5: @cfkoehler what do you think about this? I don't know if your initial intent wanted multiple reduction cases at once or not. Could be something built out later as well

src/main/java/emissary/core/TransformHistory.java
drivenflywheel: Man, I like that helper method. 👍

src/main/java/emissary/util/PayloadUtil.java
drivenflywheel: My opinion is that a filetype can have only a single explicit "history preference": MINIMAL or NO_URL.  If I were to implement it, I'd use a Map (filetype -> preference_type), and first one in wins.  Would probably log a warning upon conflicting entries for a given filetype.  That's easier for me to wrap my head around and (I think) more easily accommodates future preference options than separate Lists can.

sambish5: Yeah, I like that approach as well.

sambish5:
Made changes based off review from @dev-mlb with map suggestions from @drivenflywheel . Further specified and streamlined what happens if a singe FileType is included in more than one transform history configuration.

dev-mlb (changes requested):

src/main/java/emissary/util/PayloadUtil.java
```diff
@@ -26,6 +31,38 @@ public class PayloadUtil {
...
+        Configurator configG = null;
+        try {
+            configG = ConfigUtil.getConfigInfo(PayloadUtil.class);
+        } catch (IOException e) {
+            logger.error("Cannot open default config file", e);
+        }
+
+        // fill lists with data from PayloadUtil.cfg for transform itinerary/history
+        if (configG != null) {
+            for (ConfigEntry configEntry : configG.getEntries()) {
```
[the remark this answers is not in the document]
sambish5: I went with the "first in wins" idea because that was what was mentioned by @drivenflywheel in your last change request. Yes, REDUCED_HISTORY isn't guaranteed if it's not the first in, but there wasn't per-say one preference that carried precedence over others in my mind. If there was, I figured that would be listed first in the cfg so it would be grabbed first anyway.

I wrote out the second option you mentioned first, and then changed it to the way I ended up making it, because instead of populating the whole map and then checking if there are multiple entries, that is done when all the entries are grabbed one by one with the `config.getEntries()`. Add the first found example for a fileType, and then if it is found again, log the warning message saying XML is using whatever the first matching preference was. (As I typed this out, realized I also need to add a warning message to what I have if the same entry, i.e. `REDUCED_HISTORY = XML`, is in the same list in the cfg more than once haha)

If the concern is other non-history-itinerary things being added to cfg down the road, then yes, option two of grabbing both lists and adding them to the map, then having the multiple entries check is the better option. More code would have to be added in the future for any other history configurations, but allows for the most control as you said.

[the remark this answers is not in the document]
drivenflywheel: My thinking was somewhere in the realm of @dev-mlb's two options. `config.getEntries` is too wide a swing for my comfort.  Hopefully we're overestimating the need for future-proofing; I think Option 2 is great for now, and we make it a future problem if we need even more history formats.  I don't care which way we choose to handle conflicts (first in, last in, X > Y, whatever) , so long as we document the choice right in the code itself.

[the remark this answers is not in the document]
sambish5: Alrighty then I will alter it to be like the second option mentioned!

sambish5:
@dev-mlb @drivenflywheel updated to be more like option 2 mentioned in the last review. Should also still keep it (somewhat) simple for future preferences

src/main/java/emissary/util/PayloadUtil.java
```diff
@@ -26,6 +30,55 @@ public class PayloadUtil {
...
+    }
+
+    /**
+     * Check if fileType is already mapped to history preference. If not, set fileType -$gt; preference in map
+     *
+     * @param fileType current fileType pulled from cfg
+     * @param preference preference associated with current cfg fileType
+     */
+    protected static void setFileTypeHistoryPreference(String fileType, String preference) {
+        if (historyPreference.containsKey(fileType)) {
```
drivenflywheel, line 68: In the future (i.e., not this PR), we could add an option to fail initialization if a fileType had more than one entry, but that's not something I want called from a static initialization block.

drivenflywheel (approved):
I think this is GTG.  For not completely rational reasons, I'd prefer _(no pun intended)_ that the unit tests used an explicit file type other than "UNKNOWN", but from a code perspective "UNKNOWN" is just as good as any other file type.

dev-mlb (approved):

PR #426, opened 2023-03-02
sonar :: add synchronized to clear() and isInUse() methods in HDMobil…

dev-mlb:
…eAgent to match parent class implementation

### DIFF
--- a/src/main/java/emissary/core/HDMobileAgent.java
+++ b/src/main/java/emissary/core/HDMobileAgent.java
@@ -118,7 +118,7 @@ public class HDMobileAgent extends MobileAgent {
      * Clear the payloadList and all other state info
      */
     @Override
-    protected void clear() {
+    protected synchronized void clear() {
         super.clear();
         this.payloadList.clear();
     }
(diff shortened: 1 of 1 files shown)


jdcove2 (dismissed):
Generally, synchronization on instance methods (i.e. "this") and static methods (i.e. "Clazz.class") is discouraged because other threads can also synchronize on "this" and "Clazz.class", which can cause deadlocks. The preferred way is to use synchronization blocks on a private object. I usually create a private object just for this (e.g. "private final Object lock = new Object()").

Given that this method already has a bunch of synchronized instance methods, synchronizing two more hopefully will not be a problem. I would suggest careful full-scale testing of this PR as this is the main multi-threaded entry point.

dev-mlb:
> Generally, synchronization on instance methods (i.e. "this") and static methods (i.e. "Clazz.class") is discouraged because other threads can also synchronize on "this" and "Clazz.class", which can cause deadlocks. The preferred way is to use synchronization blocks on a private object. I usually create a private object just for this (e.g. "private final Object lock = new Object()").
> 
> Given that this method already has a bunch of synchronized instance methods, synchronizing two more hopefully will not be a problem. I would suggest careful full-scale testing of this PR as this is the main multi-threaded entry point.

Agreed, but when @Overrides of synchronized methods are not themselves synchronized, the result can be improper synchronization as callers rely on the thread-safety promised by the parent class. I felt like the best option, at the moment, was to mark these as `synchronized` to match the parent implementation and not refactor all of MobileAgent/HDMobileAgent.

fbruton (dismissed):

PR #427, opened 2023-03-02
sonar :: fix conditionally executed code should be reachable (bug)

dev-mlb:


src/main/java/emissary/command/WhatCommand.java
```diff
@@ -174,29 +173,22 @@ protected void processFilePath(final Path path) throws ParserException, IOExcept
...
+            try {
+                final DecomposedSession sessionHash = parser.getNextSession();
+                final IBaseDataObject payload = producer.createAndLoadDataObject(sessionHash, parent);
+                final Identification id = identify(payload, this.headerIdentification);
+
+                LOG.info(payload.getFilename() + "/" + fileName + ": " + id.getTypeString() + " [" + payload.dataLength() + "]");
+            } catch (ParserEOFException eof) {
+                LOG.debug("Parser reached end of file: ", eof);
+                // Expected EOF
+                break;
```
andrewbp, line 191: Sonarlint doesn't like this loop. It could be changed to either evaluate `parser.isFullyParsed()`, or update some boolean when eof is reached.

dev-mlb: I am just trying to get rid of `raf` and fix the bug for code not reachable. Changing the while loop is out of scope for this pr.

fbruton (approved):

andrewbp (approved):

PR #431, opened 2023-03-06
tests :: updates to EmissaryApiTest and adds config endpoint test

dev-mlb:


src/test/java/emissary/server/api/EmissaryApiTest.java
```diff
@@ -289,4 +294,50 @@ void apiutils() throws Exception {
...
+            assertEquals("[CLUSTER]", entity.getLocal().getConfigs().get(2).getFlavors().toString());
+            assertEquals("[node.cfg, node-CLUSTER.cfg]", entity.getLocal().getConfigs().get(2).getConfigs().toString());
+        }
+    }
+
+    private static final class TestEmissaryNode extends EmissaryNode {
+        public TestEmissaryNode() {
+            super();
+            this.nodeName = "TestNode";
+            this.nodePort = TEST_PORT;
```
fbruton, line 335: Is it necessary to set `this.nodeName` and `this.nodePort` when the result of the methods are mocked?

PR #433, opened 2023-03-07
sonar :: check the return value of inputstream.read for Executrix (bug)

dev-mlb:
Just logging if differences found to preserve original behavior


src/main/java/emissary/util/shell/Executrix.java
```diff
@@ -173,12 +173,15 @@ public static byte[] readFile(final String theFileName, final int length) throws
 
         try (InputStream theStream = Files.newInputStream(Paths.get(theFileName))) {
             final int avail = theStream.available();
-            if ((length == -1) || (length >= avail)) {
-                theContent = new byte[avail];
-            } else {
-                theContent = new byte[length];
+            final int expected = length == -1 || length >= avail ? avail : length;
```
jdcove2, line 176: The readFile(...) method depends on InputStream.available() to size the buffer that will contain the data from the file. The javadoc for InputStream.available() says, "It is never correct to use the return value of this method to allocate a buffer intended to hold all data in this stream." I would suggest either...
1. Don't use InputStream.available() and read either until the end of the stream (i.e. end of the file) or the passed in "length" has been reached.
2. Use RandomAccessFile. This allows one to get the actual length of the file in order to correctly size the buffer and then do a full read of the data into the buffer.

jdcove2 (changes requested):

src/main/java/emissary/util/shell/Executrix.java
```diff
@@ -169,16 +169,16 @@ public static byte[] readFile(final String theFileName) throws IOException {
...
-        try (InputStream theStream = Files.newInputStream(Paths.get(theFileName))) {
-            final int avail = theStream.available();
-            if ((length == -1) || (length >= avail)) {
-                theContent = new byte[avail];
-            } else {
-                theContent = new byte[length];
+        try (RandomAccessFile raf = new RandomAccessFile(theFileName, "r")) {
+            final long avail = raf.length();
+            final int expected = length == -1 || length >= raf.length() ? Ints.saturatedCast(raf.length()) : length;
+            final int actual = raf.read(theContent = new byte[expected]);
```
jdcove2, line 178: This code works with the only drawback that fewer than "expected" bytes can be read, which causes the log statement to be created. Another option is to use "raf.readFully(...)", which either reads all of the bytes (ie. actual always equals expected) or throws an IOException. Additionally, "raf.readFully(...)" does not return anything so there is no need for the "actual" variable or the log statement.

dev-mlb: Yep, I saw that method but I stayed away from it because `readFully` will throw an `EOFException` that would be a new introduction to this method signature -- I'm not sure the impact to downstream. Since `Executrix.readFile` method accepts a length that is the **max** bytes to read, and not an expected exact length, I decided to stick with `read` for backwards compatibility. 

I turned the logging down to debug since this is not something that we are currently logging, and rather you turn it on if you want it.

jdcove2: I think it is probably ok with either "raf.read(...)" or "raf.readFully(...)" The probability that read(...) would fail to fill the buffer is pretty small given it has access to all the data and readFully(...) should never throw the EOFException since the buffer is sized correctly.

jdcove2 (approved):

src/main/java/emissary/util/shell/Executrix.java
dev-mlb: I ended up changing this to readFully to simplify the code. Agreed, it is probably okay either way.

PR #434, opened 2023-03-07
sonar :: KffFile cleanup

dev-mlb:


andrewbp (approved):

jpdahlke:
I would like to see a quick unit test for this change.

jpdahlke (changes requested):
I would like to see a quick unit test for this change.

dev-mlb:
> I would like to see a quick unit test for this change.

Looking back over the code, the methods are private so there is no need for the null checking. The check method already makes sure the passed in hash bytes are not null. I updated the methods to specify `@Nonnull` instead of `@Nullable`

src/main/java/emissary/kff/KffFile.java
```diff
@@ -143,42 +138,34 @@ public String getPreferredAlgorithm() {
...
-            return false;
-        }
+    private boolean binaryFileSearch(@Nonnull byte[] hash, long crc) {
 
         // Initialize indexes for binary search
         int low = 0;
         int high = bSearchInitHigh;
 
         /* Buffer to hold a record */
-        byte[] record = new byte[recordLength];
```
dev-mlb, line 157: `record` is a restricted identifier and should not be used as an identifier (https://rules.sonarsource.com/java/RSPEC-6213)

src/main/java/emissary/kff/KffFile.java
```diff
@@ -143,42 +138,34 @@ public String getPreferredAlgorithm() {
      * @param crc Result of CRC calculation
      * @return true if the record is in the list, false if it isn't
      */
-    private boolean binaryFileSearch(@Nullable byte[] hash, long crc) {
-        if (hash == null && myPreferredAlgorithm != null) {
```
dev-mlb, line 147: The only call to this method is `check` which already checks hash for `null`. Since this is a private method, I'm thinking we can remove this useless check and add `@Nonnull` to the parameter declaration.

drivenflywheel (approved):

jpdahlke (approved):

jpdahlke:
> I would like to see a quick unit test for this change.

OBE

PR #436, opened 2023-03-08
Sanitize string list parameters in RequestUtil

sambish5:


src/main/java/emissary/server/mvc/adapters/RequestUtil.java
```diff
@@ -73,6 +74,41 @@ public static String[] getParameterValues(final ServletRequest request, final St
...
+            retList = new ArrayList<>(1);
+            retList.add(0, (String) o);
+        }
+
+        if (retList != null) {
+            for (int i = 0; i < retList.size(); i++) {
+                logger.debug("RequestUtil.getParameterValues for {} [{}]: {}", param, i, retList.get(i));
+            }
+        } else {
+            logger.debug("RequestUtil.getParameterValues for {} is null", param);
```
andrewbp, line 106: Would it be more helpful for this log (and possibly the one on line 89) to indicate `"RequestUtil.getParameterValuesStringList for {} [{}]: {}"` instead?

src/main/java/emissary/server/mvc/adapters/RequestUtil.java
```diff
@@ -73,6 +74,41 @@ public static String[] getParameterValues(final ServletRequest request, final St
         return retArray;
     }
 
+    /**
+     * Get attribute or parameters from request and return a list of Strings. Attribute has priority over parameter when
+     * both are present
+     */
+    public static List<String> getParameterValuesStringList(final ServletRequest request, final String param) {
```
dev-mlb, line 81: I'm leaning towards not including this method.

dev-mlb (changes requested):
I believe a bunch of the utils method signatures that accept a ServletRequest are hold overs when we used JSPs. A bunch of these methods appear to not be used anymore. Maybe just stick with sanitizeParametersStringList and simplify the pr? 

@jpdahlke thoughts??

src/main/java/emissary/server/mvc/adapters/RequestUtil.java
jpdahlke: That was my original reaction but I had not gone back to confirm history. If you both agree this would be unused then I would support removing this method as well.

drivenflywheel: Only currently referenced by its own unit test. +1 for removing it.

dev-mlb (approved):

andrewbp (approved):

drivenflywheel (approved):

PR #445, opened 2023-03-30
Verify class locations during initialization

fbruton:


dev-mlb (dismissed):

src/main/config/emissary.spi.ClassLocationVerificationProvider.cfg
```diff
@@ -0,0 +1,2 @@
+CLASS_NAME_emissary.analyze.Analyzer = "emissary"
+CLASS_NAME_emissary.analyze.Analyzer = "notexist"
```
jpdahlke, line 2: This works just like I anticipated, however we should take this 'fail' case out right?

fbruton: Yes, this should be removed. Left in by mistake after testing failure case.

jpdahlke (approved):

cfkoehler (approved):

PR #447, opened 2023-04-03
delete all tika-dependent software from emissary

smcgrattan:
Deletes tika dependency from pom. deletes TikaFilePlace and associated test and cfg files


pom.xml
```diff
@@ -27,24 +27,26 @@
...
     <dep.errorprone.javac.version>9+181-r4173-1</dep.errorprone.javac.version>
     <dep.errorprone.version>2.10.0</dep.errorprone.version>
+    <dep.gson.version>2.8.6</dep.gson.version>
     <dep.guava.version>31.1-jre</dep.guava.version>
     <dep.httpclient.version>4.5.13</dep.httpclient.version>
     <dep.httpcore.version>4.4.15</dep.httpcore.version>
     <dep.impsort.version>1.6.2</dep.impsort.version>
     <dep.jackson.version>2.13.4</dep.jackson.version>
-    <!-- jersey and tika provide; converge on latest -->
+    <dep.jakarta.activation-api.version>1.2.2</dep.jakarta.activation-api.version>
```
cfkoehler, line 48: What was the reason you needed to specify jakarta.activation? 
org.glassfish.jersey.media:jersey-media-json-jackson:jar:2.35:compile brings it in already but is the 1.2.1 version

pom.xml
```diff
@@ -122,6 +122,11 @@
         <artifactId>compiler</artifactId>
         <version>${dep.spullara.mustache.compiler.version}</version>
       </dependency>
+      <dependency>
+        <groupId>com.google.code.gson</groupId>
+        <artifactId>gson</artifactId>
+        <version>2.8.6</version>
```
drivenflywheel, line 128:
```suggestion
        <version>${dep.gson.version}</version>
```

pom.xml
smcgrattan: when I deleted the the tika dependency, I got a version conflict on jakarta.activation with 2 other dependencies. Rather than exclude it from one of the  dependencies, I added the newest version as a dependency (this approach was based on  previous feedback in another PR).

pom.xml
smcgrattan: Modified. If I took the time to define the dependency version, I guess I should use it, eh?

cfkoehler (approved):

jpdahlke (approved):

PR #459, opened 2023-04-14
Update dockerfiles to use java 11

cfkoehler:
Intentionally leaving the Alpine image to openjdk. Going to need a few more changes in order to install corretto.

### DIFF
--- a/contrib/docker/Dockerfile
+++ b/contrib/docker/Dockerfile
@@ -1,6 +1,8 @@
 FROM centos:7
 
 RUN yum update -y \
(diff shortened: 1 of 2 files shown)


dev-mlb (approved):
Matches #252 -- lgtm

dev-mlb:
> Whoops. Did not realize it was linked to emissary 8. I believe we can take this before we bump the compiler version if we want.

I can't think of a reason to hold off on this update, so I'm happy to get it in and slim down the java 11 pr even further.

drivenflywheel (approved):

PR #462, opened 2023-04-14
build :: set the mvn duplicate finder plugin to fail for conflicts

dev-mlb:


pom.xml
```diff
@@ -990,9 +990,9 @@
           <version>1.5.1</version>
           <configuration>
             <printEqualFiles>false</printEqualFiles>
-            <failBuildInCaseOfDifferentContentConflict>false</failBuildInCaseOfDifferentContentConflict>
-            <failBuildInCaseOfEqualContentConflict>false</failBuildInCaseOfEqualContentConflict>
-            <failBuildInCaseOfConflict>false</failBuildInCaseOfConflict>
+            <failBuildInCaseOfDifferentContentConflict>true</failBuildInCaseOfDifferentContentConflict>
+            <failBuildInCaseOfEqualContentConflict>true</failBuildInCaseOfEqualContentConflict>
```
drivenflywheel, line 994: Are we sure we should fail the build even for identical content?

dev-mlb: Nope...probably too strict

pom.xml
```diff
@@ -990,9 +990,9 @@
           <version>1.5.1</version>
           <configuration>
             <printEqualFiles>false</printEqualFiles>
-            <failBuildInCaseOfDifferentContentConflict>false</failBuildInCaseOfDifferentContentConflict>
+            <failBuildInCaseOfDifferentContentConflict>true</failBuildInCaseOfDifferentContentConflict>
             <failBuildInCaseOfEqualContentConflict>false</failBuildInCaseOfEqualContentConflict>
-            <failBuildInCaseOfConflict>false</failBuildInCaseOfConflict>
+            <failBuildInCaseOfConflict>true</failBuildInCaseOfConflict>
```
drivenflywheel, line 995: `failBuildInCaseOfConflict` is a superset to the previous two options, but it completely ignores the content checksum.   We should specify only the two independent complimentary options, since `failBuildInCaseOfConflict -> true` stomps on them and still causes build failures if we comment out the `about.html` exclusion you added earlier

```suggestion
```

drivenflywheel (changes requested):
Minor change requested

drivenflywheel (approved):

ldhardy (approved):

PR #465, opened 2023-04-19
Fix KeywordScanner case insensitive search

fbruton:


src/main/java/emissary/util/search/KeywordScanner.java
drivenflywheel, line 1: This seems overly complicated, as a side effect of the existing `static` modifier on the `get256Value` method.  If we remove that modifier, we can collapse to the original `get256Value` method signature with no overloads needed:

```
    private int get256Value(final byte b) {
        return caseSensitive ? Byte.toUnsignedInt(b) : lowercase(Byte.toUnsignedInt(b));
    }
```

That simplication allows us to:
1. revert the added complexity in the `analyze` method (no need to pass `caseSensitive` as a param)
2. remove this redundant block in the `match` method:
```
            if (!this.caseSensitive) {
                currentByte = lowercase(currentByte);
            }
```

drivenflywheel (dismissed):
Some minor simplification requested

src/main/java/emissary/util/search/KeywordScanner.java
fbruton: Thank you. Totally agree with this approach.

arp-0984 (approved):

drivenflywheel (approved):
LGTM

SemperFidelisTyrannosaurus (approved):

PR #476, opened 2023-05-03
475 updated unit test to not fail in docker

SemperFidelisTyrannosaurus:
When building emissary in a docker container, the classLocPath variable set in ClassLocationVerificationProvider is set as /app/target/classes/ in the ClassLocationVerificationProvider unit tests. This breaks the testVerifyClassInWorkingDirectory unit test, as the unit test is written such that it expects that location path to have "emissary" as a substring. Re-write the tests to make sure that it works in both docker and non-docker builds.


drivenflywheel (approved):
`ClassLocationVerificationProvider` and this test class could use some clean up, but let's get builds "un-broken" first.

arp-0984 (approved):

PR #477, opened 2023-05-03
exclude helm repo from super lint

cfkoehler:
The helm templates are set up correctly but the super linter action does not view it as valid yaml. So excluding that directory for now. They have an outstanding feature request to add helm linting in a follow on version


cfkoehler:
An example build in my fork shows that the helm directory is no longer getting linted: 
https://github.com/cfkoehler/emissary/actions/runs/4877625397/jobs/8702439477

jpdahlke (approved):

PR #478, opened 2023-05-05
shutdown :: additional cleanup

dev-mlb:
These were other resources not being cleaned up during shutdown. These don't have the negative effects that we were seeing in tests as in #474, but I thought we should attempt to shut them down anyways.


cfkoehler (approved):
All seem pretty straightforward to me. I was able to observe during the shutdown and behaved as I would expect.

fbruton (approved):

PR #481, opened 2023-05-06
shutdown :: allow an option to force the shutdown of MobileAgents in the AgentPool

dev-mlb:


jpdahlke (changes requested):
It's a little weird that running a kill and then a stop behaves in such a way where the stop command almost re-awakens the killed agents and stops them again? Saving more state seems bad (ie: if we are shutting down) but maybe the original stop needs to also perform the namespace unbinding operation?

jpdahlke (approved):

sambish5 (approved):

PR #486, opened 2023-05-11
add test for MobileAgent typeLookup and handle null Stage name

jpdahlke:
Adds a test for the MobileAgent method that interfaces with the Stage class to determine type values. 

During testing it was discovered that providing `Stage.getByName(null)` throws an NPE, so handle that scenario as well.


fbruton (approved):

drivenflywheel (approved):
Could merge "as is" or explicitly check for null to prevent the NPE.  I'm fine with either.

PR #492, opened 2023-05-31
Consistently fail ExtractionTests if we did not create a meta/nometa tag correctly

arp-0984:
Added a null check in MetadataDictionary to prevent future NPE's, and modified the ExtractionTest to consistently fail if we do not create the meta/nometa tags correctly.

### DIFF
--- a/src/test/java/emissary/test/core/junit5/ExtractionTest.java
+++ b/src/test/java/emissary/test/core/junit5/ExtractionTest.java
@@ -226,14 +226,14 @@ public abstract class ExtractionTest extends UnitTest {
         // Check specified metadata
         for (Element meta : el.getChildren("meta")) {
             String key = meta.getChildTextTrim("name");
-            checkNullChildTag("meta", key);
+            checkForMissingNameElement("meta", key, tname);
             checkStringValue(meta, payload.getStringParameter(key), tname);
         }
 
         // Check specified nometa
         for (Element meta : el.getChildren("nometa")) {
             String key = meta.getChildTextTrim("name");
-            checkNullChildTag("nometa", key);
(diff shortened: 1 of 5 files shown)


sambish5 (dismissed):
LGTM

src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -294,6 +296,12 @@ protected void checkAnswers(Element el, IBaseDataObject payload, List<IBaseDataO
         }
     }
 
+    private void checkNullChildTag(String parentTag, String key) {
```
drivenflywheel, line 299: Couple of suggestions here:

1. Consider changing the method name to `checkForMissingNameElement` and changing the failure message to say "element" rather than the ambiguous "tag".  (This is more consistent with the other `checkXXX` methods)
2. Consider adding a `String tname` last parameter so the message to indicate which BDO *(TLD, att-1, etc.)* has the malformed meta/nometa tag. (Again, this mirrors the other `checkXXX` methods.)
```suggestion
    private void checkForMissingNameElement(String parentTag, String key, String tname) {
```

drivenflywheel (changes requested):
This was a really good find!  Suggesting minor tweaks to the failure message so a dev can more easily find and fix their bug if they introduce a change that suffers from this problem.

src/test/java/emissary/test/core/junit5/ExtractionTest.java
arp-0984: Added these changes so the error message will be clearer

drivenflywheel (approved):
LGTM 🚀

sambish5 (approved):

PR #498, opened 2023-06-12
Add payload size to Finished DropOff Message

cfkoehler:
Added a config option to make this easier to integrate with upstream services. 
If advised I can work to build the message with string builder to reduce the duplication of the log statement.


src/main/java/emissary/output/DropOffPlace.java
```diff
@@ -245,9 +246,16 @@ public List<IBaseDataObject> agentProcessHeavyDuty(final List<IBaseDataObject> p
                 objectMetricsLog.info(appendEntries(outputObjectMetrics(tld, objectMetricsFields)), "Finished DropOff");
             }
 
-            logger.info("Finished DropOff for object {}, with external id: {}, with total processing time: {}ms, with filetype: {}",
-                    tld.getInternalId(), this.dropOffUtil.getBestId(tld, tld), (new Date().getTime() - tld.getCreationTimestamp().getTime()),
-                    tld.getFileType());
+            if (outputCompletionPayloadSize) {
+                logger.info("Finished DropOff for object {}, with external id: {}, with total processing time: {}ms, with filetype: {}, payload size: {}",
```
drivenflywheel, line 250: Is it common for us to specify a size without units in log messages? (legitimate question, not trolling)
```suggestion
                logger.info("Finished DropOff for object {}, with external id: {}, with total processing time: {}ms, with filetype: {}, payload size: {} bytes",
```

cfkoehler: I think we generally do. But don't think we have a hard set rule. Definitely good practice. Especially since the time before specifies (ms).

Just an oversight on my part.

drivenflywheel (approved):
LGTM 🚀

src/main/java/emissary/output/DropOffPlace.java
```diff
@@ -245,9 +246,16 @@ public List<IBaseDataObject> agentProcessHeavyDuty(final List<IBaseDataObject> p
                 objectMetricsLog.info(appendEntries(outputObjectMetrics(tld, objectMetricsFields)), "Finished DropOff");
             }
 
-            logger.info("Finished DropOff for object {}, with external id: {}, with total processing time: {}ms, with filetype: {}",
-                    tld.getInternalId(), this.dropOffUtil.getBestId(tld, tld), (new Date().getTime() - tld.getCreationTimestamp().getTime()),
-                    tld.getFileType());
+            if (outputCompletionPayloadSize) {
+                logger.info("Finished DropOff for object {}, with external id: {}, with total processing time: {}ms, with filetype: {}, payload size: {} bytes",
```
dev-mlb, line 250: I get a formatting diff here when building:
```
                logger.info(
                        "Finished DropOff for object {}, with external id: {}, with total processing time: {}ms, with filetype: {}, payload size: {} bytes",
```

dev-mlb (approved):

drivenflywheel (approved):
LGTM 🚀

PR #503, opened 2023-06-21
Mockito 5

dev-mlb:
needs #252


cfkoehler:
Rebase should fix the checks. No longer need to run against the java 8 version

cfkoehler (approved):
Appears good to me.
Same count of test success/skipped
Dependency tree looks like I would expect.

jpdahlke (approved):

PR #511, opened 2023-07-03
Added null data check to DropOffPlace completion message setup

drivenflywheel:
Resolves a potential NullPointerException in `DropOffPlace.agentProcessHeavyDuty`.  (This NPE should never have been possible outside of unit tests anyway.)

### DIFF
--- a/src/main/java/emissary/output/DropOffPlace.java
+++ b/src/main/java/emissary/output/DropOffPlace.java
@@ -246,7 +246,7 @@ public class DropOffPlace extends ServiceProviderPlace implements EmptyFormPlace
                 objectMetricsLog.info(appendEntries(outputObjectMetrics(tld, objectMetricsFields)), "Finished DropOff");
             }
 
-            if (outputCompletionPayloadSize) {
+            if (outputCompletionPayloadSize && tld.data() != null) {
                 logger.info(
                         "Finished DropOff for object {}, with external id: {}, with total processing time: {}ms, with filetype: {}, payload size: {} bytes",
                         tld.getInternalId(), this.dropOffUtil.getBestId(tld, tld), (new Date().getTime() - tld.getCreationTimestamp().getTime()),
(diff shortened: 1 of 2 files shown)


src/main/java/emissary/output/DropOffPlace.java
```diff
@@ -246,7 +246,7 @@ public List<IBaseDataObject> agentProcessHeavyDuty(final List<IBaseDataObject> p
                 objectMetricsLog.info(appendEntries(outputObjectMetrics(tld, objectMetricsFields)), "Finished DropOff");
             }
 
-            if (outputCompletionPayloadSize) {
+            if (outputCompletionPayloadSize && tld.data() != null) {
```
fbruton, line 249: Updating line 253 `tld.data().length` to `tld.dataLength()` and removing the null check should also address the NPE, with the benefit outputting the log message
```suggestion
            if (outputCompletionPayloadSize) {
```

fbruton, line 249: Updating line 253 tld.data().length to tld.dataLength() and removing the null check should also address the NPE, with the benefit outputting the log message
```suggestion
            if (outputCompletionPayloadSize) {
```

fbruton (changes requested):

src/main/java/emissary/output/DropOffPlace.java
jpdahlke: I think this still (in a test scenario) can throw an NPE due based on how the `dataLength()` is implemented. The use of this method would be preferred however, so I think both should be done.

jpdahlke (changes requested):

fbruton (approved):

jpdahlke (approved):

src/main/java/emissary/output/DropOffPlace.java
```diff
@@ -246,11 +246,11 @@ public List<IBaseDataObject> agentProcessHeavyDuty(final List<IBaseDataObject> p
...
                 logger.info(
                         "Finished DropOff for object {}, with external id: {}, with total processing time: {}ms, with filetype: {}, payload size: {} bytes",
                         tld.getInternalId(), this.dropOffUtil.getBestId(tld, tld), (new Date().getTime() - tld.getCreationTimestamp().getTime()),
-                        tld.getFileType(), tld.data().length);
+                        tld.getFileType(), tld.dataLength());
```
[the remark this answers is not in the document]
drivenflywheel: Well, I was about to push this change, but the PR got merged. 🤷

PR #515, opened 2023-07-07
RegressionTest updates.

jdcove2:
This PR accomplishes the following:
- RegressionTest basic behavior changed to sha256 hash IBDO non-printable data instead of base64'ing it.
- Refactor encoders and decoders into their own class so they can be easily changed.

### DIFF
--- a/src/main/java/emissary/core/IBaseDataObjectXmlCodecs.java
+++ b/src/main/java/emissary/core/IBaseDataObjectXmlCodecs.java
@@ -1,0 +1,688 @@
+package emissary.core;
+
+import emissary.core.channels.InMemoryChannelFactory;
+import emissary.core.channels.SeekableByteChannelFactory;
+import emissary.core.channels.SeekableByteChannelHelper;
+import emissary.util.xml.AbstractJDOMUtil;
+
+import org.apache.commons.lang3.Validate;
+import org.jdom2.Element;
+import org.jdom2.Namespace;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.lang.reflect.Method;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.AbstractMap.SimpleEntry;
+import java.util.Base64;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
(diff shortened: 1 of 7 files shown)


src/main/java/emissary/core/IBaseDataObjectXmlCodecs.java
```diff
@@ -0,0 +1,688 @@
...
+    }
+
+    /**
+     * Class that contains the element decoders.
+     */
+    public static class ElementDecoders {
+        /**
+         * Decoder for boolean elements.
+         */
+        public final ElementDecoder booleanDecoder;
```
drivenflywheel, line 141: Could be a follow-on PR, but it would be nice to make the `ElementDecoder` instances private and leverage them through type-specific methods - i.e., 
```java
   public void decodeBoolean(List<Element> elements, IBaseDataObject ibdo, String methodName) throws Exception {
       booleanDecoder.decode(elements, ibdo, methodName);
   }
```

That makes the calling pattern a bit cleaner.  For an even cleaner API, ElementDecoders could leverage a static `Map<String, String>` of ElementName -> methodName, so the API would look like:
```java
   public void decodeBoolean(Element currentElement, IBaseDataObject ibdo, String elementName) throws Exception {
       List<Element> elements = currentElement.getChildren(elementName)
       String methodName = methodMap.get(elementName);
       Validate.notNull(methodName, "Could not find the IBDO method for element name " + elementName);
       booleanDecoder.decode(elements, ibdo, methodName);
   }
```
That would simplify the usage in IBaseDataObjectXmlHelper from:
```java
decoders.booleanDecoder.decode(element.getChildren(OUTPUTABLE_ELEMENT_NAME), ibdo, OUTPUTABLE_SET_METHOD_NAME);
```
to 
```java
decoders.decodeBoolean(element, ibdo, OUTPUTABLE_ELEMENT_NAME);
```

src/main/java/emissary/core/IBaseDataObjectXmlHelper.java
```diff
@@ -113,10 +71,6 @@ public final class IBaseDataObjectXmlHelper {
      * The IBaseDataObject set method name for Data.
      */
     public static final String DATA_SET_METHOD_NAME = "setChannelFactory";
```
drivenflywheel, line 73: We should consider moving the method name constants to an `emissary.core.constants.IBDOMethodNames` class and moving the element name constants to an `emissary.test.core.constants.AnswerFileElementNames` class.

drivenflywheel (changes requested):
A small bit of additional cleanup suggested.  This PR is a nice improvement, especially the default change to persist less obfuscated data into the answer files

sambish5 (approved):

PR #517, opened 2023-07-17
Lift convertHexDate from MsSqlJflex into TimeUtil

aL118:


src/main/java/emissary/util/TimeUtil.java
```diff
@@ -2,12 +2,10 @@
 
 import org.apache.commons.lang3.StringUtils;
 
-import java.time.DateTimeException;
-import java.time.Instant;
-import java.time.ZoneId;
-import java.time.ZonedDateTime;
+import java.time.*;
```
dev-mlb, line 5: Fix imports

src/test/java/emissary/util/TimeUtilTest.java
```diff
@@ -13,10 +13,7 @@
 import java.time.zone.ZoneRulesException;
 import java.util.Date;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNull;
-import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
```
dev-mlb, line 16: Fix imports

dev-mlb (changes requested):

dev-mlb (approved):

src/main/java/emissary/util/TimeUtil.java
```diff
@@ -232,6 +234,34 @@ public static ZonedDateTime getZonedDateFromISO8601(@Nullable final String dateS
...
+
+        long daysToAdd = Long.parseLong(dateHex, 16);
+        long millisToAdd = Math.round(Long.parseLong(timeHex, 16) * 10 / 3.0);
+
+        DateTimeFormatter dtf = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS");
+
+        String startingDate = "1900-01-01 00:00:00.000";
+        LocalDateTime ldt;
+
+        ldt = LocalDateTime.parse(startingDate, dtf);
```
dev-mlb, line 258: What happened to the try around this?

dev-mlb (changes requested):

src/main/java/emissary/util/TimeUtil.java
aL118: I initially removed the try block since enhanceHexDate no longer had access to the logger. Although now that enhanceHexDate is a standalone function, it may be more appropriate to throw an exception rather than return current date. 
To preserve the legacy behavior, I added another overloaded method that returns a default string if parse exception is thrown. This can be called as `TimeUtil.convertHexDate(dateString,LocalDateTime.now().toString())`.

drivenflywheel: Let's give TimeUtil its own Logger and use that.

src/main/java/emissary/util/TimeUtil.java
```diff
@@ -232,6 +234,53 @@ public static ZonedDateTime getZonedDateFromISO8601(@Nullable final String dateS
...
 
+    /**
+     * Parses hex date string into formatted string ("yyyy-MM-dd HH:mm:ss.SSS")
+     *
+     * @param hexDate hex string representation of the date
+     * @return converted time in string format yyyy-MM-dd HH:mm:ss.SSS
+     * @throws DateTimeParseException if could not parse date
+     *
+     */
+    public static String convertHexDate(String hexDate) {
```
drivenflywheel, line 245: (Optional, since it's outside the original scope of this PR)
This is a short method, but it contains a lot of magic values.

In the `hexNumber` assignment, would we replace 2` with a constant named `HEX_PREFIX_LEN`? 
In these `parseLong` calls, can we replace `16` with an equivalent constant named `HEX_RADIX`?  (We would _not_ use that constant in the `substring(..)` call.)

The `startingDate` String (or even better, an equivalent `LocalDateTime`) should also be a constant. 

```java
final static int HEX_PREFIX_LEN = "0x".length();`
final static int HEX_RADIX = 16;
final static String STARTING_DATE = "1900-01-01 00:00:00.000";
```

src/main/java/emissary/util/TimeUtil.java
```diff
@@ -232,6 +234,53 @@ public static ZonedDateTime getZonedDateFromISO8601(@Nullable final String dateS
...
+        String hexNumber = hexDate.substring(2);
+        String dateHex = hexNumber.substring(0, 8);
+        String timeHex = hexNumber.substring(8, 16);
+
+        long daysToAdd = Long.parseLong(dateHex, 16);
+        long millisToAdd = Math.round(Long.parseLong(timeHex, 16) * 10 / 3.0);
```
drivenflywheel, line 251: Just curious - anybody understand why this is dividing by 3.0?   The source from which we're migrating this is woefully uncommented.

drivenflywheel (changes requested):
Should give TimeUtil its own Logger so we can maintain behavioral consistency with the original interface.
```java
private static Logger logger = LoggerFactory.getLogger(TimeUtil.class);
```
The "use constants" suggestions are purely optional, although I hate dragging around existing tech-debt

src/main/java/emissary/util/TimeUtil.java
jpdahlke: I'm in favor of the logger additions. 

I'd also be in favor of marking the new "legacy" method with an `@Deprecated` tag so that its obvious we prefer not to use this method and move away from it in the future. @drivenflywheel does that help to alleviate some of your concerns?

src/main/java/emissary/util/TimeUtil.java
jpdahlke: It looks like these updates were already made actually. I'll let @drivenflywheel weigh in.

jpdahlke (approved):

PR #518, opened 2023-07-19
Add more string utility methods to StringUtil

aL118:
Added stripTicks, stripIdentifierMarks, stripEdgeQuotes, fixUnicodeString, and splitObjectNames static methods.
Added new test file for StringUtil.java.


drivenflywheel:
I don't think these methods belong in a generic StringUtil class; they're too special-purpose, and should probably go into their own domain-specific utility class. (Maybe "SqlStringUtil"?)

aL118:
After speaking with Jason, I will be closing this pull request since emissary doesn't have much to do with sql. Changes will be moved to the burrito version.

PR #521, opened 2023-07-19
error-prone :: remove unnecessary 'final'

cfkoehler:
Addresses:
[WARNING]
[UnnecessaryFinal] Since Java 8, it's been unnecessary to make local variables and parameters `final` for use in lambdas or anonymous classes. Marking them as `final` is weakly discouraged, as it adds a fair amount of noise for minimal benefit.
    (see https://errorprone.info/bugpattern/UnnecessaryFinal)

### DIFF
--- a/src/main/java/emissary/core/DiffCheckConfiguration.java
+++ b/src/main/java/emissary/core/DiffCheckConfiguration.java
@@ -88,7 +88,7 @@ public class DiffCheckConfiguration {
      * 
      * @param enabled set of pre-configured options
      */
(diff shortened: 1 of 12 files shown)


drivenflywheel (changes requested):
I think I disagree with about 90% of these removals.  Did error-prone really suggest all of these? I'm not sure I saw any that fit the scenario in the "UnnecessaryFinal" rule description.

cfkoehler:
Yeah I might have been a little too zealous with this change. Might be worth setting this rule to ignore.
I also did not realize that this is technically an 'experimental' rule. Might be worth setting all experimental rules to ignore?

PR #523, opened 2023-07-27
Sort parameters before diff in IBaseDataObjectDiffHelper

sambish5:
https://github.com/NationalSecurityAgency/burrito-grande/issues/2325
Have the map of parameters be sorted before looking at differences between two IBDO's.

### DIFF
--- a/src/main/java/emissary/core/IBaseDataObjectDiffHelper.java
+++ b/src/main/java/emissary/core/IBaseDataObjectDiffHelper.java
@@ -11,8 +11,11 @@ import java.nio.channels.Channels;
 import java.nio.channels.SeekableByteChannel;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Objects;
 
 public class IBaseDataObjectDiffHelper {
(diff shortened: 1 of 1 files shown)


sambish5:
@jdcove2 your suggestions were added. I didn't even think about the TreeMap auto sorting, so good point out to make it way simpler. Also added a test in to verify that the conversion to the TreeMap actually sorts.

src/test/java/emissary/core/IBaseDataObjectDiffHelperTest.java
```diff
@@ -294,4 +298,20 @@ void testDiffList() {
         ibdo2.setClassification("classification");
         verifyDiffList(1, ibdoList1, ibdoList2);
     }
+
+    @Test
+    void testParamSort() {
+        // set-up
+        Set<String> expectedParams = new HashSet<>();
```
drivenflywheel, line 305: Should probably use a `TreeSet` for the same reason we're now using a `TreeMap`
```suggestion
        Set<String> expectedParams = new TreeSet<>();
```

drivenflywheel (changes requested):
One minor change requested, otherwise looks good

src/test/java/emissary/core/IBaseDataObjectDiffHelperTest.java
sambish5: Dang you right, let me do it real quick

drivenflywheel (approved):
LGTM 🚀

PR #527, opened 2023-07-31
Update KffFile to use FileChannelFactory instead of RandomAccessFile

drivenflywheel:
Still need to create more robust unit test to demonstrate thread safety.


src/test/java/emissary/kff/KffFileTest.java
```diff
@@ -1,20 +1,44 @@
...
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
 class KffFileTest extends UnitTest {
+    private static final Logger LOGGER = LoggerFactory.getLogger(KffFileTest.class);
 
-    private static final String expectedShaHash = "000000206738748EDD92C4E3D2E823896700F849";
```
drivenflywheel, line 17: This was a misleading variable name given how it had been used, so I replaced it with ITEM_NAME and a value that made it clear that the value isn't used in the search.

drivenflywheel:
> The 3-argument constructor should probably have "DEFAULT_RECORD_LENGTH" instead of "24" in it's "this" call.

Done

src/test/java/emissary/kff/KffFileTest.java
```diff
@@ -1,20 +1,45 @@
...
+
+import static emissary.kff.KffFile.DEFAULT_RECORD_LENGTH;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
 class KffFileTest extends UnitTest {
+    public static final Random RANDOM = new Random();
```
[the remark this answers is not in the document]
drivenflywheel: If this ever randomly generates a sequence of bytes that match one of the existing Kff entries, I'll eat my desk.

fbruton (dismissed):

src/main/java/emissary/kff/KffFile.java
```diff
@@ -31,16 +34,17 @@ public class KffFile implements KffFilter {
     private final Logger logger;
 
     /** File containing SHA-1/CRC32 results of known files */
-    protected RandomAccessFile knownFile;
+    // protected RandomAccessFile knownFile;
```
cfkoehler, line 37: Jumping in here late so sorry if I am missing some context.
Any reason to just remove the line instead of commenting out `RandomAccessFile`
Same for line 89 as well.

Not blocking approval. There are a few other smells here that we can clean up in a later pr if we want

cfkoehler (approved):

cfkoehler (approved):

fbruton (approved):

PR #528, opened 2023-08-01
Make enhancements to convertHexDate

aL118:
Issue: https://github.com/NationalSecurityAgency/burrito-grande/issues/2330

### DIFF
--- a/src/main/java/emissary/util/TimeUtil.java
+++ b/src/main/java/emissary/util/TimeUtil.java
@@ -15,6 +15,8 @@ import java.time.temporal.ChronoUnit;
 import java.time.temporal.TemporalAccessor;
 import java.time.zone.ZoneRulesException;
 import java.util.Date;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 
 public class TimeUtil {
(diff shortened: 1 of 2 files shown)


src/main/java/emissary/util/TimeUtil.java
```diff
@@ -250,24 +252,33 @@ public static ZonedDateTime getZonedDateFromISO8601(@Nullable final String dateS
      *
      */
     public static String convertHexDate(String hexDate) {
+        String regex = "^([0-9A-Fa-f]{8})([0-9A-Fa-f]{8})";
```
drivenflywheel, line 255: If we're going to use a regex to parse the input string, we should consider adding a non-capturing group  that matches the "0x" prefix, rather than our current technique of stripping off the first 2 characters regardless of what they are (i.e,  `hexDate.substring(HEX_PREFIX_LEN)`). 
```suggestion
        String regex = "^(?:0x)([0-9A-Fa-f]{8})([0-9A-Fa-f]{8})";
```

We should also move the regex string and the compiled Pattern declarations out of this method and make them constants, since they're independent of the method input parameter.

src/main/java/emissary/util/TimeUtil.java
```diff
@@ -26,7 +28,8 @@ public class TimeUtil {
     private static final DateTimeFormatter DATE_ORDINAL = DateTimeFormatter.ofPattern("yyyyDDD").withZone(GMT);
     private static final DateTimeFormatter DATE_ISO_8601_SSS = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS");
     private static final String ISO_8601_TIME_DATE_STRING = "yyyy-MM-dd'T'HH:mm:ss'Z'";
-    private static final int HEX_PREFIX_LEN = "0x".length();
+    private static final String HEX_REGEX = "^(?:0x)([0-9A-Fa-f]{8})([0-9A-Fa-f]{8})";
+    private static final Pattern P = Pattern.compile(HEX_REGEX);
```
drivenflywheel, line 32: Let's give this pattern a more descriptive name, especially since this TimeUtil class is likely to be updated with additional members and methods.  For this codebase, we tend to discourage the use of single-character identifiers outside of unambiguous local scope.

```suggestion
    private static final Pattern HEX_DATE_PATTERN = Pattern.compile(HEX_REGEX);
```

src/main/java/emissary/util/TimeUtil.java
```diff
@@ -250,24 +253,30 @@ public static ZonedDateTime getZonedDateFromISO8601(@Nullable final String dateS
...
-        String timeHex = hexNumber.substring(8, 16);
+        Matcher m = P.matcher(hexDate);
 
-        long daysToAdd = Long.parseLong(dateHex, HEX_RADIX);
-        long millisToAdd = Math.round(Long.parseLong(timeHex, HEX_RADIX) * 10 / 3.0);
+        if (m.find()) {
+            String dateHex = m.group(1);
+            String timeHex = m.group(2);
+            long daysToAdd = Long.parseLong(dateHex, HEX_RADIX);
+            long millisToAdd = Math.round(Long.parseLong(timeHex, HEX_RADIX) * 10 / 3.0);
```
drivenflywheel, line 262: Let's a new inline comment here to explain the non-obvious math

```suggestion
            // timeHex represents the number of ticks (1/300 of a second) since midnight
            long millisToAdd = Math.round(Long.parseLong(timeHex, HEX_RADIX) * 10 / 3.0);
```

src/test/java/emissary/util/TimeUtilTest.java
```diff
@@ -159,9 +159,14 @@ void testDateAsFullIOS8601() {
...
-        String cellContents = "0x00009FF700F77536";
-        String cellContents2 = "0x0000A07800E93033";
-        assertEquals("2012-02-14 15:00:57.993", TimeUtil.convertHexDate(cellContents), "convertHexDate did not match");
-        assertEquals("2012-06-22 14:09:00.757", TimeUtil.convertHexDate(cellContents2), "convertHexDate did not match");
+        String hexDate1 = "0x00009FF700F77536";
+        String hexDate2 = "0x0000A07800E93033";
+        assertEquals("2012-02-14 15:00:57.993", TimeUtil.convertHexDate(hexDate1), "convertHexDate conversion did not match");
+        assertEquals("2012-06-22 14:09:00.757", TimeUtil.convertHexDate(hexDate2), "convertHexDate conversion did not match");
+
+        String invalidHexDate1 = "0x00009FF700F7753";
```
drivenflywheel, line 167:
```suggestion
        String invalidHexDate1 = "0x00009FF700F7753"; // too few characters
```

src/test/java/emissary/util/TimeUtilTest.java
```diff
@@ -159,9 +159,14 @@ void testDateAsFullIOS8601() {
...
+        String invalidHexDate2 = "0x00009FF_00F77536";
```
drivenflywheel, line 168:
```suggestion
        String invalidHexDate2 = "0x00009FF_00F77536"; // format doesn't support any non-hex characters
```

drivenflywheel (changes requested):
Looks good overall, asking for an identifier rename and insertion of a few inline comments

drivenflywheel (approved):
Looks great. 🚀

fbruton (changes requested):

drivenflywheel (approved):

cfkoehler (approved):

PR #529, opened 2023-08-02
2361 - Added hadolint config and set failure threshold to 'error'.

Kevinnnnnnnnnnnnnnnn:
Added hadolint config, set failure threshold to "error", and added a check/run that checks out the forked repository head when triggered by a pull request to prevent the workflow trying to run in a detached head state (and thereby signalling a false pass).


.github/workflows/super-linter.yaml
```diff
@@ -21,6 +21,19 @@ jobs:
         with:
           # Full git history is needed to get a proper list of changed files within super-linter
           fetch-depth: 0
+        if: ${{ !contains(github.event_name, 'pull_request') }}
+      - name: Checkout PR Code
+        uses: actions/checkout@v3.3.0
+        with:
+          # Full git history is needed to get a proper list of changed files within super-linter
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
```
cfkoehler, line 30: Do you have any examples of the runner log saying that it is in a failed state because of the detached head? I was not able to find any in a brief search through my own fork and the upstream repo.

Kevinnnnnnnnnnnnnnnn: If I'm reading the logs correctly, it doesn't fail because of the detached head, it PASSES because of the detached head.  I believe the linter is running on HEAD and if there is no HEAD (or it's detached) it runs on 0 files, finds 0 errors, and logs that as passing.

contrib/docker/.hadolint.yaml
```diff
@@ -0,0 +1 @@
+failure-threshold: error
```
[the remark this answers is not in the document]
Kevinnnnnnnnnnnnnnnn: Ideally we would have it still print the info and warnings but not treat them as failures.  Unfortunately I didn't see an easy way to do that...

.github/workflows/super-linter.yaml
jpdahlke: Answering my own question; I think all we need is to set the ref on checkout. That seemed to work for me in my local repo.

```
      - name: Checkout Code
        uses: actions/checkout@v3.3.0
        with:
          # Full git history is needed to get a proper list of changed files within super-linter
          fetch-depth: 0
        if: github.event_name != 'pull_request'

      - name: Checkout PR Code
        uses: actions/checkout@v3.3.0
        with:
          # Full git history is needed to get a proper list of changed files within super-linter
          fetch-depth: 0
          ref: refs/heads/${{ github.head_ref }}
        if: github.event_name == 'pull_request'
```

I've also seen usages of `github.event.pull_request.head.sha` for the ref instead. Not sure which is better.

Kevinnnnnnnnnnnnnnnn: @jpdahlke That works for me as well.  I took that idea combined with using a fallback value to reduce the number of steps and conditionals, and I think we've got a nice, elegant solution here :smiley:

jpdahlke (approved):

cfkoehler (approved):

PR #530, opened 2023-08-03
Adding objectTrace logger to output configurable fields to a new log file from PickUpPlace

arp-0984:


src/main/java/emissary/pickup/PickUpPlace.java
```diff
@@ -552,10 +568,27 @@ protected boolean processDataObject(IBaseDataObject d, String fixedName, File th
...
     }
 
+    /**
+     * Creates an entry in the object trace log. Can be overridden if desired.
+     * 
+     * @param d The IBDO
+     */
+    public void objectTraceLog(IBaseDataObject d) {
+        objectTraceLogger.info(appendEntries(
+                ObjectTracing.createTraceMessageMap(new String[] {"PickUp", d.getFilename(), String.valueOf(System.currentTimeMillis())})).toString(),
```
cfkoehler, line 588: We know that every trace message is going to need the timestamp in this format. Does it make sense to just have the ObjectTracingUtil add that in?
I don't think we need to have the timestamp field configurable. Just always part of the message. 
Just a thought. Open for more discussion

src/main/java/emissary/util/ObjectTracing.java
```diff
@@ -0,0 +1,38 @@
...
+    public static Map<String, String> createTraceMessageMap(String[] fieldValues) {
+
+        Map<String, String> jsonMap = new HashMap<>();
+
+        if (fieldValues.length != fieldNames.size()) {
+            logger.error("Cannot create log entry, the number of fields does not equals the number of values");
+            return Collections.emptyMap();
+        }
+
+        for (int i = 0; i < fieldValues.length; i++) {
```
cfkoehler, line 33: I see what you are getting at with making the fields configurable and then iterating through the array to make the json map.

But from a performance and complexity stand point do we really need it to be configurable?

Say in this util class could we just have a method that is buildPickupMessage(String filename)
and build the message we need with the fields we need and have it all in code. 
That method can be overridden in other software components instead of a config file.

I like the idea of having the configuration but worried that the fields need to be kept in sync in multiple places. And the order that they are added maters to keep key's and values from getting mixed up.

I don't have a better solution. Just trying to think out loud here.

arp-0984: You have a good point. If we don't need the fields in a config file, I'll play around with another solution.

src/main/java/emissary/util/FifteenMinuteLogbackAppender.java
```diff
@@ -0,0 +1,19 @@
+package emissary.util;
+
+import ch.qos.logback.core.rolling.RollingFileAppender;
+
+public class FifteenMinuteLogbackAppender<E> extends RollingFileAppender<E> {
+    private long start = System.currentTimeMillis(); // minutes
+    private static final int ROLLOVER_TIME_MINUTES = 15;
```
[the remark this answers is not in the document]
arp-0984: I don't have any reason why we wouldn't want it to be configurable, so I can add this (and rename the class)

src/main/java/emissary/util/ObjectTracing.java
arp-0984: I tested it and your suggestion of having a method to be overridden worked out well, so I went with that solution instead of a config file for the fields.

src/main/config/logback.xml
```diff
@@ -49,6 +49,35 @@
     </then>
   </if>
 
+  <!-- Enable output metrics logging by creating an environment variable of "ObjectTrace=true" -->
+  <if condition='${ObjectTrace} == true'>
```
cfkoehler, line 53: Might want this change as well
```suggestion
  <if condition="&quot;${ObjectTrace}&quot;.equals(&quot;true&quot;)">
```

https://github.com/NationalSecurityAgency/emissary/pull/549/files

cfkoehler (commented):
I am good with this once the one comment I left is resolved.
I imported this into our other project and messed with the overrides and such and all worked as I would expect

arp-0984:
The Super-Linter job keeps failing: "DOCKERFILE_HADOLINT_LINTER_RULES rules file (/action/lib/.automation/../../contrib/docker/.hadolint.yaml) doesn't exist. Terminating..."

This may be an issue with the job itself - I saw it was updated recently. I'm not sure why my 1-line change would have caused this issue if it passed before.

src/main/java/emissary/util/ObjectTracingLogbackAppender.java
drivenflywheel, line 1: I'm not sure why this class name is coupled to Object Tracing.  It's just a `RollingFileAppender` that gets its rollover interval (configurable in minutes) from an Emissary-style Configuration rather than a logback.xml file.  We're using the appender to support Object Tracing, but the class has no Object Tracing-specific functionality

src/main/resources/emissary/util/ObjectTracing.cfg
drivenflywheel, line 1: Is this file used?

drivenflywheel (changes requested):
Would like to see the implementation better encapsulated so Places can integrate with less coupling.

src/main/config/logback.xml
```diff
@@ -49,6 +49,35 @@
...
+  <!-- Enable output metrics logging by creating an environment variable of "ObjectTrace=true" -->
+  <if condition="&quot;${ObjectTrace}&quot;.equals(&quot;true&quot;)">
+    <then>
+      <appender name="OBJECT-TRACE" class="emissary.util.ObjectTracingLogbackAppender">
+        <encoder class="net.logstash.logback.encoder.LogstashEncoder">
+          <customFields>{"node":"${emissary.node.name}"}</customFields>
+          <includeMdc>false</includeMdc>
+          <fieldNames>
+            <version>[ignore]</version>
+            <logValue>[ignore]</logValue>
```
drivenflywheel, line 61: Remove (See #549)
```suggestion
```

drivenflywheel:
If we think the events we emit we may need to have environment-specific fields, we should consider using the [Service Provider pattern](https://docs.oracle.com/javase/tutorial/ext/basics/spi.html) that leverages the support we have in the `emissary.spi` package. The `ObjectTracing` class could leverage an environment-specific "service" that's responsible for constructing the event-specific/stage-specific entries.  The [SPILoader](https://github.com/NationalSecurityAgency/emissary/blob/2d36dd7e50a950dd64f3cf2ca2d8618c3de3ba75/src/main/java/emissary/spi/SPILoader.java#L16) bootstrapping would allow us to select the specific service implementation to use at runtime, but the ObjectTracing class would internally just be coded against the service interface.  If that's too much for this initial PR, maybe we can address it in a follow-up PR.

src/main/java/emissary/util/ObjectTracing.java
```diff
@@ -0,0 +1,27 @@
...
+     * @param d the IBDO
+     * @return the map of log entries to add to the log
+     */
+    public Map<String, String> createTraceMessageMap(IBaseDataObject d) {
+
+        Map<String, String> jsonMap = new HashMap<>();
+
+        // add our fields
+        jsonMap.put("inputFileName", d.getFilename());
+        jsonMap.put("stage", "PickUpPlace");
```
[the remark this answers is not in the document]
arp-0984: I agree using the place name probably isn't the best here. Using stage names like PickUp, DropOff might be more clear for us to understand exactly where in processing we are. Since right now we are only adding this logging for the "PickUp" stage, we can always modify or add to the names later if we would like.

arp-0984:
> If we think the events we emit we may need to have environment-specific fields, we should consider using the [Service Provider pattern](https://docs.oracle.com/javase/tutorial/ext/basics/spi.html) that leverages the support we have in the `emissary.spi` package. The `ObjectTracing` class could leverage an environment-specific "service" that's responsible for constructing the event-specific/stage-specific entries. The [SPILoader](https://github.com/NationalSecurityAgency/emissary/blob/2d36dd7e50a950dd64f3cf2ca2d8618c3de3ba75/src/main/java/emissary/spi/SPILoader.java#L16) bootstrapping would allow us to select the specific service implementation to use at runtime, but the ObjectTracing class would internally just be coded against the service interface. If that's too much for this initial PR, maybe we can address it in a follow-up PR.

I'm investigating this, not sure if it is best to do in this PR or a follow-up PR.

drivenflywheel:
> I'm investigating this, not sure if it is best to do in this PR or a follow-up PR. 
Follow-up PR is probably the right choice.  I just wanted to capture the potential approach.

src/main/java/emissary/pickup/PickUpPlace.java
```diff
@@ -180,6 +184,13 @@ protected void configurePickUpPlace() {
         }
 
         ALWAYS_COPY_METADATA_VALS = configG.findEntriesAsSet("ALWAYS_COPY_METADATA");
+
+        // Setup objectTraceLogger
+        useObjectTraceLogger = configG.findBooleanEntry("USE_OBJECT_TRACE_LOGGER", useObjectTraceLogger);
```
drivenflywheel, line 189: My gut feeling is that we'll migrate away from enabling on a place-specific basis and will instead use a system-wide toggle such as a system property, environment variable, or startup parameter, with that toggle evaluated inside the ObjectTracing class itself.  Probably a bit more like the `Logger` model where we instrument log statements throughout the code and let the central configuration dictate whether that statement is ignored.

src/main/java/emissary/pickup/PickUpPlace.java
```diff
@@ -552,6 +563,12 @@ protected boolean processDataObject(IBaseDataObject d, String fixedName, File th
         dataObjectCreated(d, theFile);
         logger.info("**Deploying an agent for {} and object {} forms={} simple={}", fixedName, d.getInternalId(), d.getAllCurrentForms(),
                 (simpleMode ? "simple" : ""));
+
+        // If object tracing log that agent is being deployed for fixedName (filename)
```
drivenflywheel, line 567: Are we sure this is where the PickUp event should be emitted?  For input files containing multiple sessions, this will emit multiple PickUp events that vary only by timestamp unless we're also emitting additional IBDO fields.

cfkoehler: My understanding of the requirements is that we needed a message or trace at the object level. Other attributes will be added to the message by overriding the `emitLifecycleEvent()` method. If that makes sense?

drivenflywheel (approved):

cfkoehler (approved):

PR #532, opened 2023-08-07
Migrate HttpClient from 4.x to 5.x

fbruton:
`< before change`
`> after change`
```
46,48c46,48
< [INFO] +- org.apache.httpcomponents:httpclient:jar:4.5.13:compile
< [INFO] |  +- org.apache.httpcomponents:httpcore:jar:4.4.15:compile
< [INFO] |  \- commons-logging:commons-logging:jar:1.2:compile
---
> [INFO] +- org.apache.httpcomponents.client5:httpclient5:jar:5.2.1:compile
> [INFO] |  +- org.apache.httpcomponents.core5:httpcore5:jar:5.2.1:compile
> [INFO] |  \- org.apache.httpcomponents.core5:httpcore5-h2:jar:5.2:compile

```

### DIFF
--- a/src/main/java/emissary/client/EmissaryClient.java
+++ b/src/main/java/emissary/client/EmissaryClient.java
@@ -11,6 +11,7 @@ import org.apache.hc.client5.http.auth.StandardAuthScheme;
 import org.apache.hc.client5.http.auth.UsernamePasswordCredentials;
 import org.apache.hc.client5.http.classic.methods.HttpPost;
 import org.apache.hc.client5.http.classic.methods.HttpUriRequestBase;
+import org.apache.hc.client5.http.config.ConnectionConfig;
 import org.apache.hc.client5.http.config.RequestConfig;
 import org.apache.hc.client5.http.cookie.Cookie;
 import org.apache.hc.client5.http.entity.EntityBuilder;
@@ -47,7 +48,7 @@ public class EmissaryClient {
     public static final String DEFAULT_CONTEXT = "emissary";
     public static final String JETTY_USER_FILE_PROPERTY_NAME = "emissary.jetty.users.file";
     public static final int DEFAULT_CONNECTION_TIMEOUT = (int) TimeUnit.MINUTES.toMillis(100L); // 2 X 50 min
-    public static final int DEFAULT_CONNECTION_MANAGER_TIMEOUT = (int) TimeUnit.MINUTES.toMicros(2L);
+    public static final int DEFAULT_CONNECTION_MANAGER_TIMEOUT = (int) TimeUnit.MINUTES.toMillis(2L);
     public static final int DEFAULT_SO_TIMEOUT = (int) TimeUnit.MINUTES.toMillis(1L);
     public static final int DEFAULT_RETRIES = 3;
     public static final String DEFAULT_USERNAME = "emissary";
@@ -64,6 +65,7 @@ public class EmissaryClient {
 
     private static CloseableHttpClient staticClient = null;
     private static RequestConfig staticRequestConfig = null;
+    private static ConnectionConfig staticConnectionConfig = null;
 
     // static config variables
     public static String CONTEXT = DEFAULT_CONTEXT;
@@ -78,8 +80,9 @@ public class EmissaryClient {
     // class is thread-safe
     protected static final AuthCache AUTH_CACHE = new BasicAuthCache();
 
-    private CloseableHttpClient client = null;
-    private RequestConfig requestConfig = null;
+    private CloseableHttpClient client;
+    private RequestConfig requestConfig;
+    private ConnectionConfig connectionConfig;
 
     static {
         configure();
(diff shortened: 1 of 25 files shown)


drivenflywheel (approved):
LGTM 🚀

src/main/java/emissary/client/HTTPConnectionFactory.java
```diff
@@ -96,7 +95,8 @@ private HTTPConnectionFactory() {
             final Configurator cfg = config == null ? ConfigUtil.getConfigInfo(HTTPConnectionFactory.class) : config;
             // if someone doesn't want keep alives...
             if (!cfg.findBooleanEntry(CFG_HTTP_KEEPALIVE, DFLT_KEEPALIVE)) {
-                this.connReuseStrategy = NoConnectionReuseStrategy.INSTANCE;
+                this.connReuseStrategy = DefaultConnectionReuseStrategy.INSTANCE;
```
ldhardy, line 98: This change is probably fine, but I'm just noting here.
We're moving from a "A strategy that never re-uses a connection" posture to a more subtle re-use policy, where sometimes connections will be reused.  Again, I think it's probably safe to make this change but if it introduces problems, they will probably be tricky to track down to this change.  The strategy we changed to does the following:

```
Default implementation of a strategy deciding about connection re-use. The strategy determines whether a connection is persistent or not based on the message’s protocol version and Connection header field if present. Connections will not be re-used and will close if any of these conditions is met

- the close connection option is present in the request message
- the response message content body is incorrectly or ambiguously delineated
- the close connection option is present in the response message
- If the received protocol is HTTP/1.0 (or earlier) and keep-alive connection option is not present

In the absence of a Connection header field, the non-standard but commonly used Proxy-Connection 
header field will be used instead. If no connection options are explicitly given the default policy for the 
HTTP version is applied.  HTTP/1.1 (or later) connections are re-used by default. HTTP/1.0 (or earlier)
 connections are not re-used by default.
```

src/main/java/emissary/server/mvc/adapters/WorkSpaceAdapter.java
```diff
@@ -46,7 +47,8 @@ public EmissaryResponse outboundOpenWorkSpace(final String place, final String s
         nvps.add(new BasicNameValuePair(SPACE_NAME, space));
         method.setEntity(new UrlEncodedFormEntity(nvps, StandardCharsets.UTF_8));
         // set a timeout in case a node is unresponsive
-        method.setConfig(RequestConfig.custom().setConnectTimeout(60000).setSocketTimeout(60000).build());
+        method.setConfig(RequestConfig.custom().setConnectTimeout(Timeout.ofMilliseconds(60000)).setConnectionKeepAlive(Timeout.ofMilliseconds(60000))
```
ldhardy, line 50: Another deprecated custom.setConnectTimeout

src/test/java/emissary/client/EmissaryClientTest.java
```diff
@@ -60,9 +61,9 @@ void testDefaultRequestConfigFromClasspath() {
         RequestConfig requestConfig = client.getRequestConfig();
         // asserted values are copied from src/main/resource/emissary/client/EmissaryClient.cfg, which is packaged
         // with the jar
-        assertEquals(TimeUnit.MINUTES.toMillis(10), requestConfig.getConnectTimeout());
-        assertEquals(TimeUnit.MINUTES.toMillis(5), requestConfig.getConnectionRequestTimeout());
-        assertEquals(TimeUnit.SECONDS.toMillis(90), requestConfig.getSocketTimeout());
+        assertEquals(Timeout.ofMilliseconds(TimeUnit.MINUTES.toMillis(10)), requestConfig.getConnectTimeout());
+        assertEquals(Timeout.ofMicroseconds(TimeUnit.MINUTES.toMillis(5)), requestConfig.getConnectionRequestTimeout());
```
ldhardy, line 65: A seeming inconsistency here between milli and micro?

src/main/java/emissary/client/EmissaryClient.java
```diff
@@ -124,22 +126,22 @@ protected static void configure() {
...
         }
 
         staticRequestConfig =
-                RequestConfig.custom().setConnectTimeout(connectionTimeout)
-                        .setConnectionRequestTimeout(connectionManagerTimeout)
-                        .setSocketTimeout(socketTimeout)
-                        .setTargetPreferredAuthSchemes(Collections.singleton(AuthSchemes.DIGEST))
-                        .setProxyPreferredAuthSchemes(Collections.singleton(AuthSchemes.DIGEST))
+                RequestConfig.custom().setConnectTimeout(Timeout.ofMilliseconds(connectionTimeout))
+                        .setConnectionRequestTimeout(Timeout.ofMicroseconds(connectionManagerTimeout))
```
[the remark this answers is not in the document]
fbruton: The original code set microseconds, but there is a comment that states milliseconds. That leads me to believe that setting to microseconds was by accident. I can update to milliseconds to be consistent.

ldhardy (approved):

jpdahlke (approved):

jpdahlke:
- [ ] confirm availability of artifacts

PR #534, opened 2023-08-11
Add new Filename Generator with date, node, and filter

cfkoehler:
This is similar to https://github.com/NationalSecurityAgency/emissary/blob/master/src/main/java/emissary/output/io/DateStampFilenameGenerator.java

But instead of trying to update that class to what we need figured I would create a new one. I can deprecate the old one if we want. Or just leave it alone

### DIFF
--- a/src/main/java/emissary/output/io/DateFilterFilenameGenerator.java
+++ b/src/main/java/emissary/output/io/DateFilterFilenameGenerator.java
@@ -5,20 +5,24 @@ import emissary.util.io.FileNameGenerator;
 
 import org.apache.commons.lang3.StringUtils;
 
-import java.text.SimpleDateFormat;
-import java.util.Date;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.UUID;
 
+/**
+ * Creates a Filename Generator that contains a timestamp (julian day), uuid, node name/host, and filter name. Example
+ * filename with a filter param of json is: 20232231650_7959b045-d895-4b34-bed2-8800b5071dcd_localhost_json
+ */
 public class DateFilterFilenameGenerator implements FileNameGenerator {
 
-    protected static final SimpleDateFormat DATE_PATTERN = new SimpleDateFormat("yyyyDDDHHmm");
+    protected static final DateTimeFormatter DATE_PATTERN = DateTimeFormatter.ofPattern("yyyyDDDHHmm");
     public static final char DELIMITER = '_';
     public static final char DASH = '-';
     private final String filterNamePart;
     private static final String NODE_NAME = System.getProperty(EmissaryNode.NODE_NAME_PROPERTY);
 
     /**
-     * Create a file name generator that contains date,
+     * Create a file name generator that contains date, uuid, node name, filter extension.
      * 
      * @param filterName filter name used to create file extension
      */
@@ -40,7 +44,7 @@ public class DateFilterFilenameGenerator implements FileNameGenerator {
     }
 
     private static String now() {
-        return DATE_PATTERN.format(new Date());
+        return DATE_PATTERN.format(LocalDateTime.now());
     }
 
 }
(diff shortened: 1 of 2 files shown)


src/test/java/emissary/output/io/DateFilterFilenameGeneratorTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.commons.lang3.math.NumberUtils;
+import org.junit.jupiter.api.Test;
+
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class DateFilterFilenameGeneratorTest extends UnitTest {
```
fbruton, line 17:
```suggestion
class DateFilterFilenameGeneratorTest extends UnitTest {
```

src/main/java/emissary/output/io/DateFilterFilenameGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.UUID;
+
+public class DateFilterFilenameGenerator implements FileNameGenerator {
+
+    protected static final SimpleDateFormat DATE_PATTERN = new SimpleDateFormat("yyyyDDDHHmm");
```
fbruton, line 14:
```suggestion
    protected static final DateTimeFormatter DATE_PATTERN = DateTimeFormatter.ofPattern("yyyyDDDHHmm");
```

src/main/java/emissary/output/io/DateFilterFilenameGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+    public String nextFileName() {
+        return createFileName(filterNamePart);
+    }
+
+    public static String createFileName(String extension) {
+        return String.format("%s%s%s%s%s%s", now(), DELIMITER, UUID.randomUUID(), DELIMITER, NODE_NAME, extension);
+    }
+
+    private static String now() {
+        return DATE_PATTERN.format(new Date());
```
fbruton, line 43:
```suggestion
        return DATE_PATTERN.format(LocalDateTime.now());
```

fbruton (changes requested):

src/main/java/emissary/output/io/DateFilterFilenameGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+import emissary.directory.EmissaryNode;
+import emissary.util.io.FileNameGenerator;
+
+import org.apache.commons.lang3.StringUtils;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.UUID;
+
+public class DateFilterFilenameGenerator implements FileNameGenerator {
```
drivenflywheel, line 12: Don't really love this class name, but it'll do for now.  Should describe the generated filename _(maybe with an example?)_ in a Javadoc comment on the class itself or constructor.

drivenflywheel (changes requested):
Minor Javadoc cleanup

src/main/java/emissary/output/io/DateFilterFilenameGenerator.java
cfkoehler: I am stuck on a better name. Unless it was going to be really long to be descriptive.
But definitely will add a java doc

fbruton (approved):

src/main/java/emissary/output/io/DateFilterFilenameGenerator.java
drivenflywheel: Again, not in scope for this PR, but we could more filename generators to a child package and then remove the redundant `FilenameGenerator` suffix so the names can have sufficient info without being entire sentences. Alternatively, we could invert the generators so they take a provided NamingStrategy.  (Not sure that last one will always work.)

drivenflywheel (approved):

PR #536, opened 2023-08-15
update release action to check suffix is not specified for a release or patch

dev-mlb:


drivenflywheel (approved):
Not sure the PR title  / commit message exactly matches the implemented changes, but it's close enough to support any future discovery needs.

cfkoehler (approved):

PR #538, opened 2023-08-17
Identify invalid IBDO data/array modifications.

jdcove2:
This PR identifies invalid modifications to the IBDO data/array. This pertains only to IBDO data returned as a byte array since data returned as a ChannelFactory is immutable. An invalid modification is when a byte array is returned, the elements are modified and IBDO.setData(...)/IBDO.setChannelFactory(...) is NOT called. The problem occurs when the IBDO data is backed by a channel factory, a byte array is created and returned from the ChannelFactory, the byte array is modified and then the place competes with the byte array being garbage collected without the modifications being given back to the IBDO.

### DIFF
--- a/src/main/config/emissary.core.SafeUsageChecker.cfg
+++ b/src/main/config/emissary.core.SafeUsageChecker.cfg
@@ -1,0 +1,1 @@
+ENABLED = TRUE
--- a/src/main/java/emissary/core/BaseDataObject.java
+++ b/src/main/java/emissary/core/BaseDataObject.java
@@ -182,6 +182,8 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
      */
     protected SeekableByteChannelFactory seekableByteChannelFactory;
 
+    final SafeUsageChecker safeUsageChecker = new SafeUsageChecker();
+
     protected enum DataState {
         NO_DATA, CHANNEL_ONLY, BYTE_ARRAY_ONLY, BYTE_ARRAY_AND_CHANNEL
     }
@@ -218,6 +220,18 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
             } else {
                 return DataState.BYTE_ARRAY_AND_CHANNEL;
             }
+        }
+    }
+
+    @Override
+    /**
+     * {@inheritDoc}
+     */
+    public void checkForUnsafeDataChanges() {
+        safeUsageChecker.checkForUnsafeDataChanges();
+
+        if (theData != null) {
+            safeUsageChecker.recordSnapshot(theData);
         }
     }
 
@@ -327,6 +341,10 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
         Validate.notNull(sbcf, "Required: SeekableByteChannelFactory not null");
         this.theData = null;
         this.seekableByteChannelFactory = sbcf;
+
+        // calls to setData clear the unsafe state by definition
+        // reset the usage checker but don't capture a snapshot until someone requests the data in byte[] form
+        safeUsageChecker.reset();
     }
 
     /**
@@ -378,7 +396,12 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
                 return theData;
             case CHANNEL_ONLY:
                 // Max size here is slightly less than the true max size to avoid memory issues
-                return SeekableByteChannelHelper.getByteArrayFromBdo(this, MAX_BYTE_ARRAY_SIZE);
+                final byte[] bytes = SeekableByteChannelHelper.getByteArrayFromBdo(this, MAX_BYTE_ARRAY_SIZE);
+
+                // capture a reference to the returned byte[] so we can test for unsafe modifications of its contents
+                safeUsageChecker.recordSnapshot(bytes);
+
+                return bytes;
             case NO_DATA:
             default:
                 return null; // NOSONAR maintains backwards compatibility
@@ -391,11 +414,10 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
     @Override
     public void setData(@Nullable final byte[] newData) {
         this.seekableByteChannelFactory = null;
-        if (newData == null) {
-            this.theData = new byte[0];
-        } else {
-            this.theData = newData;
-        }
+        this.theData = newData == null ? new byte[0] : newData;
+
+        // calls to setData clear the unsafe state by definition, but we need to capture a new snapshot
+        safeUsageChecker.resetCacheThenRecordSnapshot(theData);
     }
 
     /**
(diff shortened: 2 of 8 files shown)


src/main/java/emissary/core/BaseDataObject.java
```diff
@@ -221,6 +223,21 @@ protected DataState getDataState() {
         }
     }
 
+    @Override
+    public void checkAndResetArrayHashMap(final String placeName) {
+        for (Map.Entry<byte[], String> e : arrayHashMap.entrySet()) {
+            if (!ByteUtil.sha256Bytes(e.getKey()).equals(e.getValue())) {
```
drivenflywheel, line 229: I think this can only ever be true if the DataState is BYTE_ARRAY_ONLY (i.e., `theData` != null), because that's the only scenario for which the byte[] stored in the Map key will have any live references. In the CHANNEL_ONLY scenario, the Map entries are essentially immutable so the hash of the key will ALWAYS equal it's cached has value.

drivenflywheel: Ouch.  This isn't true;  in the CHANNEL_ONLY scenario, every call to `bdo.data()` will create a new entry in the map, with the key being the returned byte[]. The byte[] object references will differ even it the bytes in the array are identical, so we'll store multiple copies of the payload bytes in the Map if `bdo.data()` is called multiple times within the same place.

drivenflywheel:
See commit 7e901af610a8d13f87bc1d6bd441c14d6e1a81dc on the PR_538_extract_SafeUsageChecker branch for a suggested refactor to help isolate the safe usage checks.

src/test/java/emissary/core/BaseDataObjectTest.java
```diff
@@ -1319,4 +1322,137 @@ void testExtractedRecordClone() {
             fail("Clone method should have been called", ex);
         }
     }
+
+    @Test
+    void testChannelFactoryInArrayOutNoSet() throws IOException {
```
drivenflywheel, line 1327: I'm not sure the names of these 3 new tests provide any insight into what they're trying to validate.  If there's too much to convey via the method name(s), we should add some clarifying comments, either inline or Javadoc

src/test/java/emissary/test/core/junit5/LogbackTester.java
```diff
@@ -0,0 +1,55 @@
...
+        logger = (Logger) LoggerFactory.getLogger(name);
+        appender = new ListAppender<>();
+
+        appender.setContext(logger.getLoggerContext());
+        appender.start();
+        logger.addAppender(appender);
+        logger.setAdditive(false);
+    }
+
+    public void checkLogList(final Level[] levels, final String[] messages, final boolean[] throwables) {
```
drivenflywheel, line 33: Maybe a future refactor, but this seems like an awkward interface. If we were to define a new class - say, `SimplifedLogEvent` - this method instead could take a `List<SimplifiedLogEvent>` and could eliminate all or most of the `Validate.xxx` calls 

Maybe something similar to...
```java
static class SimplifiedLogEvent {
   final Level level;
   final String message;
   final Optional<? extends Throwable> throwable;

   SimplifiedLogEvent (Level level, String message, @Nullable Throwable throwable) {
      ...
   }
  
}
```

drivenflywheel (changes requested):
Overall approach looks good.   Could use a little cleanup to improve readability and maintainability.

drivenflywheel (changes requested):
More cleanup recommendations.  All are cosmetic, aimed at improving maintainability.

drivenflywheel (dismissed):
Looks great 🚀

smcgrattan (approved):

jpdahlke:
Should this be feature flagged and defaulted to off to begin integration?

jpdahlke (dismissed):
Awaiting response on feature-flag feedback.

drivenflywheel:
> Should this be feature flagged and defaulted to off to begin integration?

Should be relatively easy to add a SafeUsageChecker.cfg file with an `ENABLED = "false"` entry.  That flag could no-op most SafeUsageChecker calls and force `checkForUnsafeDataChanges(..)` to return false.

drivenflywheel (dismissed):

src/test/java/emissary/core/SafeUsageCheckerTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.Arrays;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class SafeUsageCheckerTest extends UnitTest {
+    @Test
+    void testDifferentConfigs() {
+        assertTrue(SafeUsageChecker.ENABLED_FROM_CONFIGURATION, "Enabled from config file should be false");
```
drivenflywheel, line 16:
```suggestion
        assertTrue(SafeUsageChecker.ENABLED_FROM_CONFIGURATION, "Enabled from config file should be true");
```

src/main/java/emissary/core/SafeUsageChecker.java
```diff
@@ -0,0 +1,110 @@
...
+            recordSnapshot(bytes);
+        }
+    }
+
+    /**
+     * Uses the snapshot cache to determine whether any of the byte arrays have unsaved changes
+     * 
+     * @return boolean indication of unsafe changes
+     */
+    public boolean checkForUnsafeDataChanges() {
```
drivenflywheel, line 97: Optional, but I think we should choose a method name that implies a `boolean` result.  `checkForUnsafeDataChanges` is a bit ambiguous.
```suggestion
    public boolean unsafeDataChangesDetected() {
```

src/main/config/emissary.core.SafeUsageChecker.cfg
```diff
@@ -0,0 +1 @@
+ENABLED = TRUE
```
drivenflywheel, line 1: To minimize impact, should we initially default to disabled and override downstream, potentially with a flavored config?

src/main/java/emissary/core/SafeUsageChecker.java
```diff
@@ -0,0 +1,110 @@
...
+ * setChannelFactory(SeekableByteChannelFactory)} method.
+ */
+public class SafeUsageChecker {
+    protected static final Logger LOGGER = LoggerFactory.getLogger(SafeUsageChecker.class);
+
+    public static final String ENABLED_KEY = "ENABLED";
+    public static final boolean ENABLED_FROM_CONFIGURATION;
+    public static final String UNSAFE_MODIFICATION_DETECTED = "Detected unsafe changes to IBDO byte array contents";
+
+    static {
```
drivenflywheel, line 29: Suggest adding an explanatory comment to explain the choice of initialization pattern, to discourage any future change that removes the static field and collapses the initialization into the default constructor. 

```suggestion
    // to minimize I/O, we only want to read the config file once regardless of the number of instances created
    static {
```

drivenflywheel (changes requested):
Minor tweaks suggested, notably correcting an assertion message ("true" vs "false")

drivenflywheel (approved):

smcgrattan (approved):
well done

PR #542, opened 2023-08-21
XmlCodec ElementDecoder instances made private and clean API

sambish5:
Moving the XmlCodec ElementDecoder instances to be private, creating type-specific methods for each type of decoder, and creating static map of ElementName -> MethodName, which in turn helps clean the API in IBaseDataObjectXmlHelper and XmlCodecs.


src/main/java/emissary/core/IBaseDataObjectXmlCodecs.java
```diff
@@ -28,10 +29,39 @@
...
+import static emissary.core.constants.IbdoMethodNames.SET_HEADER_ENCODING;
+import static emissary.core.constants.IbdoMethodNames.SET_ID;
+import static emissary.core.constants.IbdoMethodNames.SET_NUM_CHILDREN;
+import static emissary.core.constants.IbdoMethodNames.SET_NUM_SIBLINGS;
+import static emissary.core.constants.IbdoMethodNames.SET_OUTPUTABLE;
+import static emissary.core.constants.IbdoMethodNames.SET_PARAMETER;
+import static emissary.core.constants.IbdoMethodNames.SET_PRIORITY;
+import static emissary.core.constants.IbdoMethodNames.SET_TRANSACTION_ID;
+import static emissary.core.constants.IbdoMethodNames.SET_WORK_BUNDLE_ID;
 import static emissary.core.constants.IbdoXmlElementNames.BASE64_ATTRIBUTE_NAME;
```
drivenflywheel, line 52: While we're cleaning these up, let's make this move this constant into the IBaseDataObjectXmlCodecs class itself and rename it as `BASE64`.  Its value is "base64", it's only used within this class, and it's neither an element name nor an attribute name (it's a value we sometimes assign to the "encoding" attribute).  No Javadoc comment needed for it, it's self-documenting.

src/main/java/emissary/core/IBaseDataObjectXmlCodecs.java
```diff
@@ -40,8 +70,10 @@
 import static emissary.core.constants.IbdoXmlElementNames.PRIORITY;
 import static emissary.core.constants.IbdoXmlElementNames.PROCESSING_ERROR;
 import static emissary.core.constants.IbdoXmlElementNames.SHA256_ATTRIBUTE_NAME;
```
drivenflywheel, line 72: Let's make this move this constant into the IBaseDataObjectXmlCodecs class itself and rename it as `SHA256`.  Its value is "sha256", it's only used within this class, and it's neither an element name nor an attribute name (it's a value we sometimes assign to the "encoding" attribute).  No Javadoc comment needed for it, it's self-documenting.

drivenflywheel (changes requested):
Good overall, would like a few more easy changes to tighten this up.

src/main/java/emissary/core/IBaseDataObjectXmlCodecs.java
sambish5: Along with these two attribute names, I am planning on moving `ENCODING_ATTRIBUTE_NAME` as well, since that is also only used in XmlCodecs, and I feel as it will be the only _attribute name_ remaining, it shouldn't be by itself in the XmlElementName class.

drivenflywheel (approved):
I think this looks great.  Well done! 🚀

PR #546, opened 2023-08-24
Removing pathWeight from all of Emissary

sambish5:
Remove pathWeight instances from code, as well as updating API to no longer show PW in Directories tab.


drivenflywheel (approved):
LGTM 🚀

jpdahlke (approved):

PR #549, opened 2023-09-01
fix object metrics log configuration

jpdahlke:
- resolves issue when `LogObjectMetrics` is not set or null, so the conditional check fails with an error
- removes unknown field for `logLevel` assuming this is the proper list/documentation https://javadoc.io/static/net.logstash.logback/logstash-logback-encoder/4.6/index.html?net/logstash/logback/encoder/LogstashEncoder.html

### DIFF
--- a/src/main/config/logback.xml
+++ b/src/main/config/logback.xml
@@ -21,7 +21,7 @@
   </appender>
 
   <!-- Enable output metrics logging by creating an environment variable of "LogObjectMetrics=true" -->
-  <if condition='${LogObjectMetrics} == true'>
+  <if condition="&quot;${LogObjectMetrics}&quot;.equals(&quot;true&quot;)">
     <then>
       <appender name="OBJECT-METRICS" class="ch.qos.logback.core.rolling.RollingFileAppender">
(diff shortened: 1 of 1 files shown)


cfkoehler (approved):

src/main/config/logback.xml
```diff
@@ -21,15 +21,14 @@
...
-  <if condition='${LogObjectMetrics} == true'>
+  <if condition="&quot;${LogObjectMetrics}&quot;.equals(&quot;true&quot;)">
     <then>
       <appender name="OBJECT-METRICS" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <encoder class="net.logstash.logback.encoder.LogstashEncoder">
           <customFields>{"node":"${emissary.node.name}","port":"${emissary.node.port}","application":"server"}</customFields>
           <includeMdc>false</includeMdc>
           <fieldNames>
             <version>[ignore]</version>
-            <logValue>[ignore]</logValue>
```
cfkoehler, line 32: Whoops. Not sure where I came up with that field from

sambish5 (approved):

PR #552, opened 2023-09-01
jetty 10

dev-mlb:
or #550

### DIFF
--- a/src/main/java/emissary/server/EmissaryServer.java
+++ b/src/main/java/emissary/server/EmissaryServer.java
@@ -25,6 +25,7 @@ import emissary.spi.SPILoader;
 import ch.qos.logback.classic.ViewStatusMessagesServlet;
 import com.google.common.annotations.VisibleForTesting;
 import org.apache.hc.client5.http.classic.methods.HttpGet;
+import org.eclipse.jetty.http.HttpVersion;
 import org.eclipse.jetty.security.ConstraintMapping;
 import org.eclipse.jetty.security.ConstraintSecurityHandler;
 import org.eclipse.jetty.security.HashLoginService;
@@ -61,7 +62,6 @@ import java.net.UnknownHostException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.security.GeneralSecurityException;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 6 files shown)


src/main/java/emissary/server/EmissaryServer.java
fbruton, line 1: Remove throws `GeneralSecurityException` from  the `configureServer()` method.  Code was updated to catch and throw an `IOException`.

fbruton (changes requested):
This transitive dependency is  a bit concerning. Does jersey need to be updated to 3.x?
```
[INFO] +- org.glassfish.jersey.containers:jersey-container-jetty-http:jar:2.40:compile
[INFO] |  +- org.glassfish.hk2.external:jakarta.inject:jar:2.6.1:compile
[INFO] |  +- org.eclipse.jetty:jetty-continuation:jar:9.4.51.v20230217:compile
```

src/main/java/emissary/server/EmissaryServer.java
dev-mlb: Done

dev-mlb:
> This transitive dependency is a bit concerning. Does jersey need to be updated to 3.x?
> 
> ```
> [INFO] +- org.glassfish.jersey.containers:jersey-container-jetty-http:jar:2.40:compile
> [INFO] |  +- org.glassfish.hk2.external:jakarta.inject:jar:2.6.1:compile
> [INFO] |  +- org.eclipse.jetty:jetty-continuation:jar:9.4.51.v20230217:compile
> ```

Yeah, it's unfortunate but this code is deprecated and not in the next version of Jetty due to the new Async stuff. It is just this jar, which has only a few classes in it. If I upgrade to Jersey 3+, the Jakarta conversion starts which would probably be easier to take #550...

jpdahlke:
> > This transitive dependency is a bit concerning. Does jersey need to be updated to 3.x?
> > ```
> > [INFO] +- org.glassfish.jersey.containers:jersey-container-jetty-http:jar:2.40:compile
> > [INFO] |  +- org.glassfish.hk2.external:jakarta.inject:jar:2.6.1:compile
> > [INFO] |  +- org.eclipse.jetty:jetty-continuation:jar:9.4.51.v20230217:compile
> > ```
> 
> Yeah, it's unfortunate but this code is deprecated and not in the next version of Jetty due to the new Async stuff. It is just this jar, which has only a few classes in it. If I upgrade to Jersey 3+, the Jakarta conversion starts which would probably be easier to take #550...

I think I would like to try going forward with Jetty 10 first even with this jar.

cfkoehler (approved):
I was not sure the best way to review this.
I started up a server with and without the new SniHost option and all seems to work as expected.

Also started up multiple emissary as a local cluster and that all seemed correct.

All the pom changes make sense when looking at the dependency tree.

jpdahlke (approved):
I integrated this downstream and don't see any immediate issues.

PR #554, opened 2023-09-05
Fix IBaseDataObjectXmlCodec bug.

jdcove2:
IBaseDataObjectXmlCodec does not currently handle multiple parameter values for a single key correctly when converting XML -> IBDO.

### DIFF
--- a/src/main/java/emissary/core/IBaseDataObjectXmlCodecs.java
+++ b/src/main/java/emissary/core/IBaseDataObjectXmlCodecs.java
@@ -431,7 +431,7 @@ public final class IBaseDataObjectXmlCodecs {
      * object.
      */
     public static final ElementDecoder DEFAULT_STRING_OBJECT_DECODER = (elements, ibdo, ibdoMethodName) -> {
-        final Method method = getIbdoMethod(ibdoMethodName, String.class, Object.class);
+        final Method method = getIbdoMethod(ibdoMethodName, String.class, CharSequence.class);
 
         for (final Element element : elements) {
             final Element nameElement = element.getChild(NAME);
--- a/src/test/java/emissary/core/IBaseDataObjectXmlHelperTest.java
+++ b/src/test/java/emissary/core/IBaseDataObjectXmlHelperTest.java
@@ -19,6 +19,7 @@ import java.nio.ByteBuffer;
 import java.nio.channels.SeekableByteChannel;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Random;
 
@@ -81,7 +82,8 @@ class IBaseDataObjectXmlHelperTest extends UnitTest {
         expectedIbdo.setTransactionId("TransactionId");
         expectedIbdo.setWorkBundleId("WorkBundleId");
         expectedIbdo.putParameter("Parameter1Key", "Parameter1Value");
-        expectedIbdo.putParameter("Parameter2Key", "Parameter2Value");
+        expectedIbdo.putParameter("Parameter2Key", Arrays.asList("Parameter2Value1", "Parameter2Value2"));
+        expectedIbdo.putParameter("Parameter3Key", Arrays.asList(10L, 20L));
         expectedIbdo.addAlternateView("AlternateView1Key", "AlternateView1Value".getBytes(StandardCharsets.ISO_8859_1));
         expectedIbdo.addAlternateView("AlternateView2Key", "AlternateView2Value".getBytes(StandardCharsets.ISO_8859_1));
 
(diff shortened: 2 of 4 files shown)


src/test/java/emissary/core/IBaseDataObjectXmlHelperTest.java
```diff
@@ -81,7 +82,7 @@ void testParentIbdoAllFieldsChanged() throws Exception {
         expectedIbdo.setTransactionId("TransactionId");
         expectedIbdo.setWorkBundleId("WorkBundleId");
         expectedIbdo.putParameter("Parameter1Key", "Parameter1Value");
-        expectedIbdo.putParameter("Parameter2Key", "Parameter2Value");
+        expectedIbdo.putParameter("Parameter2Key", new ArrayList<Object>(Arrays.asList(new Object[] {"Parameter2Value1", "Parameter2Value2"})));
```
drivenflywheel, line 85: Can remove redundant List wrapper 
```suggestion
        expectedIbdo.putParameter("Parameter2Key", Arrays.asList(new Object[] {"Parameter2Value1", "Parameter2Value2"}));
```

drivenflywheel (changes requested):
Verified that this fixes the bug.  One minor change suggested

drivenflywheel (approved):
LGTM 🚀

src/main/java/emissary/core/IBaseDataObjectXmlCodecs.java
```diff
@@ -432,6 +433,7 @@ public ElementEncoders(
      */
     public static final ElementDecoder DEFAULT_STRING_OBJECT_DECODER = (elements, ibdo, ibdoMethodName) -> {
         final Method method = getIbdoMethod(ibdoMethodName, String.class, Object.class);
+        final LinkedListMultimap<String, Object> parameters = LinkedListMultimap.create(100);
```
fbruton, line 436: Another approach is to update `IbdoMethodNames`to  `public static final String SET_PARAMETER = "appendParameter";` and make the following updates. 
```suggestion
final Method method = getIbdoMethod(ibdoMethodName, String.class, CharSequence.class);
```

src/main/java/emissary/core/IBaseDataObjectXmlCodecs.java
```diff
@@ -443,7 +445,11 @@ public ElementEncoders(
             final String valueEncoding = valueElement.getAttributeValue(ENCODING_ATTRIBUTE_NAME);
             final String valueDecoded = new String(extractBytes(valueEncoding, value));
 
-            method.invoke(ibdo, nameDecoded, valueDecoded);
+            parameters.put(nameDecoded, valueDecoded);
+        }
+
+        for (String key : parameters.keySet()) {
+            method.invoke(ibdo, key, parameters.get(key));
```
fbruton, line 452:
```suggestion
            method.invoke(ibdo, nameDecoded, valueDecoded);
```

fbruton (changes requested):

src/main/java/emissary/core/IBaseDataObjectXmlCodecs.java
```diff
@@ -431,7 +431,7 @@ public ElementEncoders(
      * object.
      */
     public static final ElementDecoder DEFAULT_STRING_OBJECT_DECODER = (elements, ibdo, ibdoMethodName) -> {
-        final Method method = getIbdoMethod(ibdoMethodName, String.class, Object.class);
+        final Method method = getIbdoMethod(ibdoMethodName, String.class, CharSequence.class);
```
drivenflywheel, line 434: Hmm..  for parameters, we're encoding `String` _(param name)_ and `Collection<Object>` _(value)_, but now decoding the Objects explicitly as CharSequences, explicitly breaking the `StringObjectEncoder`/`StringObjectDecoder` symmetry. . It may have been implicitly broken already though, since the base StringObjectEncoder calls `toString()` when writing the XML element value.

src/test/java/emissary/core/IBaseDataObjectXmlHelperTest.java
```diff
@@ -81,7 +82,7 @@ void testParentIbdoAllFieldsChanged() throws Exception {
         expectedIbdo.setTransactionId("TransactionId");
         expectedIbdo.setWorkBundleId("WorkBundleId");
         expectedIbdo.putParameter("Parameter1Key", "Parameter1Value");
-        expectedIbdo.putParameter("Parameter2Key", "Parameter2Value");
+        expectedIbdo.putParameter("Parameter2Key", Arrays.asList("Parameter2Value1", "Parameter2Value2"));
```
drivenflywheel, line 85: This test fails if either of the following lines is added;
```java
expectedIbdo.putParameter("Parameter3Key", Arrays.asList(10L, 20L));
expectedIbdo.putParameter("Parameter4Key", Arrays.asList(new Long(10L), new Long(20L)));
```

fbruton (approved):

drivenflywheel (approved):

PR #575, opened 2023-09-10
update httpclient call to use a response handler

dev-mlb:


sambish5 (approved):

src/main/java/emissary/client/EmissaryResponse.java
```diff
@@ -117,5 +119,15 @@ private <T extends BaseEntity> T makeErrorEntity(String msg, Class<T> mapper) {
         return r;
     }
 
+    public static class EmissaryResponseResponseHandler implements HttpClientResponseHandler<EmissaryResponse> {
```
drivenflywheel, line 122: Not that it matters, but why not just `EmissaryResponseHandler`?

drivenflywheel (dismissed):

src/main/java/emissary/client/EmissaryResponse.java
dev-mlb: Because that would make too much sense :P

dev-mlb: Done. Updated to ResponseEmissaryResponseResponseHandlerResponse.

drivenflywheel (approved):
🚀

sambish5 (approved):

PR #580, opened 2023-09-11
remove obsolete language attribute

dev-mlb:


drivenflywheel (approved):
We don't need no stinkin' language! 🚀

PR #586, opened 2023-09-11
Fix ServiceConfigGuide bug.

jdcove2:
ServiceConfigGuide uses java.io.StreamTokenizer to parse the configuration data. StreamTokenizer returns two types of informational tokens, String and Number (i.e. double). ServiceConfigGuide converts Numbers to Strings using Integer.toString(). This limited the range of the number to that of an Integer. A "quick fix" to expand this range is to use Long.toString().


src/test/java/emissary/config/ServiceConfigGuideTest.java
```diff
@@ -305,13 +305,14 @@ void testEntryPrimitives() throws IOException {
                         + "BOOLEANQ = \"BOGUS\"\n" + "SZP = \"123\"\n" + "SZB = \"123b\"\n" + "SZK = \"123k\"\n" + "SZM = \"123m\"\n"
                         + "SZG = \"123g\"\n" + "SZT = \"123t\"\n" + "SZQ = \"red balloons\"\n" + "SZ1 = \"111\"\n" + "SZ2 = \"222\"\n"
                         + "SZ3 = \"333\"\n" + "SZ4 = \"444\"\n" + "SZ5 = \"555\"\n" + "SZ6 = \"666\"\n" + "SZ7 = \"777\"\n" + "SZ8 = \"888\"\n"
-                        + "SZ9 = \"999\"\n" + "SZ0 = \"000\"\n" + "STRING = \"chars\"\n").getBytes();
+                        + "SZ9 = \"999\"\n" + "SZ0 = \"000\"\n" + "STRING = \"chars\"\n" + "LONG2 = 12345678901234").getBytes();
```
drivenflywheel, line 308: Just for awareness:  prior to this PR's change in ServiceConfgGuide.java, the only way to configure a Long value greater than Integer.MAX_VALUE was to surround the value with quotes.  While that technique/workaround worked, it wasn't documented and certainly wasn't something that people would consciously do.

drivenflywheel (approved):

smcgrattan (approved):
LGTM

PR #594, opened 2023-09-20
Add helper to find preferredData by regex and add to Analyzer

fbruton:


src/main/java/emissary/core/IBaseDataObjectHelper.java
```diff
@@ -215,4 +218,40 @@ public static void addParentInformationToChildren(final IBaseDataObject parent,
             }
         }
     }
+
+    /**
+     * Search for the first preferred view by regular expression or use the primary data if none match
+     *
+     * @param payload the payload to pull data from
+     */
+    public static byte[] findPreferredDataByRegex(final IBaseDataObject payload, List<Pattern> preferredViewNamePatterns) {
```
drivenflywheel, line 227: Why prefix this method name with "find" but prefix the non-regex one with "get"?  Aside from the pattern usage, they're provide very parallel functionality, with both falling back to using `data()` is a suitable view isn't present.

src/test/java/emissary/core/IBaseDataObjectHelperTest.java
```diff
@@ -353,4 +356,39 @@ void testAddParentInformationToChildren() {
 
         assertTrue(true);
     }
+
+    @Test
+    void testGetPreferredData() {
```
drivenflywheel, line 361: Test name (`testGetPreferredData`) doesn't match the method-under-test (`findPreferredDataByRegex`)

src/main/java/emissary/analyze/Analyzer.java
```diff
@@ -1,19 +1,25 @@
...
 
 /**
  * Base class for analyzers
  */
 public abstract class Analyzer extends ServiceProviderPlace {
-    protected List<String> PREFERRED_VIEWS;
-    protected String ANALYZED_DATA_NAME;
+    protected List<String> preferredViews;
+    protected String analyzedDataName;
+    protected List<Pattern> perferredViewPatternList;
```
drivenflywheel, line 19: Minor typo.  Might also consider dropping the `List` suffix and just pluralizing to `preferredViewPatterns`. 
```suggestion
    protected List<Pattern> preferredViewPatternList;
```

drivenflywheel (commented):
Minor suggestions, only cosmetic.

src/main/java/emissary/core/IBaseDataObjectHelper.java
fbruton: The "get" method name was a direct copy from the `Analyze` class. I'll update the method name to be consistent.

drivenflywheel (approved):
LGTM  :rocket:

dev-mlb (approved):

PR #597, opened 2023-09-26
update IBaseDataObjectDiffHelperTest to extend unittest

dev-mlb:


drivenflywheel (approved):
Good find.  I wonder if there's a way to automate detection of these cases..

jpdahlke (approved):

PR #598, opened 2023-09-27
addressing unexpected place startups

sambish5:
This logs the invisible places started up during server startup. In the future, we could build out functionality for stopping invisible places that started up unannounced.


cfkoehler (dismissed):
Ran locally and forced the behavior so looks good to me

sambish5:
so a couple of updates to this pr:
- rather than the verify method not returning anything, I've made it return a boolean. Feel this will be useful in the future for wanting Emissary to shutdown IF places are started invisibly, but for now it still at least currently logs invis place startups as well.
- @jdcove2 mentioned moving activeDirPlaces to the method, I currently left that out, as I think it may be useful to elsewhere for shutdown later on, and if not, then I can move it within the method.
- wrote out unit test to test the functionality of verifyNoInvisiblePlaceStarted

cfkoehler (approved):

PR #600, opened 2023-10-02
Using FILE_ABSOLUTEPATH and Original-Filename to populate FILEXT

arp-0984:
Modifying DropOffUtil.extractUniqueFileExtensions() to support full file paths and adding FILE_ABSOLUTEPATH as a field to look in to populate the FILEXT field on the IBDO. The existing behavior only used the Original-Filename field to populate FILEXT. 

As part of this change, I also added a static method DropOffUtil.getBestFilenames() that looks at the IBDO in the FILE_ABSOLUTEPATH and Original-Filename fields and returns all values found in both of those fields.

### DIFF
--- a/src/main/java/emissary/core/constants/Parameters.java
+++ b/src/main/java/emissary/core/constants/Parameters.java
@@ -11,6 +11,7 @@ public class Parameters {
     public static final String INPUT_FILENAME = "INPUT_FILENAME";
     public static final String ORIGINAL_FILENAME = "Original-Filename";
     public static final String SUMMARY = "SUMMARY";
+    public static final String FILE_ABSOLUTEPATH = "FILE_ABSOLUTEPATH";
 
     // Common parameter prefixes
 
--- a/src/main/java/emissary/core/constants/Parameters.java
+++ b/src/main/java/emissary/core/constants/Parameters.java
@@ -5,13 +5,14 @@ public class Parameters {
     // Common parameters
     public static final String DOCUMENT_TITLE = "DocumentTitle";
     public static final String EVENT_DATE = "EventDate";
+    public static final String FILEXT = "FILEXT";
+    public static final String FILE_ABSOLUTEPATH = "FILE_ABSOLUTEPATH";
     public static final String FILE_DATE = "FILE_DATE";
     public static final String FILE_NAME = "FILE_NAME";
     public static final String INPUT_FILEDATE = "INPUT_FILEDATE";
     public static final String INPUT_FILENAME = "INPUT_FILENAME";
     public static final String ORIGINAL_FILENAME = "Original-Filename";
     public static final String SUMMARY = "SUMMARY";
-    public static final String FILE_ABSOLUTEPATH = "FILE_ABSOLUTEPATH";
 
     // Common parameter prefixes
 
--- a/src/main/java/emissary/output/DropOffUtil.java
+++ b/src/main/java/emissary/output/DropOffUtil.java
@@ -8,6 +8,7 @@ import emissary.util.ShortNameComparator;
 import emissary.util.TimeUtil;
 import emissary.util.shell.Executrix;
 
+import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -33,8 +34,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.SimpleTimeZone;
 import java.util.UUID;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 
 import static emissary.core.Form.PREFIXES_LANG;
@@ -85,7 +84,7 @@ public class DropOffUtil {
     private static final String DEFAULT_EVENT_DATE_TO_NOW = "DEFAULT_EVENT_DATE_TO_NOW";
     protected boolean defaultEventDateToNow = true;
 
-    private final Pattern JUST_FILENAME_PATTERN = Pattern.compile("(^|\\\\|/)[^\\\\/]+$");
+    private static List<String> DEFAULT_FILENAME_FIELDS;
 
     /**
      * Create with the default configuration
@@ -149,6 +148,15 @@ public class DropOffUtil {
             this.maxFilextLen = actualConfigG.findIntEntry("MAX_FILEXT_LEN", this.maxFilextLen);
             if (this.maxFilextLen < 0) {
                 this.maxFilextLen = Integer.MAX_VALUE;
+            }
+
+            DEFAULT_FILENAME_FIELDS = new ArrayList<>();
+            List<String> defaultFilenameFields = actualConfigG.findEntries("FILENAME_FIELDS");
+            if (!defaultFilenameFields.isEmpty()) {
+                DEFAULT_FILENAME_FIELDS.addAll(defaultFilenameFields);
+            } else {
+                DEFAULT_FILENAME_FIELDS.add(ORIGINAL_FILENAME);
+                DEFAULT_FILENAME_FIELDS.add(FILE_ABSOLUTEPATH);
             }
         } else {
             logger.debug("Configuration is null for DropOffUtil, using defaults");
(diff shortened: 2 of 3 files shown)


src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -969,25 +974,32 @@ public Date extractEventDateFrom(final IBaseDataObject d, final boolean lastReso
...
+
+            // if what we have is a full filepath, extract just the filename (text after the file path separator)
+            Matcher matcher = JUST_FILENAME_PATTERN.matcher(filename);
+            String justFilename = filename;
+            if (matcher.find()) {
+                justFilename = matcher.group(0);
+            }
+
+            // get the text after the last period
+            if (justFilename.lastIndexOf('.') > -1) {
```
drivenflywheel, line 997: Consider using [FileUtils.getExtension(...)](https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/FilenameUtils.html#getExtension(java.lang.String)) and then applying the range/length filtering.  
```suggestion
           final String fext = FileUtils.getExtension(filename);
           if (StringUtils.isNotBlank(fext) && fext.length() <= this.maxFilextLen) {
               extensions.add(fext);
           } 
```

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -997,6 +1009,31 @@ void extractUniqueFileExtensions(IBaseDataObject p) {
...
+    public static List<String> getBestFilenames(IBaseDataObject d) {
+        String[] fieldsToTry = {ORIGINAL_FILENAME, FILE_ABSOLUTEPATH};
+
+        List<String> filenames = new ArrayList<>();
+
+        for (String ibdoField : fieldsToTry) {
+            if (d.hasParameter(ibdoField)) {
+                for (Object filename : d.getParameter(ibdoField)) {
+                    String stringFileName = (String) filename;
+                    if (StringUtils.isNotEmpty(stringFileName)) {
```
drivenflywheel, line 1028: `isNotBlank` will also protect against null values.  Should be able to explicitly declare in the method's contract (documented in Javadoc) that the returned List will only return non-blank values, so callers will know they can skip null-checks on the returned values.
```suggestion
                    if (StringUtils.isNotBlank(stringFileName)) {
```

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -997,6 +1009,31 @@ void extractUniqueFileExtensions(IBaseDataObject p) {
...
     }
 
+    /**
+     * Checks the Original-Filename and FILE_ABSOLUTEPATH for the filename of the object. Returns a list with of the
+     * non-empty strings found in these fields. If nothing is found in either field, return an empty list.
+     *
+     * @param d The IBDO
+     * @return The list of filenames found in the field Original-Filename or FILE_ABSOLUTEPATH
+     */
+    public static List<String> getBestFilenames(IBaseDataObject d) {
```
drivenflywheel, line 1019: Consider renaming to clarify that the values come from IBDO parameters, not from the BDO's filename field.
```suggestion
    public static List<String> getFullFilePathsFromParams(IBaseDataObject d) {
```

src/test/java/emissary/output/DropOffUtilTest.java
```diff
@@ -601,6 +602,41 @@ void testExtractUniqueFileExtensions() {
         assertTrue(fileExts.contains("mp3"), "FILEXT values should contain \"mp3\"");
     }
 
+    @Test
+    void testExtractFullFilepaths() {
+        DropOffUtil util = new DropOffUtil();
+
+        final IBaseDataObject ibdo = new BaseDataObject();
+        ibdo.setParameter(FILE_ABSOLUTEPATH, "D:\\Users\\jdoe\\Documents\\Taxes 2023.csv");
```
drivenflywheel, line 610: Use a loop or `@ParameterizedTest` for the nearly identical data-driven evaluation.

Also need to make sure we're testing at least one case for which both parameters have populated values; these all test the case for which FILE_ABSOLUTEPATH has a single value.

Should also add a "dot file" case (e.g., " /home/joe/.bashrc") and one for which the extension is longer than the configured max length.

src/test/java/emissary/output/DropOffUtilTest.java
```diff
@@ -601,6 +602,41 @@ void testExtractUniqueFileExtensions() {
         assertTrue(fileExts.contains("mp3"), "FILEXT values should contain \"mp3\"");
     }
 
+    @Test
+    void testExtractFullFilepaths() {
```
drivenflywheel, line 606: Method name doesn't match the logic or name of the method-under-test (`extractUniqueFileExtensions`)

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -997,6 +1009,31 @@ void extractUniqueFileExtensions(IBaseDataObject p) {
...
+        String[] fieldsToTry = {ORIGINAL_FILENAME, FILE_ABSOLUTEPATH};
```
drivenflywheel, line 1020: We should probably make this a DEFAULT_FILENAME_FIELDS constant but allow the "working" list to be overwritten in the `configure(Configurator)` method.

drivenflywheel (changes requested):
This is going to be helpful functionality.

src/main/java/emissary/core/constants/Parameters.java
```diff
@@ -11,6 +11,7 @@ public class Parameters {
     public static final String INPUT_FILENAME = "INPUT_FILENAME";
     public static final String ORIGINAL_FILENAME = "Original-Filename";
     public static final String SUMMARY = "SUMMARY";
+    public static final String FILE_ABSOLUTEPATH = "FILE_ABSOLUTEPATH";
```
drivenflywheel, line 14: As a general practice, we should try to keep these ordered.

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -82,6 +84,8 @@ public class DropOffUtil {
     private static final String DEFAULT_EVENT_DATE_TO_NOW = "DEFAULT_EVENT_DATE_TO_NOW";
     protected boolean defaultEventDateToNow = true;
 
+    private static List<String> DEFAULT_FILENAME_FIELDS;
```
drivenflywheel, line 87: ALL_CAPS naming is usually reserved for constant (`static final`) fields

drivenflywheel (changes requested):
Some minor cleanup needed

src/test/java/emissary/output/DropOffUtilTest.java
```diff
@@ -601,6 +602,36 @@ void testExtractUniqueFileExtensions() {
...
+    void testExtractFileExtensionsWithFullFilepaths() {
+        DropOffUtil util = new DropOffUtil();
+
+        // tests a combination of either FILE_ABSOLUTEPATH and Original-Filename, neither, and both set at once
+        String[] fileAbsolutepaths = {"D:\\Users\\jdoe\\Documents\\Taxes 2023.csv", "", "/paper.abc.zzz",
+                "/home/jdoe/SHARED_D.IR/cat.mov", "/home/user/.bashrc", ""};
+        String[] originalFilenames = {"", "D:\\Users\\jdoe\\interesting.folder\\a.table", "flowers.456.123",
+                "/home/jdoe/SHARED_D.IR/cat", "taxes.thisfileextensionistoolong", ""};
+
+        String[][] extensions = {{"csv"}, {"table"}, {"zzz", "123"}, {"mov"}, {"bashrc"}, {""}};
```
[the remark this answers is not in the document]
arp-0984: Thanks, good catch.

drivenflywheel (dismissed):

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -969,32 +983,63 @@ public Date extractEventDateFrom(final IBaseDataObject d, final boolean lastReso
...
+    }
+
+    /**
+     * Checks the Original-Filename and FILE_ABSOLUTEPATH for the filename of the object. Returns a list with of the
+     * non-empty strings found in these fields. If nothing is found in either field, return an empty list.
+     *
+     * @param d The IBDO
+     * @return The list of filenames found in the field Original-Filename or FILE_ABSOLUTEPATH
+     */
+    public static List<String> getFullFilepathsFromParams(IBaseDataObject d) {
```
fbruton, line 1028: Potential race condition for field`defaultFilenameFields` if the constructor `public DropOffUtil(final Configurator configG)` is called with a different set of configs.

fbruton (changes requested):

src/main/java/emissary/output/DropOffUtil.java
arp-0984: I added using a lock around the array initialization in configure and when we access it here which should ensure we aren't reading it and recreating it at the same time.

fbruton: Adding the lock doesn't address mutability issue of `defaultFilenameFields`, which can be change if a new instance of `DropOffUtil` is instantiated with a different config.

Here is a test explaining the issue:
`
    @Test
    void testModifiedFileNameFields() throws IOException {
        IBaseDataObject ibdo = new BaseDataObject();
        List<String> bestFilenames;

        ibdo.setParameter(ORIGINAL_FILENAME, "");
        ibdo.setParameter(FILE_ABSOLUTEPATH, "");
        bestFilenames = DropOffUtil.getFullFilepathsFromParams(ibdo);
        assertEquals(0, bestFilenames.size(), "No filename should have been found");

        ibdo.setParameter(FILE_ABSOLUTEPATH, "theOtherFile.csv");
        bestFilenames = DropOffUtil.getFullFilepathsFromParams(ibdo);
        assertEquals(1, bestFilenames.size(), "There should be one filename extracted");
        assertEquals("theOtherFile.csv", bestFilenames.get(0), "The FILE_ABSOLUTEPATH should have been extracted");

        ibdo.setParameter(ORIGINAL_FILENAME, "file.docx");
        bestFilenames = DropOffUtil.getFullFilepathsFromParams(ibdo);
        assertEquals(2, bestFilenames.size(), "There should be two filenames extracted");
        assertEquals("file.docx", bestFilenames.get(0), "The Original-Filename should have been extracted");
        assertEquals("theOtherFile.csv", bestFilenames.get(1), "The Original-Filename should have been extracted");

        Configurator config = ConfigUtil.getConfigInfo(DropOffUtil.class);
        config.removeEntry("FILENAME_FIELDS", "*");
        // Adding a field name that is not currently stored in the IBDO
        config.addEntry("FILENAME_FIELDS", "DIFFERENT_FILENAME_FIELD");

        // Contruction of DropOffUtil with a different config will affect the behavior of DropOffUtil.getFullFilepathsFromParams
        new DropOffUtil(config);

        bestFilenames = DropOffUtil.getFullFilepathsFromParams(ibdo);
        // Expect this test to now fail with the updated FILENAME_FIELDS configs
        assertEquals(2, bestFilenames.size(), "There should be two filenames extracted");

    }
    ```

fbruton (changes requested):
Commented with a unit-test exercising issue.

src/main/java/emissary/output/DropOffUtil.java
arp-0984: Thanks for that example. I'm thinking to create a new util class for the two static functions to avoid the issue of multiple DropOffUtil instances being created with different configs.

arp-0984: I resolved this issue by removing the ability to specify these fields in a config file. I overloaded the function getFullFilepathsFromParams so we can optionally provide a list of fields to look in instead of the default fields, in case there is a situation where we need this flexibility.

fbruton (approved):

drivenflywheel (approved):

PR #601, opened 2023-10-02
add run method to Pausable

dev-mlb:


jpdahlke (approved):

cfkoehler (approved):

drivenflywheel:
Sonarqube has way too many bad recommendations. The violated rule isn't applicable to abstract classes, but that doesn't stop it from false positives

PR #609, opened 2023-10-05
Add additional IBDO parameter diff options.

jdcove2:
Currently, IBaseDataObjectDiffHelper returns all of the parameter keys and values for both IBDO's if there is any difference. This can make finding the actually differences difficult if there are a lot of parameters.

This PR does the following:
1. Add an option to display only the keys that differ for both IBDO's.
2. Add an option to display only the keys and values that differ for both IBDO's.
3. Make No. 1 the default behavior instead of the current behavior.

### DIFF
--- a/src/main/java/emissary/core/IBaseDataObjectDiffHelper.java
+++ b/src/main/java/emissary/core/IBaseDataObjectDiffHelper.java
@@ -223,7 +223,7 @@ public class IBaseDataObjectDiffHelper {
      * @param map1 the first map to compare.
      * @param map2 the second map to compare.
      * @param identifier an identifier to describe the context of this map comparison.
-     * @param differences the string list differences are to be added to.
+     * @param differences the string list of differences to be added to.
      */
     public static void diff(final Map<String, byte[]> map1, final Map<String, byte[]> map2, final String identifier,
             final List<String> differences) {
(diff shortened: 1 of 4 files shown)


drivenflywheel:
> 3\. Make No. 1 the default behavior instead of the current behavior.

We're replacing the default behavior with a new behavior; is there a way to explicitly get the current behavior or is the current behavior being abandoned?

drivenflywheel (changes requested):
Minor tweaks suggested

drivenflywheel (dismissed):

src/main/java/emissary/core/IBaseDataObjectDiffHelper.java
```diff
@@ -226,6 +238,44 @@ public static void diff(final Map<String, byte[]> map1, final Map<String, byte[]
         }
     }
 
+    public static void keyValueMapDiff(final Map<String, Collection<String>> parameter1, final Map<String, Collection<String>> parameter2,
```
arp-0984, line 241: Do you think it would be helpful to add javadoc comments for these two functions?

src/main/java/emissary/core/IBaseDataObjectDiffHelper.java
```diff
@@ -226,6 +238,44 @@ public static void diff(final Map<String, byte[]> map1, final Map<String, byte[]
...
+
+        for (Entry<String, Collection<String>> p1Entry : p1Entries) {
+            if (p2Entries.contains(p1Entry)) {
+                p1.remove(p1Entry.getKey());
+                p2.remove(p1Entry.getKey());
+            }
+        }
+
+        if (!p1.isEmpty() || !p2.isEmpty()) {
+            differences.add(String.format("%s%s: %s : %s", identifier, ARE_NOT_EQUAL + "-Differing Keys/Valus", p1, p2));
```
arp-0984, line 256:
```suggestion
            differences.add(String.format("%s%s: %s : %s", identifier, ARE_NOT_EQUAL + "-Differing Keys/Values", p1, p2));
```

arp-0984 (commented):
Only minor comments, overall looks good though.

drivenflywheel (approved):

arp-0984 (approved):

PR #613, opened 2023-10-16
Removing date parsing bug workaround

arp-0984:
There two few issues with date parsing in java 8 that should be resolved in java 11. The point of this ticket is to make sure our date parsing code no longer has issues so we can move over to using the DateTimeFormatter from SimpleDateFormat.

 - FlexibleDateTimeParser has a custom workaround of removing offsets from date strings due to an issue with parsing date strings with a time zone and an offset in java 8. The previous behavior would ignore the offset and use the timezone, so we added a custom solution to strip out the three-letter time zone. Now that we are using java11, the modifications are no longer needed so remove this workaround.

 - http://bugs.java.com/bugdatabase/view_bug.do?bug_id=8032051. Added a test to make sure our code is successfully able to parse dates that have a shorter offset.

### DIFF
--- a/src/test/java/emissary/util/FlexibleDateTimeParserTest.java
+++ b/src/test/java/emissary/util/FlexibleDateTimeParserTest.java
@@ -88,6 +88,12 @@ class FlexibleDateTimeParserTest extends UnitTest {
         test("Mon, 4 Jan 2016 18:20:30 EST +0000", EXPECTED_FULL, pattern);
         test("Mon, 4 Jan 2016 18:20:30EST+0000", EXPECTED_FULL, pattern);
         test("Mon, 4 Jan 2016 18:20:30EST +0000", EXPECTED_FULL, pattern);
+
+        // additional tests with ambiguous abbreviations -- they should continue to be ignored and have consistent
+        // behavior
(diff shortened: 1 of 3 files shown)


fbruton (approved):

src/test/java/emissary/util/FlexibleDateTimeParserTest.java
```diff
@@ -66,14 +66,13 @@ private void test(String date, long expected, DateTimeFormatter formatter) {
     }
 
     /**
-     * Three-letter time zone IDs often point to multiple timezones. Java 8 uses the timezone over the offset causing
-     * problems with the datetime in verifies. Java 9 fixes this issue. Since 9 isn't released and, even if it was, it would
-     * take some time to transition, a regex strips out the short timezone if there is an offset present.
+     * Three-letter time zone IDs often point to multiple timezones. With Java 9+, there are no longer inconsistencies with
```
drivenflywheel, line 69: Should we just remove this entire Javadoc comment?

drivenflywheel (approved):

src/test/java/emissary/util/FlexibleDateTimeParserTest.java
jpdahlke: I think this whole class is on the chopping block if we get the rest of the updates we want in.

PR #617, opened 2023-10-25
Add a newInputStream() to the IBaseDataObject.

jdcove2:
This PR adds a convenience method to the IBaseDataObject to create an InputStream from either the byte[] or SeekableByteChannelFactory that defines the data.

### DIFF
--- a/src/main/java/emissary/core/BaseDataObject.java
+++ b/src/main/java/emissary/core/BaseDataObject.java
@@ -15,10 +15,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.Serializable;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
+import java.nio.channels.Channels;
 import java.nio.channels.SeekableByteChannel;
 import java.rmi.Remote;
 import java.util.ArrayList;
@@ -367,6 +369,23 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
             default:
                 return null;
         }
+    }
+
+    /**
+     * Returns a new InputStream to the data that this BaseDataObject contains.
+     * 
+     * NOTE 1: Mutating the data elements of this IBaseDataObject while reading from the InputStream will have indeterminate
+     * results.
+     * 
+     * NOTE 2: The calling code is responsible for closing the returned InputStream.
+     * 
+     * @return a new InputStream to the data that this BaseDataObject contains.
+     */
+    @Override
+    public InputStream newInputStream() {
+        final SeekableByteChannelFactory sbcf = getChannelFactory();
+
+        return sbcf == null ? null : Channels.newInputStream(sbcf.create());
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/emissary/core/IBaseDataObject.java
```diff
@@ -66,6 +67,18 @@ enum MergePolicy {
      */
     void setChannelFactory(final SeekableByteChannelFactory sbcf);
 
+    /**
+     * Returns a new InputStream to the data that this BaseDataObject contains.
+     *
```
drivenflywheel, line 72: Use `<p>` or `<ul>` + `<li>` markup to preserve the line breaks in the rendered Javadoc

src/main/java/emissary/core/BaseDataObject.java
```diff
@@ -369,6 +371,23 @@ public SeekableByteChannelFactory getChannelFactory() {
         }
     }
 
+    /**
```
drivenflywheel, line 374: Don't duplicate the javadoc comment from the interface method. It will inherit by default, and you can explicitly specify inheritence by using the following:
```java
/**
 * {@inheritDoc}
 */
```    
Duplicating just increases the likelihood of unintended future drift.

src/main/java/emissary/core/IBaseDataObject.java
```diff
@@ -66,6 +67,18 @@ enum MergePolicy {
...
 
+    /**
+     * Returns a new InputStream to the data that this BaseDataObject contains.
+     * 
+     * NOTE 1: Mutating the data elements of this IBaseDataObject while reading from the InputStream will have indeterminate
+     * results.
+     * 
+     * NOTE 2: The calling code is responsible for closing the returned InputStream.
+     * 
+     * @return a new InputStream to the data that this BaseDataObject contains.
```
drivenflywheel, line 78:
```suggestion
     * @return a new stream that reads the data that this object contains, or null if this object has no data
```

drivenflywheel (changes requested):

src/main/java/emissary/core/IBaseDataObject.java
```diff
@@ -66,6 +67,18 @@ enum MergePolicy {
...
+     */
+    InputStream newInputStream();
```
drivenflywheel, line 80: Looking at how we're using this in the unit tests, I'm not loving this method name.  The name `newInputStream` works well on the `Channels` class because that method takes a parameter, but that's not the case here.

I think something like `getDataStream()` does a better job of conveying that this method provides streaming access to the BDO's data.

src/test/java/emissary/core/BaseDataObjectTest.java
```diff
@@ -1406,4 +1409,34 @@ void testArrayInArrayOutNoSet() throws IOException {
             logbackTester.checkLogList(LEVELS_ONE_WARN, ONE_UNSAFE_MODIFICATION_DETECTED, NO_THROWABLES);
         }
     }
+
+    @Test
+    void testNewInputStream() throws IOException {
```
drivenflywheel, line 1414: The variable names in this method hurt my brain.  In the span of 14 actual lines of code, the word "byte" is repeated _26 times_, and it's just too much cognitive noise.

fbruton (approved):

PR #626, opened 2023-11-20
Fixed Initialization Race Condition in DisposeHelperTest

nixon124:
Fixed an initialization race condition that occurred in DisposeHelperTest class.

### DIFF
--- a/src/test/java/emissary/util/DisposeHelperTest.java
+++ b/src/test/java/emissary/util/DisposeHelperTest.java
@@ -9,7 +9,6 @@ import ch.qos.logback.classic.Logger;
 import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.core.read.ListAppender;
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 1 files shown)


src/test/java/emissary/util/DisposeHelperTest.java
```diff
@@ -22,6 +23,9 @@
 
 class DisposeHelperTest {
 
+    static {
```
drivenflywheel, line 26: This solution works, but we should avoid using a static class initialization block and instead use the common JUnit pattern of performing "before all tests" setup in a static method that's annotated with the `org.junit.jupiter.api.BeforeAll` annotation.  _(We should also move this new method to be just above the existing `setup()` method.)_

```suggestion
    @BeforeAll
    static void setupClass() {
```

drivenflywheel (changes requested):
Minor change requested to follow common JUnit practices.

src/test/java/emissary/util/DisposeHelperTest.java
nixon124: Got it.

drivenflywheel (dismissed):
🚀

fbruton (changes requested):
Suggest extending `UnitTest`.

fbruton (approved):

drivenflywheel (approved):
🚀

PR #627, opened 2023-11-27
Modifying the ObjectTrace logger to output a single message for a bundle instead of a message for each session

arp-0984:
Overloading the emitLifecycleEvent method because we do not yet have a IBDO object where we need to emit this message and we do not need it to output the fields we need. The old emitLifecycleEvent can probably be removed once this is fully integrated.

### DIFF
--- a/src/main/java/emissary/util/ObjectTracing.java
+++ b/src/main/java/emissary/util/ObjectTracing.java
@@ -21,14 +21,7 @@ public class ObjectTracing {
     public ObjectTracing() {}
 
     public void emitLifecycleEvent(IBaseDataObject d, ObjectTracing.Stage stage) {
-
-        Map<String, String> jsonMap = new HashMap<>();
-
-        // add our fields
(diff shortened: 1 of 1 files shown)


src/main/java/emissary/util/ObjectTracing.java
```diff
@@ -30,4 +30,15 @@ public void emitLifecycleEvent(IBaseDataObject d, ObjectTracing.Stage stage) {
 
         objectTraceLogger.info(appendEntries(jsonMap), "");
```
fbruton, line 31: Update implementation to call newly created method to reduce duplication:
```
    public void emitLifecycleEvent(IBaseDataObject d, ObjectTracing.Stage stage) {
        
        emitLifecycleEvent(d.getFilename(), stage);
    }
```

fbruton (changes requested):

cfkoehler (approved):
I integrated this with the upstream project and got single output per file which is expected

fbruton (approved):

PR #630, opened 2023-12-04
Using SPI to load environment-specific fields for the objectTrace logger

arp-0984:
Suggested improvement to the objectTrace logger to use a SPI pattern. In order to add an objectTrace log, classes just need to call ObjectTracingService.emitLifecycleEvent(). Any ObjectTrace providers that want to include fields in the log message can be added to the emissary.spi.ObjectTracing file. This is the [SPI reference](https://docs.oracle.com/javase/tutorial/ext/basics/spi.html) that I used to get started.
I am open to discussion/improvements regarding this implementation. 

Also included a few minor improvements to the objectTrace logger:
- Changing the environmental variable from ObjectTrace to LOG_OBJECT_TRACE
- Changing the output folder from logs to localoutput
- Use the environmental value in FilePickUpPlace.cfg instead of a hard-coded value

Includes the commits from #627, and will need to be rebased afterwards


src/main/config/logback.xml
```diff
@@ -21,7 +21,7 @@
   </appender>
 
   <!-- Enable output metrics logging by creating an environment variable of "ObjectTrace=true" -->
```
fbruton, line 23:
```suggestion
  <!-- Enable output metrics logging by creating an environment variable of "LOG_OBJECT_TRACE=true" -->
```

fbruton (changes requested):

arp-0984:
> Ran the following command, but did not find any output in localoutput: `LOG_OBJECT_TRACE=true ./emissary server`

Per our discussion, added ${PROJECT_BASE} to the file path, since the behavior was creating localoutput in the emissary root directory instead of target/localoutput

fbruton (approved):

cfkoehler (approved):

PR #632, opened 2023-12-06
EmissaryServer shutdown when unannounced places are started in strict-mode

sambish5:
If EmissaryServer is started up not in strict-mode, it still will just log unannounced places that are started up.
In strict-mode, the starting of unannounced places will result in EmissaryServer being shutdown.
Added a test to EmissaryServerIT for this, as well as updated StartupTest based on changes made to Startup.

### DIFF
--- a/src/main/java/emissary/server/EmissaryServer.java
+++ b/src/main/java/emissary/server/EmissaryServer.java
@@ -1,5 +1,6 @@
 package emissary.server;
 
+import emissary.admin.Startup;
 import emissary.client.EmissaryClient;
 import emissary.client.EmissaryResponse;
 import emissary.client.HTTPConnectionFactory;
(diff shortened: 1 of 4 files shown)


src/main/java/emissary/server/EmissaryServer.java
```diff
@@ -12,7 +13,6 @@
 import emissary.core.Namespace;
 import emissary.core.NamespaceException;
 import emissary.core.ResourceWatcher;
-import emissary.core.sentinel.Sentinel;
```
drivenflywheel, line 15: Why the Sentinel removal?

sambish5: 100% didn't even know I removed it, completely accidental. Good catch! Added it back in

drivenflywheel (approved):
Verified with local testing and a "registered" coordination place that coordinates the unregistered DelayPlace:
```
2023-12-13 17:32:16,452 localhost-8001  INFO emissary.server.EmissaryServer -   - Server shut down due to invisible place startups on strict-mode: [DelayPlace]

```

dev-mlb (approved):

jpdahlke:
- [ ] waiting for this until after internal changes are made so we can honor this behavior

cfkoehler:
> * [ ]  waiting for this until after internal changes are made so we can honor this behavior

@jpdahlke I believe we are good on this now after last weeks deployment

PR #634, opened 2023-12-14
Add optional log checking to RegressionTest.

jdcove2:
This PR adds optional log checking to RegressionTest. It is disabled by default and enabled by overriding getLogbackLoggerName() and returning the name of the logger to track.


arp-0984 (approved):
Looks reasonable to me

fbruton (changes requested):
Recommend updating an existing regression test to exercise this new feature.

sambish5 (approved):

fbruton (approved):

src/test/java/emissary/test/core/junit5/RegressionTest.java
```diff
@@ -175,6 +199,26 @@ protected void checkAnswersPreHook(final Document answers, final IBaseDataObject
         }
     }
 
+    /**
+     * This method returns the logger name to capture log events from or null if log events are not to be checked.
+     * 
+     * @return the logger name to capture log events from or null (the default) if log events are not to be checked.
+     */
+    protected String getLogbackLoggerName() {
```
drivenflywheel, line 207: We don't _need_ to change this, but this is a clumsy design pattern.  Why override a method if the goal is just to set a String value?  RegressionTest should just read a config file entry and use it to assign a field value.  That pattern is deeply ingrained in the emissary code base. 
To be honest, we should be doing the same with respect to generating answer files.  Requiring a `generateAnswers()` override just to toggle a boolean value is poor design, and makes devs write throwaway code.

```ini
LOGGER_TO_CAPTURE = "name of logger"
GENERATE_ANSWER_FILES = "true"
```

drivenflywheel: Fine, skip the config file, use protected fields in RegressionTest.  This is Object-Oriented Design, but awful encapsulation.

PR #637, opened 2023-12-28
update to Jetty11

jpdahlke:
OBEs https://github.com/NationalSecurityAgency/emissary/pull/550

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,14 +63,13 @@
     <dep.jackson.version>2.13.4</dep.jackson.version>
     <dep.jakarta.activation-api.version>1.2.2</dep.jakarta.activation-api.version>
     <dep.jakarta.servlet-api.version>4.0.3</dep.jakarta.servlet-api.version>
-    <dep.jakarta.xml.bind-api.version>2.3.3</dep.jakarta.xml.bind-api.version>
-    <dep.javax.annotation-api.version>1.3.2</dep.javax.annotation-api.version>
-    <dep.jaxb.runtime.version>2.3.8</dep.jaxb.runtime.version>
+    <dep.jakarta.xml.bind-api.version>3.0.1</dep.jakarta.xml.bind-api.version>
+    <dep.jaxb.runtime.version>3.0.1</dep.jaxb.runtime.version>
     <dep.jdom2.version>2.0.6.1</dep.jdom2.version>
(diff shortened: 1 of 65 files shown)


pom.xml
```diff
@@ -61,16 +61,13 @@
     <dep.httpcore.version>5.2.1</dep.httpcore.version>
     <dep.impsort.version>1.8.0</dep.impsort.version>
     <dep.jackson.version>2.13.4</dep.jackson.version>
-    <dep.jakarta.activation-api.version>1.2.2</dep.jakarta.activation-api.version>
-    <dep.jakarta.servlet-api.version>4.0.3</dep.jakarta.servlet-api.version>
```
jpdahlke, line 65: This needs to get off of the 4.x line and up to 5.x+ to be truly compatible with jetty 11 and other jakarta libraries. It's my understanding that this not only needs to not be 4.x, but it can't be 6.x until Jetty 12 and the actual recommended version to use is the `org.eclipse.jetty.toolchain` library that we previously excluded in the dependency tree so I removed the exclusion on that.

pom.xml
```diff
@@ -61,16 +61,13 @@
     <dep.httpcore.version>5.2.1</dep.httpcore.version>
     <dep.impsort.version>1.8.0</dep.impsort.version>
     <dep.jackson.version>2.13.4</dep.jackson.version>
-    <dep.jakarta.activation-api.version>1.2.2</dep.jakarta.activation-api.version>
```
jpdahlke, line 64: This would be 2.0.1 if it was specified, but removing the override and allowing `jakarta.xml.bind-api` to bring in its own seemed fine.

pom.xml
```diff
@@ -481,8 +455,8 @@
       <artifactId>jersey-mvc</artifactId>
       <exclusions>
         <exclusion>
-          <groupId>javax.servlet</groupId>
-          <artifactId>servlet-api</artifactId>
+          <groupId>jakarta.servlet</groupId>
+          <artifactId>jakarta.servlet-api</artifactId>
```
jpdahlke, line 459: Need to properly exclude the new libraries, javax is no longer asked for and this servlet-api will conflict with the toolchain version from jetty

drivenflywheel (approved):

fbruton (approved):

PR #640, opened 2023-12-29
fix Class raw type warning in Factory.java

jpdahlke:
```
[WARNING] /home/runner/work/emissary/emissary/src/main/java/emissary/core/Factory.java:[63,103] found raw type: java.lang.Class
  missing type arguments for generic class java.lang.Class<T>
```


drivenflywheel (approved):
🔥

PR #641, opened 2023-12-29
address warning for unused TimedResource in t-w-r declarations

jpdahlke:
```
[WARNING] /home/runner/work/emissary/emissary/src/main/java/emissary/core/MobileAgent.java:[343,28] auto-closeable resource timer is never referenced in body of corresponding try statement
[WARNING] /home/runner/work/emissary/emissary/src/main/java/emissary/place/CoordinationPlace.java:[201,32] auto-closeable resource tr is never referenced in body of corresponding try statement
[WARNING] /home/runner/work/emissary/emissary/src/main/java/emissary/core/HDMobileAgent.java:[409,28] auto-closeable resource tr is never referenced in body of corresponding try statement
```

The proposed solution is a bit of a step-back since it does not take advantage of the language feature for AutoCloseable resources. This scenario lives in an edge case where the language doesn't expect a usage where the declared resource is never used in the try block. IntelliJ actually has a nod to this and using a variable name of `ignored` silences the IDE warning. That doesn't work for maven and adding the `-try` lint flag is much too large in scope. 

A different solution would be to add `@SuppressWarnings("unused")` on the method, but that's still a bit heavier in scope than I would like. The code could be refactored to minimize the impact of such an annotation if that is preferred though.

### DIFF
--- a/src/main/java/emissary/core/HDMobileAgent.java
+++ b/src/main/java/emissary/core/HDMobileAgent.java
@@ -406,7 +406,8 @@ public class HDMobileAgent extends MobileAgent {
         logger.debug("In atPlaceHD {} with {} payload items", place, payloadListArg.size());
 
         List<IBaseDataObject> ret = Collections.emptyList();
-        try (TimedResource tr = resourceWatcherStart(place)) {
+        TimedResource tr = resourceWatcherStart(place);
+        try {
             // Process and get back a list of sprouted payloads
             lastPlaceProcessed = place.getDirectoryEntry().getKey();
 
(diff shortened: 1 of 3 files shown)


drivenflywheel:
An alternative approach which seems to also suppress the warning is to use an assertion within the t-w-r body:
```java
assert tr != null; // silence "auto-closeable resource tr is never referenced in body"
```

jpdahlke:
> An alternative approach which seems to also suppress the warning is to use an assertion within the t-w-r body:
> 
> ```java
> assert tr != null; // silence "auto-closeable resource tr is never referenced in body"
> ```

I like this idea better. Thank you.

sambish5 (approved):

drivenflywheel (approved):

PR #642, opened 2024-01-03
Adding Oss profile for maven central publish

cfkoehler:


RELEASING.md
```diff
@@ -105,3 +106,14 @@ Options to use:
...
 ```
 Pushes release artifacts to a repo using `maven deploy`.
+
+## Publishing a Release to Maven Central
+
+Policy does not allow publishing to maven central via Action and is only authorized to be perforemd by specific individuals. 
+
+Using the `oss` profile will make the artifacts available here: [Sonatype/Maven Central](https://central.sonatype.com/artifact/gov.nsa/emissary)
+
+In order for the following profile to work authroization is needed to publish to the gov.nsa maven group as well as having GPG siging configured in maven settings.
```
jpdahlke, line 116:
```suggestion
In order for the following profile to work authorization is needed to publish to the gov.nsa maven group as well as having GPG signing configured in maven settings.
```

pom.xml
```diff
@@ -1364,6 +1364,38 @@
...
       </build>
     </profile>
+    <profile>
+      <id>oss</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-gpg-plugin</artifactId>
+            <version>1.5</version>
```
jpdahlke, line 1374: - [ ] available internally?

pom.xml
```diff
@@ -1364,6 +1364,38 @@
...
+                  <goal>sign</goal>
+                </goals>
+                <phase>verify</phase>
+              </execution>
+            </executions>
+          </plugin>
+          <plugin>
+            <groupId>org.sonatype.plugins</groupId>
+            <artifactId>nexus-staging-maven-plugin</artifactId>
+            <version>1.6.7</version>
```
jpdahlke, line 1388: - [ ] available internally?

pom.xml
```diff
@@ -1364,6 +1364,38 @@
...
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-gpg-plugin</artifactId>
+            <version>1.5</version>
+            <executions>
+              <execution>
+                <id>sign-artifacts</id>
+                <goals>
+                  <goal>sign</goal>
+                </goals>
+                <phase>verify</phase>
```
jpdahlke, line 1381: Is this more a deploy phase goal? I suppose it probably has to be done before the deploy phase?

jpdahlke (changes requested):

pom.xml
cfkoehler: I think technically that is fine. But all the examples show it as being done at the verify stage. 
https://maven.apache.org/plugins/maven-gpg-plugin/usage.html

Maybe because the deployer is "verifying" that the code is signed and authorized 🤷

pom.xml
cfkoehler: yes

pom.xml
cfkoehler: Nope :(
only 3.1.0. I will test if that works for us

jpdahlke (approved):

dev-mlb (approved):

PR #644, opened 2024-01-10
Add ComparisonPlace.

jdcove2:
This PR adds ComparisonPlace. ComparisonPlace runs the configured two places and gives them identical IBDO's. ComparisonPlace then compares the output of the two places and logs any differences.


cfkoehler:
@jdcove2 why is this label as 'test-only' when this is a new place and feature. 
I would have expected to see this labeled as 'feature'. Probably something that will never be run on a production system but still doing more then just creating/modifying java test code.

Happy to be proven wrong. I have not fully digested this PR yet.

src/main/java/emissary/place/ComparisonPlace.java
```diff
@@ -0,0 +1,142 @@
...
+        final String placeBClassName = configG.findStringEntry(PLACE_B_CLASSNAME);
+        final String placeBConfigName = configG.findStringEntry(PLACE_B_CONFIGNAME);
+
+        loggingIdentifier = configG.findStringEntry(LOGGING_IDENTIFIER);
+
+        Validate.notNull(placeAClassName, PLACE_A_CLASSNAME + NOT_NULL_STRING);
+        Validate.notNull(placeAConfigName, PLACE_A_CONFIGNAME + NOT_NULL_STRING);
+        Validate.notNull(placeBClassName, PLACE_B_CLASSNAME + NOT_NULL_STRING);
+        Validate.notNull(placeBConfigName, PLACE_B_CONFIGNAME + NOT_NULL_STRING);
+        Validate.notNull(loggingIdentifier, LOGGING_IDENTIFIER + NOT_NULL_STRING);
```
drivenflywheel, line 64: These validation messages don't really tell the dev how to fix the misconfiguration. "PLACE_A_CLASSNAME == null not allowed!"  is not as instructive as "No configuration entry found with key PLACE_A_CLASSNAME"

src/main/java/emissary/place/ComparisonPlace.java
```diff
@@ -0,0 +1,142 @@
...
+
+        placeA = createPlace(placeAClassName, placeAConfigName);
+        placeB = createPlace(placeBClassName, placeBConfigName);
+
+        Validate.isTrue(placeA.processMethodImplemented == placeB.processMethodImplemented, PROCESS_PROCESSHD_MSG);
+    }
+
+    @Override
+    public List<IBaseDataObject> processHeavyDuty(IBaseDataObject ibdoA) throws ResourceException {
```
drivenflywheel, line 73: We need a comment either here or within the class Javadoc comment clearly detailing the side effects of having this place configured. 
- How does this place impact the IBDO and any children? 
- How should this place be registered with the routing/directory framework? 
  - What about "PlaceA" and "PlaceB"? 
  - Do any service proxies need to be modified?

drivenflywheel (changes requested):
Minor changes requested

drivenflywheel (commented):
Need to rebase.   Unit tests in this PR were broken by PR #664, and merging this PR in its current state will prevent master from building.

drivenflywheel (approved):

rpg36 (approved):
Approving, but bringing up the concern that this place will introduce invisible places. So it might conflict with the work that was done recently to address invisible places.

PR #635, opened 2023-12-19
Added empty string check

SemperFidelisTyrannosaurus:

PR #645, opened 2024-01-12
prevent empty form failure on ending multi-line magic rule

jpdahlke:
Building on the original pr #635 but with the following suggestions:
- updating the commit message to be more contextually descriptive
- separate out the test magic rule from the original file since it's not meant to be leveraged in a real scenario
- rename the custom test artifact to a different suffix so it is ignored by `getMyTestParameterFiles` and remove confusion as to the file's intent
- update hard-coded resource paths to leverage the `ResourceReader` for resilience against class re-naming or package moves

This PR does not target master, but instead the origin PR.


drivenflywheel (approved):
All reasonable changes

SemperFidelisTyrannosaurus (approved):

PR #648, opened 2024-01-16
update error-prone, patterns, and compiler config

jpdahlke:
Updates error-prone and sets us up to hopefully clear the issues and activate this by default to fail builds.

Here are the outstanding issues needing attentions based on this PR:
```
     23 [MemberName]
     13 [SystemOut]
     10 [ReturnMissingNullable]
      8 [FieldMissingNullable]
      5 [YodaCondition]
      4 [NonFinalStaticField]
      4 [MissingSummary]
      4 [ConstantField]
      4 [BooleanParameter]
      3 [CatchingUnchecked]
      2 [UnnecessarilyFullyQualified]
      2 [UngroupedOverloads]
      2 [ThreadPriorityCheck]
      2 [ParameterMissingNullable]
      2 [NotJavadoc]
      2 [CheckedExceptionNotThrown]
      2 [AvoidObjectArrays]
      1 [UnnecessaryParentheses]
      1 [ReturnsNullCollection]
      1 [MissingOverride]
      1 [MethodCanBeStatic]
      1 [JdkObsolete]
      1 [Interruption]
      1 [FieldCanBeFinal]
      1 [EqualsGetClass]
```

PR #652, opened 2024-01-18
Resolving error-prone MemberNames

sambish5:
Using the updates from https://github.com/NationalSecurityAgency/emissary/pull/648, updating error-prone MemberNames throughout the project.


arp-0984 (approved):

jpdahlke:
Note to check if these get referenced internally and the scope of that first.

PR #653, opened 2024-01-19
LoggingPrintStreamTest Build Fixes

nixon124:


sambish5 (approved):

nixon124:
My assumption was that since the Try-with-resources automatically closes any resource at the end of the statement, adding a (resource).close(); statement within it triggered a build warning. I figured using a try finally would resolve the build warnings, but I'm also ok with removing the try statements altogether.

drivenflywheel (approved):

PR #654, opened 2024-01-22
removing SystemOut error-prone warnings

sambish5:
removing error-prone `SystemOut` warnings.
Altered in KeyManipulator main to just use a logger, EmissaryTest now adds the necessary bytes to the `ByteArrayOutputStream` rather than directly writing to `System.out.println`


src/main/java/emissary/directory/KeyManipulator.java
drivenflywheel, line 1: For the `KeyManipulator` class, we should just delete the `main` method.  That method is meant for testing this class from the command line, and Logger references in the `main` method just mean that command classpath now needs more than just the emissary jar:
```bash
emissary (pr/654 $=)]$ java -cp target/emissary-8.0.0-SNAPSHOT.jar emissary.directory.KeyManipulator 
Exception in thread "main" java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory
	at emissary.directory.KeyManipulator.<clinit>(KeyManipulator.java:73)
Caused by: java.lang.ClassNotFoundException: org.slf4j.LoggerFactory
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:525)

```

drivenflywheel (changes requested):

src/main/java/emissary/directory/KeyManipulator.java
sambish5: Figured it would be better to just remove, but wanted to switch to the log and  see what other people thought. I'll delete the main method!

drivenflywheel (approved):

cfkoehler (approved):

PR #668, opened 2024-01-31
1441 - Add Time in Place to Transform History

Kevinnnnnnnnnnnnnnnn:
First working draft, no logic as of yet to enable/disable.

### DIFF
--- a/src/main/java/emissary/core/BaseDataObject.java
+++ b/src/main/java/emissary/core/BaseDataObject.java
@@ -774,6 +774,11 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
     }
 
     @Override
+    public void addTimeInLastPlace(long timeInPlace) {
+        this.history.addTimeInPlace(timeInPlace);
+    }
+
+    @Override
     public void setHistory(TransformHistory newHistory) {
         this.history.set(newHistory);
     }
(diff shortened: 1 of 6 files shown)


fbruton:
I ran across the post about configuring performance metrics.

https://stackoverflow.com/questions/34911347/system-nanotime-for-calculating-execution-time-and-performance-impact
```
private static final boolean ADD_TIME_IN_PLACE_TO_TRANSFORM_HISTORY = true; // change to false while development

public static long nanoTime() {
    return ADD_TIME_IN_PLACE_TO_TRANSFORM_HISTORY ? System.nanoTime() : 0;
}
```

Kevinnnnnnnnnnnnnnnn:
> I ran across the post about configuring performance metrics.
> 
> https://stackoverflow.com/questions/34911347/system-nanotime-for-calculating-execution-time-and-performance-impact
> 
> ```
> private static final boolean ADD_TIME_IN_PLACE_TO_TRANSFORM_HISTORY = true; // change to false while development
> 
> public static long nanoTime() {
>     return ADD_TIME_IN_PLACE_TO_TRANSFORM_HISTORY ? System.nanoTime() : 0;
> }
> ```

Oh yeah, that's not a bad way to do it!  Did you read the other post on that thread though about the cost of calling System.nanoTime() versus the cost of concatenating a number to a string and performing I/O operations?  I thought that was interesting.

Also, my next hurdle is figuring out a good way to let the user turn timing on and off.  An environment variable seems like the obvious solution so that it could be turned on and off without having to restart the server or anything.

drivenflywheel (changes requested):
Good starting point. Definitely needs support for toggling on/off.

PR #669, opened 2024-01-31
Have strict server shutdown when CoordinationPlace place fails to create/DNE

sambish5:
Previously, when a server was started in strict mode, if a place failed to create/did not exist in CoordinationPlace, the build would still succeed. This update causes that failure to shutdown the server.
The server will still build if started without the strict flag, like it normally would.


src/main/java/emissary/admin/Startup.java
```diff
@@ -421,20 +422,27 @@ protected void stopAndWaitForPlaceCreation() {
...
-                        System.exit(1);
-                    }
+                    failedPlaceStartups = true;
+                    String failedPlaceList = String.join("; ", this.failedPlaces);
+                    logger.warn("The following places have failed to start: {}", failedPlaceList);
+                }
+                if (!CoordinationPlace.getFailedCoordinationPlaces().isEmpty()) {
+                    failedPlaceStartups = true;
+                    String failedCoordPlaceList = String.join("; ", CoordinationPlace.getFailedCoordinationPlaces());
+                    logger.warn("The following coordination places have failed to start: {}", failedCoordPlaceList);
```
dev-mlb, line 435: This logs the actual place that failed to start, not the coordination place. Do we want to:
1. Log the coordination place
2. Log the place that failed to start
3. Log both??

sambish5: Something I didn't even think about in the moment (😅 _sigh_). Personally, I think both is probably the best option, so we can point to the coordination place and provide the service coordination place that is explicitly causing the issue.
If you agree I can go in and make that change.

dev-mlb (approved):

cfkoehler (approved):
Looks good. I was able to force the shutdowns during runtime as expected

PR #670, opened 2024-02-01
Adding Convenience Method to Configurator interface / ServiceConfigGuide class

nixon124:


src/main/java/emissary/config/ServiceConfigGuide.java
```diff
@@ -845,6 +846,20 @@ public Map<String, Set<String>> findStringMatchMultiMap(@Nullable final String p
...
 
+    /**
+     * Return the first string entry matching the key parameter
+     *
+     * @param theParameter key to match
+     * @return the first matching value
+     * @throws IllegalArgumentException if no non-blank value is found
+     */
+    @Override
+    public String findRequiredStringEntry(final String theParameter) throws IllegalArgumentException {
```
drivenflywheel, line 857: Please add unit test coverage that includes one or more valid entries, a zero-length entry (`FOO = ""`), and a missing entry

nixon124: Got it.

src/test/java/emissary/config/ServiceConfigGuideTest.java
```diff
@@ -611,4 +611,31 @@ void testFindStringEntry_WithDefaultAndConfiguredNull_Emissary201() {
...
+        config.addEntry(emptyProp, emptyValue);
+
+        // Working entry
+        final String value = config.findRequiredStringEntry(workingProp);
+        assertEquals(value, nonNullValue, "Does this work?");
+
+        // Zero-Length Entry
+        assertThrows(IllegalArgumentException.class, () -> {
+            config.findRequiredStringEntry(emptyProp);
+        });
```
drivenflywheel, line 634: For future reference... The `assertThrows` call actually returns the thrown Exception, so you could capture it and inspect its message.
```suggestion
        Exception ex = assertThrows(IllegalArgumentException.class, () -> config.findRequiredStringEntry(emptyProp));
        assertNotNull(ex.getMessage(), "exception message should not be null");
        assertTrue(ex.getMessage().contains(emptyProp), "Exception message should name the offending config entry");
        
```

drivenflywheel (approved):

drivenflywheel (approved):
🏅

cfkoehler:
@nixon124 pipeline and local build shows that the formatter needs to be run and those changes committed.

drivenflywheel (approved):

jpdahlke (approved):
As a follow-up it would be nice to use this somewhere to illustrate

PR #680, opened 2024-02-05
Replaced SimpleDateFormat with DateTimeFormatter in DropOffUtil

nixon124:


sambish5 (dismissed):

arp-0984 (dismissed):

src/main/java/emissary/output/DropOffUtil.java
```diff
@@ -172,9 +175,8 @@ public String generateBuildFileName() {
     }
 
     private static String getDate(final Date d) {
-        synchronized (DATE_PATTERN) {
-            return DATE_PATTERN.format(d);
-        }
+        LocalDateTime ldt = LocalDateTime.ofInstant(d.toInstant(), ZoneId.systemDefault());
+        return ldt.format(DATE_PATTERN);
```
jpdahlke, line 179: Just so I understand too, does this behavior require the use of LocalDateTime and ZoneId.systemDefault()? I believe we want to use the "current date" for this value and I agree that the data should probably explicitly use the systemDefault zone (which in basically every case is UTC) but does this do something other than:
```
return DATE_PATTERN.format(d.toInstant().atZone(ZoneId.systemDefault());
```
where we still get the use of the system default zone but without the use of LocalDateTime?

jpdahlke (changes requested):

src/main/java/emissary/output/DropOffUtil.java
dev-mlb: Maybe this should move to the TimeUtil class?? Make a method like getDateOrdinalWithTime to replace this method??

nixon124: I can change it to that since they both do the same thing. Is moving the method into the TimeUtil class something that should be done here or should another ticket be created for it?

jpdahlke: If like this code update as it is, I'm good with taking this and then making a new ticket to place it in TimeUtil. I'd also be curious if other places use the same or similar date format. Then we could modify all of those places as well.

arp-0984 (approved):

PR #682, opened 2024-02-06
Resolving error prone YodaConditions

sambish5:
Resolved YodaCondition error prone warnings. Pretty simple fixes only in MobileAgent, but resolves all of the apparent condition warnings.

### DIFF
--- a/src/main/java/emissary/core/MobileAgent.java
+++ b/src/main/java/emissary/core/MobileAgent.java
@@ -434,7 +434,7 @@ public abstract class MobileAgent implements IMobileAgent, MobileAgentMBean {
 
         // Stop looping from occurring
         if (payloadArg.transformHistory().size() > this.MAX_ITINERARY_STEPS &&
-                !ERROR_FORM.equals(payloadArg.currentForm())) {
+                !payloadArg.currentForm().equals(ERROR_FORM)) {
             payloadArg.replaceCurrentForm(ERROR_FORM);
             payloadArg.addProcessingError("Agent stopped due to larger than max transform history size (looping?)");
         }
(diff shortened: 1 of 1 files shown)


sambish5:
@drivenflywheel then for now I will close this PR to look for a different solution.
The other suggested option was something along the lines of `!Objects.equals(payloadArg.currentForm(), ERROR_FORM)` for example at line 437. Either way, I will close this current PR for now.

jpdahlke:
If we would like to have a discussion about deactivating this rule, that's ok too. I don't mind either way. The effort was partly to take the out of the box settings and turn everything on and remove what we didn't like.

drivenflywheel:
> If we would like to have a discussion about deactivating this rule, that's ok too. I don't mind either way. The effort was partly to take the out of the box settings and turn everything on and remove what we didn't like. If error-prone and sonar differ/disagree, we should resolve that as well. I don't think this rule has a sonar equivalent, but you could check.

Sounds like a great idea.  Conceptually, I agree that "harder to read, Yoda code can be", but our code is pretty consistent about using this construct as a concise and effective shorthand to avoid potential NPEs.

sambish5:
> If we would like to have a discussion about deactivating this rule, that's ok too. I don't mind either way. The effort was partly to take the out of the box settings and turn everything on and remove what we didn't like. If error-prone and sonar differ/disagree, we should resolve that as well. I don't think this rule has a sonar equivalent, but you could check.

Sonar does not appear to have an equivalent rule that yells about this. I was not aware that originally this format was used to avoid NPE's, but with that being the case (and with this rule only being yelled about in 4 spots), I think it could be deactivated and not cause any problems.

PR #696, opened 2024-02-12
New method to copy metadata from parent to child with exclusions

smcgrattan:
Adds method to copy metadata from parent to child while specifying parameters not to copy

### DIFF
--- a/src/main/java/emissary/core/IBaseDataObjectHelper.java
+++ b/src/main/java/emissary/core/IBaseDataObjectHelper.java
@@ -19,6 +19,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Pattern;
+import java.util.stream.StreamSupport;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/emissary/core/IBaseDataObjectHelper.java
```diff
@@ -219,6 +220,26 @@ public static void addParentInformationToChildren(final IBaseDataObject parent,
...
+     * @param excludedParametersPattern compiled regular expression specifying parameter names not to copy down
+     */
+
+    public static void addParentInformationToChildExcluding(final IBaseDataObject parentIBaseDataObject,
+            final IBaseDataObject childIBaseDataObject,
+            final Pattern excludedParametersPattern) {
+        Validate.notNull(parentIBaseDataObject, "Required: parentIBaseDataObject not null");
+        Validate.notNull(childIBaseDataObject, "Required: childIBaseDataObject not null");
+        Validate.notNull(excludedParametersPattern, "Required: excludedParametersPattern not null");
+        StreamSupport.stream(parentIBaseDataObject.getParameterKeys().spliterator(), false)
```
cfkoehler, line 238: Java 11 allows this to be simplified if we want.
```suggestion
        parentIBaseDataObject.getParameterKeys().stream()
```

PR #718, opened 2024-03-06
Add IBDO method to ensure data is available.

jdcove2:
This PR adds a method to IBaseDataObject and BaseDataObject that checks whether there is valid data and that it is non-zero in length.

### DIFF
--- a/src/main/java/emissary/core/BaseDataObject.java
+++ b/src/main/java/emissary/core/BaseDataObject.java
@@ -463,6 +463,16 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
     }
 
     /**
+     * Checks if the data is defined with a non-zero length.
+     * 
+     * @return if data is undefined or zero length.
+     */
+    @Override
+    public boolean isDataEmpty() throws IOException {
+        return getDataState() == DataState.NO_DATA || getChannelSize() == 0;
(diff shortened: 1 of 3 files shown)


fbruton (dismissed):

src/main/java/emissary/core/BaseDataObject.java
```diff
@@ -462,6 +462,16 @@ public void setData(@Nullable final byte[] newData, final int offset, final int
...
 
+    /**
+     * Checks if the data is defined with a non-zero length.
+     * 
+     * @return if data is undefined or zero length.
+     */
+    @Override
+    public boolean isDataEmpty() throws IOException {
+        return getDataState() == DataState.NO_DATA || getChannelSize() == 0;
+    }
```
drivenflywheel, line 473: Could we invert this logic and change its name to `hasContent`?  That name seems more intuitive to me, and it keeps the internals internal. (`getChannelSize` already includes the `getDataState()` check, so its inclusion here would be redundant).
```suggestion
    public boolean hasContent() throws IOException {
        return getChannelSize() > 0;
    }
```

drivenflywheel (changes requested):

fbruton (approved):

drivenflywheel (approved):

PR #720, opened 2024-03-13
Remove old ObjectTracing class

arp-0984:
Dependency on downstream project has been removed, so remove this class as it is onw ununsed.


drivenflywheel (approved):
Yeet!

cfkoehler (approved):

PR #721, opened 2024-03-15
moving invis place check out of EmissaryServer

sambish5:
Moving the invis place check out of EmissaryServer. The hope is with this that this will allow the return of the fully `configuredServer` before shutting down if there are invis places. Can also be moved to ServiceCommand, but I felt this location made the most sense.


sambish5:
closing this due to https://github.com/NationalSecurityAgency/emissary/pull/723 actually solving the issue

PR #723, opened 2024-03-20
update startup to use set

dev-mlb:


drivenflywheel:
Do we care at all about the lack of implicit ordering?  I think the answer is "no".

sambish5 (approved):

drivenflywheel (approved):
I think this is good as-is.  _(Note: I have not tested any downstream integration.)_

PR #724, opened 2024-03-20
Add length attribute to RegressionTest XML for SBC and byte[] IBDO fields.

jdcove2:
This PR adds a length attribute to the XML produced by RegressionTest for IBDO fields that have a type of SeekableByteChannel or byte[].

This is the length of the actual data and most likely will be different than the length of the XML field value due to the type of encoding.


src/test/java/emissary/test/core/junit5/RegressionTest.java
```diff
@@ -51,7 +51,7 @@ public abstract class RegressionTest extends ExtractionTest {
     /**
      * The list of actual logEvents generated by executing the place.
      */
-    private List<SimplifiedLogEvent> actualSimplifiedLogEvents;
+    protected List<SimplifiedLogEvent> actualSimplifiedLogEvents;
```
cfkoehler, line 54: Please correct me if I am wrong but I don't see a need for the modifiers in this class to be modified to support the SBC length attribute in the XML

drivenflywheel (approved):

cfkoehler (approved):

PR #727, opened 2024-03-25
Create InterruptibleCharSequence

fbruton:


sambish5 (dismissed):

cfkoehler (approved):
Clean and simple. I was able to integrate upstream and force an interrupt

src/test/java/emissary/core/InterruptibleCharSequenceTest.java
```diff
@@ -0,0 +1,104 @@
...
+            tryMatch(inputWrapped, regex);
+            end = System.currentTimeMillis();
+            logger.info("InterruptibleCharSequence {}", end - start);
+            interruptibleCharSequenceTotalTime += end - start;
+        }
+        float overheadPercentage = (float) interruptibleCharSequenceTotalTime / stringTotalTime;
+        logger.info("InterruptibleCharSequence overhead percentage: {}.", overheadPercentage);
+    }
+
+    public boolean tryMatch(CharSequence input, String regex) {
```
drivenflywheel, line 48: (optional - could make this method static)
```suggestion
    public static boolean tryMatch(CharSequence input, String regex) {
```

src/test/java/emissary/core/InterruptibleCharSequenceTest.java
```diff
@@ -0,0 +1,104 @@
...
+            }
+            t.interrupt();
+            Object result = blockingQueue.take();
+
+            // Assert that InterruptedByTimeoutException was thrown
+            if (result instanceof InterruptedByTimeoutException) {
+                return;
+            } else {
+                fail("failed to interrupt InterruptibleCharSequence with given sleeping intervals");
+            }
```
drivenflywheel, line 98: Simplify?
```suggestion
            assertTrue(result instanceof InterruptedByTimeoutException, "failed to interrupt InterruptibleCharSequence with given sleeping intervals");
```

src/main/java/emissary/core/InterruptibleCharSequence.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public InterruptibleCharSequence(CharSequence inner) {
+        super();
+        this.inner = inner;
+    }
+
+    @Override
+    public char charAt(int index) {
+        if (Thread.currentThread().isInterrupted()) {
+            throw new InterruptedByTimeoutException("Interrupted reading CharSequence");
```
drivenflywheel, line 14: I'm not sure this new exception type is needed: we can observe that the current thread is interrupted, but can't definitely assert that it's due to a timeout.  I think a standard `InterruptedException` might be sufficient.

src/test/java/emissary/core/InterruptibleCharSequenceTest.java
fbruton: The while loop needs to exit or the next iteration will return false. I struggled with making this code more concise and open to other suggestions.

src/main/java/emissary/core/InterruptibleCharSequence.java
fbruton: I believe that any exception thrown from within `charAt` needs to be an instance of a `RuntimeException`. `InterruptedException` only extends `Exception`, which would require that `charAt` declare an exception can be thrown.

src/test/java/emissary/core/InterruptibleCharSequenceTest.java
drivenflywheel: Maybe?  
```java
   assertTrue(... same stuff);
   return;
```

drivenflywheel (approved):

src/test/java/emissary/core/InterruptibleCharSequenceTest.java
fbruton: It's caused the PMD verify to fail the build. Didn't like `return` being the last statement in the loop.

sambish5 (approved):

PR #730, opened 2024-03-29
errorprone :: single instance warnings in some helper and test classes

jdcove2:
This PR fixes error-prone warnings.


drivenflywheel (dismissed):

src/main/java/emissary/core/channels/SeekableByteChannelHelper.java
```diff
@@ -158,7 +158,7 @@ public static int getFromInputStream(final InputStream inputStream, final ByteBu
         // Read direct into buffer's array if possible, otherwise copy through an internal buffer
         final int bytesToRead = byteBuffer.remaining();
         if (byteBuffer.hasArray()) {
-            final int bytesRead = inputStream.read(byteBuffer.array(), byteBuffer.position(), bytesToRead);
+            final int bytesRead = inputStream.read(byteBuffer.array(), byteBuffer.arrayOffset() + byteBuffer.position(), bytesToRead);
```
jpdahlke, line 161: What was flagged to initiate this change?

jpdahlke: I think it's a good fix to make, but I didn't see it in the list on the originating ticket. In the future, feel free to add notes about stuff like this to the PR description to give reviewers a heads up. That can potentially reduce confusion on if the change was intentional.

jpdahlke (approved):

drivenflywheel (approved):

PR #732, opened 2024-04-01
errorprone :: @Nullable

dev-mlb:
ReturnMissingNullable, FieldMissingNullable, & ParameterMissingNullable


sambish5 (approved):

src/main/java/emissary/config/ConfigUtil.java
```diff
@@ -74,28 +75,35 @@ public class ConfigUtil {
     public static final String PROJECT_BASE_ENV = "PROJECT_BASE";
 
     /** The package name where config stuff may be found */
+    @Nullable
```
drivenflywheel, line 78: The fact that we need to put a `@Nullable` annotation on fields that we explicitly initialize as `null`

drivenflywheel (approved):

PR #733, opened 2024-04-02
move to use 'blank space'

jpdahlke:


sambish5 (approved):

src/main/java/emissary/util/DataUtil.java
```diff
@@ -50,15 +50,15 @@ public static boolean isNotEmpty(@Nullable final IBaseDataObject d) {
     }
 
     /**
-     * Return true if the data slot is empty or just whitespace and/or control chars
+     * Return true if the data slot is empty or just blank space and/or control chars
```
drivenflywheel, line 53: This is _(and already was)_ a misleading description of the functionality.  The underlying code doesn't examine individual characters at all:
```
    public static boolean isEmpty(@Nullable final byte[] data) {
        return (data == null) || (data.length == 0);
    }
```

src/main/java/emissary/util/DataUtil.java
```diff
@@ -50,15 +50,15 @@ public static boolean isNotEmpty(@Nullable final IBaseDataObject d) {
...
      */
     public static boolean isEmpty(final IBaseDataObject d) {
         final byte[] data = d.data();
         return isEmpty(data);
     }
 
     /**
-     * Return true if the data is empty or consists only of whitespace or control characters
+     * Return true if the data is empty or consists only of blank space or control characters
```
drivenflywheel, line 61:
```suggestion
     * Return true if the data is null or empty
```

src/main/java/emissary/util/DataUtil.java
```diff
@@ -68,7 +68,7 @@ public static boolean isEmpty(@Nullable final byte[] data) {
     }
 
     /**
-     * Return true if the data slot is empty or just one whitespace character
+     * Return true if the data slot is empty or just one blank space character
```
drivenflywheel, line 71:
```suggestion
     * Return true if the data slot is empty
```

drivenflywheel (changes requested):

sambish5 (approved):

drivenflywheel (approved):

PR #756, opened 2024-04-12
executrix :: binary mode

dev-mlb:

PR #757, opened 2024-04-18
errorprone :: MissingSummary

sambish5:


src/test/java/emissary/test/core/junit5/RegressionTestUtil.java
```diff
@@ -118,7 +118,7 @@ public static Path getTestResx() {
     }
 
     /**
-     * @see ExtractionTest#checkAnswers(Document, IBaseDataObject, List, String)
+     * See {@link ExtractionTest#checkAnswers(Element, IBaseDataObject, List, String)}
```
jpdahlke, line 121:
```suggestion
     * See {@link ExtractionTest#checkAnswers(Document, IBaseDataObject, List, String)}
```
This reference is still valid, right?

sambish5: Good pickup here actually. The method with the original signature does exist, I accidentally had it pointing to the other checkAnswers method. I'll switch it to reference the right one.

jpdahlke (approved):

src/main/java/emissary/util/shell/ReadOutputBuffer.java
```diff
@@ -16,11 +16,6 @@
 import java.io.UnsupportedEncodingException;
 import javax.annotation.Nullable;
 
-/**
- *
- * @author ce
- * @version 1.0
- */
```
drivenflywheel, line 23: May collide with an identical change in #756

src/test/java/emissary/test/core/junit5/RegressionTestUtil.java
```diff
@@ -118,7 +118,7 @@ public static Path getTestResx() {
     }
 
     /**
-     * @see ExtractionTest#checkAnswers(Document, IBaseDataObject, List, String)
+     * See {@link ExtractionTest#checkAnswers(Document, IBaseDataObject, List, String)}
```
drivenflywheel, line 121: This is only a marginally accurate reference.  Yes, it will get errorprone to stop complaining, but there's a fairly significant divergence in what's checked, especially the logged events.

sambish5: The reference could be easily removed if you see that as the best option. I agree it's only marginally accurate

drivenflywheel: I'm OK with us just wiping this javadoc comment for now.  This is a plumbing method that's only used from a few locations,

drivenflywheel (approved):

jpdahlke (approved):

PR #760, opened 2024-04-23
Pull out LoggingInputStream for use with any InputStream.

jdcove2:
This PR pulls out the private LoggingInputStream class from inside the LoggingInputStreamFactory class so that it can be used with any InputStream.


src/main/java/emissary/core/channels/LoggingInputStream.java
```diff
@@ -0,0 +1,135 @@
+package emissary.core.channels;
```
drivenflywheel, line 1: Since this class isn't directly related to channels in any way, we should move it to the `emissary.util.io` package, which has other custom streams such `LoggingPrintStream` and `BufferedWriterOutputStream`
```suggestion
package emissary.util.io;
```

drivenflywheel (changes requested):

drivenflywheel (approved):

cfkoehler (approved):

PR #761, opened 2024-04-24
switch to using ubi as the standard github build with Docker

dev-mlb:


.github/workflows/maven-ci.yml
```diff
@@ -24,24 +24,29 @@ jobs:
...
-          distribution: "corretto"
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
 
-      - name: Build with Maven
-        run: mvn -B -e -ntp "-Dstyle.color=always" -Dmaven.compiler.release=${{ matrix.java-compiler }} verify
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
 
-      - name: Verify Changed Files
```
cfkoehler, line 40: Not sure if we care. But now the only step that checks for changed files from the formatter is the mac build.

cfkoehler:
Not needed here. But believe this changes a few of the steps in DEVELOPING.md for using docker
https://github.com/NationalSecurityAgency/emissary/blob/30f75f32565b0a07f6ef10bd843c441da7026117/DEVELOPING.md?plain=1#L615
Basically no longer needing to run a maven build first since it is done in the image build process.

cfkoehler (approved):

.github/workflows/maven-ci.yml
dev-mlb: I originally had it as a step in the Dockerfile to check, but I don't want it to fail if you're doing local development and testing.

dev-mlb: Separate runner??

cfkoehler: Feels a little overkill. But runners are free. So sure. Unless you think it is too much.

dev-mlb: Feels a little overkill, but sorta nice to have it as a separate runner. We could probably add things to it in the future if we wanted to? @jpdahlke Thoughts?

ldhardy: We've had PRs just to correct a branch that got merged that produces formatting changes.  Personally, I wish the build would fail sooner before the branch gets merged or even the PR gets created. But as long as this will make sure the changes get addressed before merge than it's probably fine to have it separated out.

cfkoehler: @ldhardy 
Here in emissary those formatting changes have blocked the merge since the CI check fails. Don't think we have needed those follow prs recently here. 

But yeah good point.

cfkoehler (dismissed):

jpdahlke:
This seems like a good change to me. Once this is in we can update the settings to require the actions to complete for merge.

jpdahlke:
in favor of #765

PR #764, opened 2024-04-26
shutdown RollManager before ResourceWatcher

sambish5:
There was an issue with roll actions trying to roll metrics after the ResourceWatcher was stopped. This switch should allow metrics to be output, close all rollers, and then the ResourceWatcher will stop.

### DIFF
--- a/src/main/java/emissary/server/EmissaryServer.java
+++ b/src/main/java/emissary/server/EmissaryServer.java
@@ -366,6 +366,12 @@ public class EmissaryServer {
         }
         LOG.info("Done stopping all places");
 
+        try {
+            MetricsManager.lookup().shutdown();
+        } catch (Exception ex) {
+            LOG.warn("No metrics manager available");
+        }
+
         // Print the stats
         try {
             ResourceWatcher rw = ResourceWatcher.lookup();
@@ -373,12 +379,6 @@ public class EmissaryServer {
             rw.quit();
         } catch (Exception ex) {
             LOG.warn("No resource statistics available");
-        }
-
-        try {
-            MetricsManager.lookup().shutdown();
-        } catch (Exception ex) {
-            LOG.warn("No metrics manager available");
         }
 
         SPILoader.unload();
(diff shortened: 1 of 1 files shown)


src/main/java/emissary/server/EmissaryServer.java
```diff
@@ -366,6 +366,12 @@ public static void stopServer(final String name, final boolean force, final bool
         }
         LOG.info("Done stopping all places");
 
+        try {
+            MetricsManager.lookup().shutdown();
+        } catch (Exception ex) {
+            LOG.warn("No metrics manager available");
+        }
+
```
jpdahlke, line 374: I think the RollManager.shutdown(); from 384 needs to move ahead of this block as well. 

Stop places, roll everything, then stop metrics, then stop resource watcher.

sambish5: oops agreed, I'll make that change quick

src/main/java/emissary/server/EmissaryServer.java
```diff
@@ -366,6 +366,14 @@ public static void stopServer(final String name, final boolean force, final bool
...
+        RollManager.shutdown();
```
dev-mlb, line 375: Should the roller go before metrics, or does that cause issues? It shouldn't, right??

jpdahlke: Yeah, I would say rollers should stop right after the places stop.

sambish5: It doesn't cause issues being moved before metrics shutdown. @jpdahlke I can easily add a message saying "Rollers have been closed" if wanted, but with `RollManager.shutdown()` being called, that puts out a log message as is -> `INFO emissary.roll.RollManager -   - Closing all rollers (1)`. If we want two I can add one in, if the existing one is enough  I'll just move RollManager shutdown to before Metrics

jpdahlke: Ah, if there already is one then that's sufficient.

jpdahlke (approved):

dev-mlb (dismissed):

dev-mlb (approved):

jpdahlke (approved):

PR #765, opened 2024-04-27
docker ci build

dev-mlb:
Another way to do #761


jpdahlke:
I think I like this approach better than #761. It seems to reduce some of the duplication between the docker files, specifically with the maven and verify portions.

Should the file live in the .github directory now or does that just make the relative paths too cumbersome? I think it's fine to live under contrib/docker FWIW.

ldhardy (approved):
I like this approach better than the 761 approach.

PR #770, opened 2024-05-03
Updating FlexibleDateTimeFormatter to cover edge cases

arp-0984:
Adding extra date format patterns to cover edge cases and updating characters to remove from dates. 

Attempting these formats is optional and specified with a boolean passed to the parse() function. The basic set of formats (FORMAT_DATETIME_MAIN) are attempted first to parse the date, and if none of them are successful, then the more comprehensive set of date formats (FORMAT_DATETIME_EXTRA) are attempted. 

Testing with the new patterns covers all the edge cases I could find. I attempted to explain the new patterns in the config file. Utilizing DateTimeFormatter is tricky because it requires an exact match of a pattern to the provided date string. The goal here is not to be able to literally parse any date, but just match the behavior of the existing DateTimeFormatterLegacy class.


drivenflywheel (approved):
Comments noted for follow up work

PR #771, opened 2024-05-08
provide more info when SBCF are different

rupert-griffin:
When two `SeekableByteChannelFactory` objects are found to not be the same, the `differences` string is currently appended with ```"<data parameter name> not equal. 1.cs=<byte channel 1 size> 2.cs=<byte channel 2 size>"```. 

This doesn't account for channels of the same size with different data. Finding these differences currently means stepping through the program with the debugger to manually convert each channel's input stream back into a string. Appending the data directly from each object makes this issue much easier to debug.

`differences` was also removed as the error message for `assertNull` in `RegressionTestUtil` because it already prints when the assert fails.

### DIFF
--- a/src/main/java/emissary/core/IBaseDataObjectDiffHelper.java
+++ b/src/main/java/emissary/core/IBaseDataObjectDiffHelper.java
@@ -4,14 +4,12 @@ import emissary.core.channels.SeekableByteChannelFactory;
 import emissary.core.constants.IbdoXmlElementNames;
 
 import org.apache.commons.io.IOUtils;
-import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.Validate;
 
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.channels.Channels;
 import java.nio.channels.SeekableByteChannel;
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -122,8 +120,7 @@ public class IBaseDataObjectDiffHelper {
         final int ibdoList2Size = (ibdoList2 == null) ? 0 : ibdoList2.size();
 
         if (ibdoList1Size != ibdoList2Size) {
-            differences.add(String.format("%s%s%sIBaseDataObject List Size 1: %s%sIBaseDataObject List Size 2: %s",
-                    identifier, ARE_NOT_EQUAL, StringUtils.LF, ibdoList1Size, StringUtils.LF, ibdoList2Size));
+            differences.add(diffNotEqStr(identifier, "IBaseDataObject_list_size", ibdoList1Size, ibdoList2Size));
         } else if (ibdoList1 != null && ibdoList2 != null) {
             final List<String> childDifferences = new ArrayList<>();
             for (int i = 0; i < ibdoList1.size(); i++) {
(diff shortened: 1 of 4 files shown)


src/main/java/emissary/core/IBaseDataObjectDiffHelper.java
```diff
@@ -157,8 +158,10 @@ public static void diff(final SeekableByteChannelFactory sbcf1, final SeekableBy
                     InputStream is1 = Channels.newInputStream(sbc1);
                     InputStream is2 = Channels.newInputStream(sbc2)) {
                 if (!IOUtils.contentEquals(is1, is2)) {
-                    differences.add(String.format("%s not equal. 1.cs=%s 2.cs=%s",
-                            identifier, sbc1.size(), sbc2.size()));
+                    differences.add(String.format("%s not equal. 1.is=%s 2.is=%s",
```
drivenflywheel, line 161: We definitely don't want to spit out a string representation of the entire payload; that could be well over a GB for each object being compared.

rupert-griffin: That makes sense, I hadn't considered that.

rupert-griffin: Would it make sense to have it display a string representation of the first `x` characters? Where `x` is set to 64 for ease of debugging

drivenflywheel: We could consider making that an option, but it would only be helpful if 1) the payload is text-like and decodes well using UTF_8  2) the difference occurs within the first few bytes.

src/main/java/emissary/core/IBaseDataObjectDiffHelper.java
```diff
@@ -121,7 +122,8 @@ public static void diff(final List<IBaseDataObject> ibdoList1, final List<IBaseD
         final int ibdoList2Size = (ibdoList2 == null) ? 0 : ibdoList2.size();
 
         if (ibdoList1Size != ibdoList2Size) {
-            differences.add(String.format("%s%s: 1.s=%s 2.s=%s", identifier, ARE_NOT_EQUAL, ibdoList1Size, ibdoList2Size));
+            differences.add(String.format("%s%s%sIBaseDataObject List Size 1: %s%sIBaseDataObject List Size 2: %s",
+                    identifier, ARE_NOT_EQUAL, StringUtils.LF, ibdoList1Size, StringUtils.LF, ibdoList2Size));
```
drivenflywheel, line 126: If we insert line breaks, it will break parsing of aggregated messages.

rupert-griffin:
Line breaks and payload string removed from log messaging. Returned to original formatting with slightly adjusted wording for improved clarity.

Before:
`[ERROR]   ExamplePlaceTest>RegressionTest.testExtractionPlace:266->ExtractionTest.testExtractionPlace:114->RegressionTest.checkAnswers:337 com.burrito.transform.ExamplePlace: PDiff: data not equal. 1.cs=64 2.cs=64 ==> expected: <null> but was: <com.burrito.transform.FfmpegPlace: PDiff: data not equal. 1.cs=64 2.cs=64>`

After:
`[ERROR]   ExamplePlaceTest>RegressionTest.testExtractionPlace:266->ExtractionTest.testExtractionPlace:114->RegressionTest.checkAnswers:337 expected: <null> but was: <com.burrito.transform.ExamplePlace: parent_differences: <data> elements are not equal. SeekableByteChannelFactory_1=64_byte_stream : SeekableByteChannelFactory_2=64_byte_stream >`

rupert-griffin:
Line breaks and payload string removed from log messaging. Returned to original formatting with slightly adjusted wording for improved clarity.

Before:
[ERROR]   ExamplePlaceTest>RegressionTest.testExtractionPlace:266->ExtractionTest.testExtractionPlace:114->RegressionTest.checkAnswers:337 com.example.transform.ExamplePlace: PDiff: data not equal. 1.cs=64 2.cs=64 ==> expected: <null> but was: <com.example.transform.FfmpegPlace: PDiff: data not equal. 1.cs=64 2.cs=64>

After:
[ERROR]   ExamplePlaceTest>RegressionTest.testExtractionPlace:266->ExtractionTest.testExtractionPlace:114->RegressionTest.checkAnswers:337 expected: <null> but was: <com.example.transform.ExamplePlace: parent_differences: <data> elements are not equal. SeekableByteChannelFactory_1=64_byte_stream : SeekableByteChannelFactory_2=64_byte_stream >

cfkoehler:
I have lost track of this change. @drivenflywheel or @jdcove2 do you think it is something we still want to do?

rupert-griffin:
Putting this aside.

PR #774, opened 2024-05-13
errorprone :: EmptyCatch

sambish5:


src/test/java/emissary/place/ServiceProviderPlaceTest.java
```diff
@@ -267,6 +267,7 @@ private void runFileConfiguredTest(boolean usePackage, int ctorType) throws IOEx
                 try {
                     fos.close();
                 } catch (IOException ignore) {
+                    // Ignore
```
jpdahlke, line 270: This does solve the error-prone warning, but would it be more beneficial to just convert this try block to a try-with-resources and remove the need to manually close `fos` at all?

jpdahlke (approved):

PR #780, opened 2024-05-19
errorprone :: ThrowSpecificExceptions

dev-mlb:
26 [ThrowSpecificExceptions]

### DIFF
--- a/src/main/java/emissary/core/EmissaryRuntimeException.java
+++ b/src/main/java/emissary/core/EmissaryRuntimeException.java
@@ -1,0 +1,38 @@
+package emissary.core;
+
+/**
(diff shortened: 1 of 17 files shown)


src/main/java/emissary/core/EmissaryRuntimeException.java
```diff
@@ -0,0 +1,38 @@
+package emissary.core;
+
+/**
+ * This is the top of the exception hierarchy for Emissary All emissary exceptions extend from here
```
drivenflywheel, line 4: Not sure what description we want here, but this one's not accurate.  With the addition of this class, I'm not sure it's still true in the  `EmissaryException` class either.

drivenflywheel (approved):

PR #815, opened 2024-06-06
add workflow to cleanup action caches

cfkoehler:
building off of recommendations here: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#managing-caches


jpdahlke:
https://github.com/NationalSecurityAgency/emissary/pull/816

PR #818, opened 2024-06-13
sonarlint :: add unload method to remove ThreadLocal objects

jpdahlke:
https://rules.sonarsource.com/java/RSPEC-5164/

Another somewhat silly warning because having a method and calling that method are two different things, but it doesn't hurt to add a method to provide the capability. I just don't know that it's useful (or necessary to call them) with how we are using these objects since they live for the duration of the application.


fbruton (approved):

drivenflywheel (approved):
🔥

PR #821, opened 2024-06-17
updating RegressionTest generateAnswers for when two classes are used for test .dat & answer files

sambish5:
This updates `generateAnswers` to be able to generate answer files in the correct location when `getMyTestParameterFiles` is passed two separate classes. This allows files to generate/regenerate in the provided answer class, based on .dat files in the other class directory.

### DIFF
--- a/src/test/java/emissary/test/core/junit5/RegressionTestUtil.java
+++ b/src/test/java/emissary/test/core/junit5/RegressionTestUtil.java
@@ -95,6 +95,13 @@ public final class RegressionTestUtil {
      */
     private static final Path TEST_RESX = getTestResx();
 
+    /**
+     * Used for RegressionTest generateAnswer when {@link UnitTest#getMyTestParameterFiles(Class, Class)} has two different
+     * classes, and we need to generate answer files in the ansClass directory
+     */
+    @Nullable
+    private static String ansClass = null;
+
     private RegressionTestUtil() {}
 
     /**
@@ -200,7 +207,14 @@ public final class RegressionTestUtil {
             return null;
         }
 
-        final String xmlPath = resource.substring(0, datPos) + ResourceReader.XML_SUFFIX;
+        String xmlPath;
+        if (ansClass == null) {
+            xmlPath = resource.substring(0, datPos) + ResourceReader.XML_SUFFIX;
+        } else {
+            String ansPath = ansClass.replace(".", "/");
+            int testNamePos = resource.lastIndexOf("/");
+            xmlPath = ansPath + resource.substring(testNamePos, datPos) + ResourceReader.XML_SUFFIX;
+        }
         return TEST_RESX.resolve(xmlPath);
     }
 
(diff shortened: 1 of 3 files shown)


drivenflywheel (changes requested):
This is fundamentally broken - in fact, we'll also need to rework the related changes in the `UnitTest` class because these changes aren't thread-safe, and could corrupt results when concurrently testing multiple classes that extend UnitTest or ExtractionTest.  We can't use a static member in the base class to "store" the answer file paths.

h/t to @jdcove2 for noticing and pointing this out.

src/test/java/emissary/test/core/junit5/RegressionTestUtil.java
```diff
@@ -95,6 +95,13 @@ public final class RegressionTestUtil {
      */
     private static final Path TEST_RESX = getTestResx();
 
+    /**
+     * Used for RegressionTest generateAnswer when {@link UnitTest#getMyTestParameterFiles(Class, Class)} has two different
+     * classes, and we need to generate answer files in the ansClass directory
+     */
+    @Nullable
+    private static String ansClass = null;
```
drivenflywheel, line 103: This isn't thread-safe.  We can't use a static member to store state that multiple test classes need to manage independently.

sambish5: This and the `List<String> answerFiles` in UnitTest, correct?

drivenflywheel: Yep - I'll have a `UnitTest` PR up shortly.  Hopefully it'll also make RegresstionTest changes easy

sambish5:
updated based on the new thread-safe version of UnitTest (#826) that allows for different directories used for test data/answer files

src/test/java/emissary/test/core/junit5/RegressionTestUtil.java
```diff
@@ -95,6 +96,8 @@ public final class RegressionTestUtil {
      */
     private static final Path TEST_RESX = getTestResx();
 
+    private static AtomicReference<Class<?>> answerFileClassRef = null;
```
drivenflywheel, line 99: This isn't thread-safe

drivenflywheel (changes requested):
Still susceptible to race conditions across tests that extend RegressionTest

src/test/java/emissary/test/core/junit5/RegressionTestUtil.java
sambish5: I altered it so the answerFileClasRef is never stored within RegressionTestUtil. I'm not sure if we want to keep the methods and add duplicates with different signatures, or if this is the preferred option.

drivenflywheel (approved):
Working as intended when locally testing integration

src/test/java/emissary/test/core/junit5/RegressionTestUtil.java
drivenflywheel: I think this change is fine.  If we do any further cleanup, I'd probably start again in `UnitTest` and remove the `AtomicReference<?>` wrapper, and get the `Ref` suffix off the variable and parameter names.  There's probably a bit more ceremony than it actually needs.  This essentially private to RegressionTest -> RegressionTestUtil, so I'm not currently concerned if it isn't the most intuitive interface.

PR #839, opened 2024-07-03
Update startup logging

dev-mlb:
Startup was updated to use sets, so this adds logging to notify the user that a duplicate class was configured. Ideally we test for this and see if it hits in prod and then later we'll add it to strict mode failures. I also removed the logging of the full location of a place during startup and tried to make the output a little more useful by adding status and class name. Lastly, I added color output to logback to make the logs a little easier to read.


cfkoehler (approved):
Integrated into upstream and works as expected

drivenflywheel (approved):

PR #843, opened 2024-07-09
b2895-json escape bug fix

DonResnik:
To test and see the json escape work:

```mvn clean  verify -Dtest=JsonEscapePlaceTest```

To see it error out, revert ```src/main/java/emissary/transform/decode/JsonEscape.java``` with original line, rerun the test and it will treat ```\n``` as a new line and the test will fail

### DIFF
--- a/src/main/java/emissary/transform/decode/JsonEscape.java
+++ b/src/main/java/emissary/transform/decode/JsonEscape.java
@@ -7,7 +7,7 @@ import java.io.IOException;
 
 public class JsonEscape {
 
-    private final static String ESCAPES = "ntr\"'/";
+    private final static String ESCAPES = "ntr\"'/\\";
 
     /**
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/main/java/emissary/transform/decode/JsonEscape.java [hunk 1]
```diff
@@ -7,7 +7,7 @@
 
 public class JsonEscape {
 
-    private final static String ESCAPES = "ntr\"'/";
+    private final static String ESCAPES = "ntr\"'/\\";
```
sambish5, line 10: The ordering should be `private static final`

DonResnik: this has been updated.  Sorry for the late response, I just realized my email address for notifications was old, it is now current

drivenflywheel (approved):

sambish5 (approved):

PR #860, opened 2024-07-25
errorprone :: suppress HidingField

sambish5:
only occurs when testing classes use `import ch.qos.logback.classic.Logger;` against UnitTest's slf4j logger

### DIFF
--- a/src/test/java/emissary/core/channels/LoggingChannelFactoryTest.java
+++ b/src/test/java/emissary/core/channels/LoggingChannelFactoryTest.java
@@ -27,6 +27,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 class LoggingChannelFactoryTest extends UnitTest {
     private ListAppender<ILoggingEvent> appender;
+    @SuppressWarnings("HidingField")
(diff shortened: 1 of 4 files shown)


drivenflywheel:
> only occurs when testing classes use `import ch.qos.logback.classic.Logger;` against UnitTest's slf4j logger

Occurs when testing classes that extend UnitTest declare their own `logger` field that hides UnitTest's existing `protected` logger field.  Might be able to remove the masking fields and use the inherited one rather than suppressing the warning.

sambish5:
> > only occurs when testing classes use `import ch.qos.logback.classic.Logger;` against UnitTest's slf4j logger
> 
> Occurs when testing classes that extend UnitTest declare their own `logger` field that hides UnitTest's existing `protected` logger field. Might be able to remove the masking fields and use the inherited one rather than suppressing the warning.

The logger could be removed in `LogbackTesterTest` and allow it to use the `UnitTest` logger if wanted. In the other three tests, since `logger.addAppender()` & `logger.detachAppender()` are called, the loggers in those testing classes cannot be removed

drivenflywheel (approved):

cfkoehler (approved):

PR #871, opened 2024-08-02
reduce transform history output

dev-mlb:


cfkoehler (approved):
Integrated upstream and got compact history as expected when enabled.

sambish5 (approved):

PR #875, opened 2024-08-02
errorprone :: ConstantField - UnitTest.TMPDIR

jpdahlke:
I think this has far too many touch points in the test code to update right now

### DIFF
--- a/src/test/java/emissary/test/core/junit5/UnitTest.java
+++ b/src/test/java/emissary/test/core/junit5/UnitTest.java
@@ -52,6 +52,7 @@ public abstract class UnitTest {
 
     @TempDir
     public static File temporaryDirectory;
+    @SuppressWarnings("ConstantField")
     protected static String TMPDIR = "/tmp";
     @Nullable
     protected Package thisPackage = null;


drivenflywheel (approved):

src/test/java/emissary/test/core/junit5/UnitTest.java
```diff
@@ -52,6 +52,7 @@ public abstract class UnitTest {
 
     @TempDir
     public static File temporaryDirectory;
+    @SuppressWarnings("ConstantField")
     protected static String TMPDIR = "/tmp";
```
fbruton, line 56: Any reason why this field can't renamed to comply with the naming convention of a non-constant field.
```suggestion
    protected static String tmpDir = "/tmp";
```

drivenflywheel: > Any reason why this field can't renamed to comply with the naming convention of a non-constant field.

Just potential integration cost

drivenflywheel (approved):

src/test/java/emissary/test/core/junit5/UnitTest.java
jpdahlke: Correct, it just has enough touch points downstream that it spilled over the acceptable arbitrary limit.

sambish5 (approved):

PR #877, opened 2024-08-06
refactor CoordinationPlace to improve extensibility and reduce complexity

fbruton:


src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -220,49 +215,88 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
...
+
+        }
+    }
+
+    protected boolean processOutputFormOnError() {
+        return false;
+    }
+
+    protected void processOutputForm(IBaseDataObject d) {
+        // Process the ouptut form according to configuration
```
cfkoehler, line 288:
```suggestion
        // Process the output form according to configuration
```

cfkoehler (dismissed):
Just one spelling mistake that was copied over. Otherwise I did not find any functional differences which is expected/

cfkoehler (dismissed):

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -220,49 +215,124 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
...
+            }
+        }
+    }
+
+    /**
+     * If true, do not continue processing other places after an error occurs
+     *
+     * @return boolean to not continue processing after an error
+     */
+    protected boolean allowShouldNotContinueOnError() {
```
jpdahlke, line 334:
```suggestion
    protected boolean shouldNotContinueOnError() {
```

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -220,49 +215,124 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
...
-        return sproutCollection;
+    /**
+     * Return true when an error has occurred and the places is configured to not continue
+     *
+     * @param p place that is currently processing the ibdo
+     * @param errorOccurred true if an error occurred
+     *
+     * @return true if processing should not continue
+     */
+    private boolean shouldNotContinueOnError(IServiceProviderPlace p, boolean errorOccurred) {
```
jpdahlke, line 272: I think the logic here should be reversed; `shouldContinueOnErrror` will return a "false" on "continue" rather than "true" on a "not continue".

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -220,49 +215,124 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
...
+        if (allowShouldNotContinueOnError() && errorOccurred) {
```
jpdahlke, line 273: I'm not sure this has to exist as part of the override. I could see wanting to override the private method in this case to do something other than the default scenario. 

`protected boolean shouldContinue(p, errorOccurred)`

`protected boolean continueOnError()`

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -220,49 +215,124 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
...
-        cleanUpHook(d);
+    /**
+     * Allow derived classes a shot to handle a place exception
+     *
+     * @param hd true if doing heavy-duty processing
+     * @param ex exception thrown by the place
+     *
+     * @return if an error occurred
+     */
+    protected boolean handlePlaceException(boolean hd, Exception ex) {
```
drivenflywheel, line 259: If this method will have a boolean return value, its name should reflect probably that.  Something like `handledPlaceException` or `placeExceptionWasHandled`.  

Should we consider passing a reference to the place that threw the exception? (i.e., adding another parameter)

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -220,49 +215,124 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
...
+        logger.warn("agentProcess {} called from Coordinate problem", (hd ? "HeavyDuty" : "Call"), ex);
```
drivenflywheel, line 260: Should lose the extra space here that's breaking up the called method name:
```suggestion
        logger.warn("agentProcess{} called from Coordinate problem", (hd ? "HeavyDuty" : "Call"), ex);
```

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -220,49 +215,124 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
...
+    protected boolean allowProcessOutputFormOnError() {
+        return false;
+    }
+
+    /**
+     * Process the outputs form according to configuration
+     *
+     * @param d the ibdo to process
+     */
+    protected void processOutputForm(IBaseDataObject d) {
```
drivenflywheel, line 319: Maybe `applyOutputForm` describes the functionality more specifically than `processOutputForm` does?

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -220,49 +215,124 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
...
+
+        }
+    }
+
+    /**
+     * If true, process the output form the same for the default and error condition
+     *
+     * @return boolean to allow processing of output for when an error occurs
+     */
+    protected boolean allowProcessOutputFormOnError() {
```
drivenflywheel, line 310: Maybe?
```suggestion
    protected boolean shouldApplyOutputFormOnError() {
```

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -220,49 +215,124 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
...
+    }
+
+    /**
+     * Process the form
+     *
+     * @param d the ibdo to process
+     * @param errorOccurred true if an error occurred
+     */
+    protected void processForm(IBaseDataObject d, boolean errorOccurred) {
+        if (allowProcessOutputFormOnError() || !errorOccurred) {
```
drivenflywheel, line 296: Should probably swap the order of these so we can short-circuit on a boolean value without having to evaluate the method call
```suggestion
        if (!errorOccurred || allowProcessOutputFormOnError())  {
```

drivenflywheel (commented):
Minor naming suggestions, but nothing I feel strongly about.  I like the direction this is going.

cfkoehler (approved):

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -239,30 +232,108 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
...
-        return sproutCollection;
+    /**
+     * Return false when an error has occurred and the places is configured to not continue
+     *
+     * @param p place that is currently processing the ibdo
+     * @param errorOccurred true if an error occurred
+     *
+     * @return false if processing should not continue
+     */
+    protected boolean shouldContinue(IServiceProviderPlace p, boolean errorOccurred) {
```
drivenflywheel, line 274: Let's move this new overload to be adjacent to the existing `shouldContinue` method

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -239,30 +232,108 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
...
+            nukeMyProxies(d);
+        }
+    }
+
+    /**
+     * If true, process the output form the same for the default and error condition
+     *
+     * @return boolean to allow processing of output for when an error occurs
+     */
+    protected boolean shouldApplyOutputFormOnError() {
```
drivenflywheel, line 311: (optional) Could we move this just before the other `shouldXXX` methods?

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -239,30 +232,108 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
...
+    /**
+     * Allow derived classes a shot to handle a place exception
+     *
+     * @param p place that was processing when the exception was thrown
+     * @param hd true if doing heavy-duty processing
+     * @param ex exception thrown by the place
+     * 
+     * @return if an error occurred
+     */
+    protected boolean handledPlaceException(IServiceProviderPlace p, boolean hd, Exception ex) {
```
jpdahlke, line 261:
```suggestion
    protected boolean handlePlaceException(IServiceProviderPlace p, boolean hd, Exception ex) {
```

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -239,30 +232,108 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
...
+        logger.warn("agentProcess{} called from Coordinate problem", (hd ? "HeavyDuty" : "Call"), ex);
+        return true;
+    }
 
-        return sproutCollection;
+    /**
+     * Return false when an error has occurred and the places is configured to not continue
```
jpdahlke, line 267:
```suggestion
     * Return whether to continue traversing the list of coordinated places when an error occurs in one of them
```

src/main/java/emissary/place/CoordinationPlace.java
drivenflywheel: Scratch that - this is already close to where it's used so the cohesion is still pretty good

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -239,30 +232,108 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
...
+        if (updateTransformHistory) {
+            DirectoryEntry de = p.getDirectoryEntry();
+            de.setDataType(d.currentForm());
+            // append to the transform history, with flag indicating that the visit was coordinated
+            d.appendTransformHistory(de.getKey(), true);
+        }
+    }
+
+    /**
+     * Apply the form
```
jpdahlke, line 292:
```suggestion
     * How to handle applying the output form if an error occurred during processing
```

jpdahlke (changes requested):
minor comments

drivenflywheel (approved):

cfkoehler (approved):

jpdahlke (approved):

PR #883, opened 2024-08-08
Add RegressionTest error message for when possible non-deterministic behavior is detected.

jdcove2:
This PR adds an error message when RegressionTest is generating answers and differences are detected. This usually indicates that there is non-deterministic state in the IBDO because the place is executed once to generate the answer file and once to check against the answer file. If these two executions produce different results, then this most likely indicates non-deterministic state.

### DIFF
--- a/src/test/java/emissary/test/core/junit5/RegressionTestUtil.java
+++ b/src/test/java/emissary/test/core/junit5/RegressionTestUtil.java
@@ -71,7 +71,7 @@ public final class RegressionTestUtil {
     public static final String THROWABLE_MESSAGE_NAME = "throwableMessage";
 
     public static final String NON_DETERMINISTIC_ERROR_MESSAGE =
-            "\nNOTE: Since 'generateAnswers' is true, these differences could be because of non-deterministic state in the IBDO\n";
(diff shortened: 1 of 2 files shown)


cfkoehler:
Do you have an issue or ticket that initiated this work? 
I think this is something we should add just curious on the motivation for doing it now.

src/test/java/emissary/test/core/junit5/RegressionTestUtil.java
```diff
@@ -70,6 +70,9 @@ public final class RegressionTestUtil {
      */
     public static final String THROWABLE_MESSAGE_NAME = "throwableMessage";
 
+    public static final String NON_DETERMINISTIC_ERROR_MESSAGE =
+            "\nNOTE: Since 'generateAnswers' is true, these differences could be because of non-deterministic state in the IBDO\n";
```
drivenflywheel, line 74:
```suggestion
            "\nNOTE: Since 'generateAnswers' is true, these differences could indicate non-deterministic processing in the tested code path\n";
```

drivenflywheel (changes requested):
Minor verbiage change requested, but otherwise looks good

drivenflywheel (approved):

cfkoehler (approved):

PR #887, opened 2024-08-21
building out VersionPlace to add emissary version info to metadata

sambish5:
Building out a VersionPlace so that Emissary version information can be added to the IBDO metadata. Can be stored as either just the version info (i.e. 8.9.0-SNAPSHOT) or the version with the given date for more detail.


src/main/resources/emissary/place/VersionPlace.cfg
```diff
@@ -0,0 +1,8 @@
+PLACE_NAME = "VersionPlace"
+SERVICE_NAME = "VERSION"
+SERVICE_TYPE = "VERIFY"
+SERVICE_DESCRIPTION = "add emissary version to payload"
+SERVICE_COST = 50
+SERVICE_QUALITY = 50
+
+SERVICE_PROXY = "UNKNOWN"
```
dev-mlb, line 8: Do we want to call this for everything, i.e. "*"?

dev-mlb (changes requested):

src/main/java/emissary/place/VersionPlace.java
```diff
@@ -0,0 +1,73 @@
...
+        configurePlace();
+    }
+
+    private void configurePlace() {
+        includeDate = configG.findBooleanEntry("INCLUDE_DATE", true);
+    }
+
+    @Override
+    public void process(IBaseDataObject payload) throws ResourceException {
+        Version version = new Version();
```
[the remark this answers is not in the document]
sambish5: added getEmissaryVersion() to help this so the process method is just putParameter

src/main/resources/emissary/place/VersionPlace.cfg
sambish5: I didn't know if that was wanted, so for now just made it UNKNOWN

src/main/java/emissary/place/VersionPlace.java
```diff
@@ -0,0 +1,76 @@
...
+    private void configurePlace() {
+        includeDate = configG.findBooleanEntry("INCLUDE_DATE", true);
+    }
+
+    @Override
+    public void process(IBaseDataObject payload) throws ResourceException {
+        payload.putParameter(EMISSARY_VERSION, getEmissaryVersion());
+    }
+
+    private String getEmissaryVersion() {
```
drivenflywheel, line 67: Let's use this approach to populate a field in the `configurePlace` method, and then the `process` method can use that field value.

drivenflywheel (changes requested):

src/main/java/emissary/place/VersionPlace.java
sambish5: yeah brain-farted there not to just doing that :laughing:, made the change

drivenflywheel (approved):

drivenflywheel (approved):

src/main/java/emissary/place/VersionPlace.java
```diff
@@ -0,0 +1,78 @@
...
+     * Create with the default configuration
+     */
+    public VersionPlace() throws IOException {
+        super();
+        configurePlace();
+    }
+
+    private void configurePlace() {
+        includeDate = configG.findBooleanEntry("INCLUDE_DATE", true);
+        formattedVersion = getEmissaryVersion();
```
cfkoehler, line 61: Realizing that this sounds a bit crazy and we should probably fix/simplify this. 
Do we want to use the version that is added to the git properties file during build time? 
https://github.com/NationalSecurityAgency/emissary/blob/main/src/main/java/emissary/util/GitRepositoryState.java#L154

This is used at startup to print the version.

sambish5: Didn't even realize this was a thing in the project haha. I think it may be smart right now to use the Version object to get this information, but in the future if we want to add commit hash/branch info, just use the GitRepoState class. Others may have differing opinions though

cfkoehler: yep agree

cfkoehler (approved):

src/main/resources/emissary/place/VersionPlace.cfg
jpdahlke: Address in follow-on

PR #888, opened 2024-08-21
Fix LogbackTester bug which did not allow Exceptions with no (i.e. null) messages.

jdcove2:
This PR fixes LogbackTester so that it accepts Exceptions in log statements that do not have a message (i.e. the message is 'null').


cfkoehler:
Do we want to allow exception messages null message values?

drivenflywheel:
> Do we want to allow exception messages to have null message values?

Most `NullPointerException`s I see have a null message.  (Wish that wasn't the case.)

cfkoehler:
Thanks for the response.

cfkoehler (approved):

PR #896, opened 2024-08-26
add an ExtractedRecord object that cannot have extracted records

dev-mlb:


drivenflywheel:
I like this approach better than the "Option 2" approach.  Honestly, I'd love to decompose the IBDO mega-interface, having it extend a simpler interface (IBaseRecord?) that doesn't have any of the content components - no `data()`, `setData()` `channelFactory`, `newInputStream()`, `header()`, `footer()`, `size()`, `dataLength()` ...  IBDO would add those on top of the new interface, `ExtractedRecord` would implement that new interface, and (eventually?), the ExtractedRecords collection would be a collection of `IBaseRecord` or `ExtractedRecord`, not a collection of IBDO.

jpdahlke:
> I like this approach better than the "Option 2" approach. Honestly, I'd love to decompose the IBDO mega-interface, having it extend a simpler interface (IBaseRecord?) that doesn't have any of the content components - no `data()`, `setData()` `channelFactory`, `newInputStream()`, `header()`, `footer()`, `size()`, `dataLength()` ... IBDO would add those on top of the new interface, `ExtractedRecord` would implement that new interface, and (eventually?), the ExtractedRecords collection would be a collection of `IBaseRecord` or `ExtractedRecord`, not a collection of IBDO.
> 
> Basically, it would be nice if ExtractedRecord didn't have to override a lot of methods just to say it doesn't support those methods. Would be better if it just didn't have them to reference.

I like where this idea is going I think. I wanted to lay something like this out visually with some other thoughts/adjustments. 

IBaseRecord (many things would just move up to this level withholding some of the data manipulation methods described above)
 -> IBaseDataObject (allows for manipulation of data and extracted records to be added like today)
   -> TopLevelObject (FUTURE: could anchor all trees, replaces need for isTld check in some places)
   -> BaseDataObject (a child object like today)
 -> IExtractedRecord (does not allow for manipulation of data once created)
   -> ExtractedRecord

All that said, this PR seems to give us the immediate need of an enforcement mechanism (which was the ask) while setting the stage for doing what you suggest. I guess the question is, how immediate is the need and do we tackle this in parts with this being the first step or do we start a larger process now? I would imagine that the actual existence of an ExtractedRecord class will start buying us some benefit without too many downstream changes or the need for interface changes in the immediate. As long as we have a compatible loose plan for getting there, I don't think this implementation boxes us out.

drivenflywheel:
> As long as we have a compatible loose plan for getting there, I don't think this implementation boxes us out.

Agreed. I think we could move forward with this approach.

dev-mlb:
@jpdahlke @drivenflywheel This is a much bigger change than I originally intended, but I added the IBaseRecord and IExtractedRecord interfaces. I created an abstract BaseRecord object that had most of the BaseDataObject code in it. For this ticket, I only kept the extracted records in the IBaseDataObject. I figured we could break this down into smaller chunks (data, channels, etc.) later on and tackle them one by one.

jpdahlke:
I think [commit](https://github.com/NationalSecurityAgency/emissary/pull/896/commits/97dfc4bfccbeb27821d27eb7c1d7294285c81018) illustrates how easy it would be to change the interfaces around once we have an ExtractedRecord class (albeit a breaking change downstream which we wouldn't take yet). With only the first commit, we'd be subject to unsupported method exceptions, but I'm guessing the pros outweigh the cons? What do you think @drivenflywheel?

dev-mlb:
Updated here #922

PR #898, opened 2024-08-27
VersionPlace - add EMISSARY_VERSION_HASH to metadata

sambish5:
follow up  to https://github.com/NationalSecurityAgency/emissary/pull/887
- updated `VersionPlace.cfg` to have `SERVICE_PROXY = "*"`
- added full git commit id to variables in `GitRepositoryState`
- Built out functionality to add either the full or abbreviated commit hash to the payload under the key `EMISSARY_VERSION_HASH`. Will default to the full hash, but the functionality is there to add `USE_ABBREV_HASH` to the config if that's the desired version stored.


src/main/java/emissary/place/VersionPlace.java
```diff
@@ -58,12 +62,15 @@ public VersionPlace() throws IOException {
 
     private void configurePlace() {
         includeDate = configG.findBooleanEntry("INCLUDE_DATE", true);
+        useAbbrevHash = configG.findBooleanEntry("USE_ABBREV_HASH", false);
```
cfkoehler, line 65: Open to what others think but I would prefer this default to `true`.

jpdahlke: I think we use the full hash now? Do you propose changing to just use abbreviated instead in the RELEASE file? I'm good with either, I don't think it would break anything downstream.

drivenflywheel: For a metadata field that will already have the version string, I'd vote for the abbreviation since it really just differentiates between SNAPSHOTs.  No need for the full 40-char string

cfkoehler: I would need to double check but I do not think we save the hash as a metadata field at all currently.  I had not considered how this interacts with the RELEASE file. 

But like @drivenflywheel said the full hash just seemed more then necessary. But this is configurable so it really does not matter.

cfkoehler (approved):

jpdahlke (approved):

PR #900, opened 2024-08-28
update command validation for mode value and use an enum for possible values 

jpdahlke:
This quickly unraveled from a simple errorprone update for an UnusedMethod `setMode` into addressing a few touch points for this command option that weren't straightforward. This effort attempts to address a few things:
- stop setting and then using an environment variable for mode
- fully annotate the possible modes
- allow mixed case values for mode at the command line
- continue to validate the values provided by the user using picocli methods
- clean up some setup exception handling


dev-mlb:
Cluster mode seems to be broken...server and feeder start up, but register peer doesn't seem to fire...

jpdahlke:
Handled via #1024 and #1042

PR #901, opened 2024-08-29
update to VersionPlaceTest to allow it to run standalone

jpdahlke:
Looking for feedback on if this is a better strategy for the unit test portion since it's runnable in an IDE this way rather than relying on the maven plugin to generate a properties file at build time.

This also incorporated the change to use abbreviated hashing by default.


cfkoehler (approved):

sambish5 (approved):
LGTM

PR #907, opened 2024-09-01
remove EmissaryIsolatedClassLoaderExtension class

jpdahlke:
I'm not sure that this is used anywhere else, but now that junit4 is gone I'm not sure that it's required to exist to work around whatever was the original issue?


jpdahlke:
This must have been to solve a transient issue since I did not see failures locally at first but I do now.

dev-mlb (approved):

PR #911, opened 2024-09-04
extracted records cannot have extracted records - option 3 - Split IBDO

jdcove2:
This PR satisfies the first step in the below process of updating Emissary to enforce Extracted Records not being able to have Extracted Records while BaseDataObjects can. These code changes have been tested with the "downstream" projects and they all build and run without modification.

1. Split the IBDO into an IExtractedRecord interface (everything in the current IBDO except extracted record methods) that is extended by the IBaseDataObject interface (the extracted record methods).
2. Update "downstream" projects to correctly use IExtractedRecord for Extracted Records.
3. Update the extracted record methods to use IExtractedRecord instead of IBaseDataObject, which will now enforce the goal of extracted records not being able to contain extracted methods.


jpdahlke:
This is an interesting way of flipping the interface structure to solve the literal enforcement ask, but after talking with @drivenflywheel and @dev-mlb, this confuses the concepts of ExtractedRecord a bit with the rest of the current code base and doesn't move us in a good direction for future enhancements to these interfaces.

PR #919, opened 2024-09-09
Adding support to link a tld to a child IBDO for inheriting parameters

rpg36:
An effort to link children to the top level document to inherit their metadata instead of copying it down.

### DIFF
--- a/src/main/java/emissary/core/BaseDataObject.java
+++ b/src/main/java/emissary/core/BaseDataObject.java
@@ -194,9 +194,7 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
 
     final SafeUsageChecker safeUsageChecker = new SafeUsageChecker();
 
-    protected final IBaseDataObject parent;
-
-    protected final Set<String> inheritedKeys;
+    protected final IBaseDataObject tld;
 
     protected enum DataState {
         NO_DATA, CHANNEL_ONLY, BYTE_ARRAY_ONLY, BYTE_ARRAY_AND_CHANNEL
@@ -255,8 +253,7 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
     public BaseDataObject() {
         this.theData = null;
         setCreationTimestamp(Instant.now());
-        parent = null;
-        inheritedKeys = null;
+        tld = null;
     }
 
     /**
@@ -270,8 +267,7 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
         setData(newData);
         setFilename(name);
         setCreationTimestamp(Instant.now());
-        parent = null;
-        inheritedKeys = null;
+        tld = null;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/emissary/core/BaseDataObject.java
```diff
@@ -194,6 +194,10 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
 
     final SafeUsageChecker safeUsageChecker = new SafeUsageChecker();
 
+    protected final IBaseDataObject parent;
+
+    protected final Set<String> inheritedKeys;
```
fbruton, line 199: Recommend removing this field from the BaseDataObject` and passing as a parameter to `IBaseDataObjectHelper#getParametersWithInheritance()`

src/main/java/emissary/core/BaseDataObject.java
```diff
@@ -194,6 +194,10 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
 
     final SafeUsageChecker safeUsageChecker = new SafeUsageChecker();
 
+    protected final IBaseDataObject parent;
```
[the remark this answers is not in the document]
fbruton: I agree with @drivenflywheel. Parent and TLD have different meanings when the BaseDataObject is more than one level down in the hierarchy.

fbruton (changes requested):

src/main/java/emissary/core/IBaseDataObjectHelper.java
```diff
@@ -260,4 +262,25 @@ public static byte[] findPreferredData(final IBaseDataObject payload, List<Strin
...
         return payload.data();
     }
+
+    /**
+     * Get the effective parameters for a given payload, considering parameters inherited from the parents
+     * 
+     * @param payload Get parameters from this IBaseDataObject
+     * @return The full set parameters including inherited ones
+     */
+    public static Map<String, Collection<Object>> getParametersWithInheritance(IBaseDataObject payload) {
```
fbruton, line 272:
```suggestion
    public static Map<String, Collection<Object>> getCombinedTldParameters(IBaseDataObject payload) {
```

src/main/java/emissary/core/IBaseDataObjectHelper.java
```diff
@@ -260,4 +262,25 @@ public static byte[] findPreferredData(final IBaseDataObject payload, List<Strin
...
+        IBaseDataObject parent = payload.getParent();
+        Set<String> inheritedParameters = payload.getInheritedParameters();
+        Map<String, Collection<Object>> effectiveParams = new HashMap<>(payload.getParameters());
+        if (parent != null) {
+            for (String key : parent.getParameterKeys()) {
+                if (inheritedParameters == null || inheritedParameters.contains(key)) {
```
fbruton, line 278: Only add the TLD parameter if current `BaseDataObject` parameter's doesn't contain the key.

fbruton: See LogFilter.java

src/main/java/emissary/core/BaseDataObject.java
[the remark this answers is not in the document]
rpg36: Naming conventions updated.

fbruton (changes requested):

src/main/java/emissary/core/BaseDataObject.java
```diff
@@ -288,6 +292,28 @@ public BaseDataObject(final byte[] newData, final String name, final String form
         }
     }
 
+    public BaseDataObject(final byte[] newData, final String name, @Nullable final String form, IBaseDataObject tld) {
+        setData(newData);
```
dev-mlb, line 296: Any reason not to use `this(newData, name, form);`? Do you not want the creation timestamp set??

PR #923, opened 2024-09-12
update 'dep.' names and use 'plugin.' prefix for plugins in pom

jpdahlke:
- drop numeric portion in name for deps
- use "plugin." for plugin versions instead of "dep."
- annotate all plugin versions used
- sync jxr plugin on 3.3.0 (it was called out twice with different versions)

If this went too far with annotating all the versions I can scale it back, but having an easy way to find all plugin versions was nice.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,30 +57,26 @@
     <dep.commons-exec.version>1.3</dep.commons-exec.version>
     <dep.commons-io.version>2.13.0</dep.commons-io.version>
     <dep.commons-lang.version>3.13.0</dep.commons-lang.version>
-    <dep.commons-pool2.version>2.11.1</dep.commons-pool2.version>
+    <dep.commons-pool.version>2.11.1</dep.commons-pool.version>
     <dep.dropwizard.metrics.version>4.2.25</dep.dropwizard.metrics.version>
     <dep.errorprone.version>2.23.0</dep.errorprone.version>
     <dep.gson.version>2.10.1</dep.gson.version>
     <dep.guava.version>32.1.2-jre</dep.guava.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -91,6 +87,40 @@
...
+    <plugin.maven-jxr.version>3.3.0</plugin.maven-jxr.version>
+    <plugin.maven-pmd.version>3.20.0</plugin.maven-pmd.version>
+    <plugin.maven-project-info-reports.version>3.4.2</plugin.maven-project-info-reports.version>
+    <plugin.maven-release.version>3.0.0</plugin.maven-release.version>
+    <plugin.maven-replacer.version>1.5.3</plugin.maven-replacer.version>
+    <plugin.maven-resources.version>3.3.0</plugin.maven-resources.version>
+    <plugin.maven-scm-publish.version>3.2.1</plugin.maven-scm-publish.version>
+    <plugin.maven-site.version>3.12.1</plugin.maven-site.version>
+    <plugin.maven-source.version>3.2.1</plugin.maven-source.version>
+    <!-- seems to have an impact with java 11 integration, update with caution -->
```
cfkoehler, line 118: Do we think this comment is still relevant?
Maybe addressed in a future PR if we upgrade to 3.5.0 or 3.1.2

cfkoehler (approved):
I think the scope is ok. Able to follow the changes

sambish5 (approved):

pom.xml
jpdahlke: Yeah, there are a few things that need addressing like that.

fbruton (approved):

PR #925, opened 2024-09-12
Made AbstractFilter Denylist settings configurable

rupert-griffin:
Can now configure the allowed characters in a denylist entry. 
Use Regex format. e.g. `ALLOWED_NAME_CHARS = "a-zA-Z0-9_"`

Allowed characters will get substituted into the patterns for view name formatting and character ordering. These regexes can also be configured differently, either using `%s` to substitute the allowed character string, or manually typing out the allowed chars. e.g.:
`VIEWNAME_FORMAT = "^[%s]+(\\.[%s]+)?\\*?$"`
`VIEWNAME_FORMAT = "^[a-zA-Z0-9_]+(\\.[a-zA-Z0-9_]+)?\\*?$"`

### DIFF
--- a/src/main/java/emissary/output/filter/AbstractFilter.java
+++ b/src/main/java/emissary/output/filter/AbstractFilter.java
@@ -88,14 +88,11 @@ public abstract class AbstractFilter implements IDropOffFilter {
     @Nullable
     protected DropOffUtil dropOffUtil = null;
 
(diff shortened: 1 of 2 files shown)


fbruton (changes requested):
Please add a unit-test that exercises the regular expressions.

fbruton (dismissed):

cfkoehler (approved):

fbruton (approved):

PR #933, opened 2024-09-17
Shortening the logged message if an exception occurs during Executrix execution

drivenflywheel:


cfkoehler (approved):
Note to self to check logstash groks for these messages.

PR #952, opened 2024-09-25
updates to VersionPlace for less code-reusage downstream

sambish5:
Basically gets rid of the use of `emissary.util.Version`, and relies on `GitRepositoryState` to get the version and hash. This allows for a lot less code-reusage for classes downstream that extend this.
Created setters for the booleans so they can be set downstream.

One thing to keep in mind for this is the BuildTime is only reset in the `emissary.git.properties` file when "mvn clean" is run

### DIFF
--- a/src/main/java/emissary/place/VersionPlace.java
+++ b/src/main/java/emissary/place/VersionPlace.java
@@ -3,7 +3,6 @@ package emissary.place;
 import emissary.core.IBaseDataObject;
 import emissary.core.ResourceException;
 import emissary.util.GitRepositoryState;
-import emissary.util.Version;
 
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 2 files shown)


src/main/java/emissary/place/VersionPlace.java
```diff
@@ -100,4 +100,12 @@ private String getEmissaryVersionHash() {
             return gitRepositoryState.getCommitId();
         }
     }
+
+    public void setIncludeDate(Boolean includeDate) {
```
dev-mlb, line 104: add getters

dev-mlb (changes requested):

src/main/java/emissary/place/VersionPlace.java
sambish5: added them in

dev-mlb (approved):

jpdahlke (approved):

PR #981, opened 2024-10-17
errorprone :: MixedMutabilityReturnType

dev-mlb:


drivenflywheel (approved):

src/main/java/emissary/directory/DirectoryPlace.java
```diff
@@ -973,15 +973,15 @@ protected List<DirectoryEntry> nextKeys(final String dataId, final IBaseDataObje
         // Nothing for the dataId or any wildcarded versions, we are done
         if ((currentList == null) || currentList.isEmpty()) {
             logger.debug("nextKey - nothing found here for {}", dataId);
-            return Collections.emptyList();
+            return List.of();
```
drivenflywheel, line 976: This approach doesn't seem to actually resolve the warning.  This is still returning an immutable collection, but the collection returned on line 1036 isn't immutable.

jpdahlke (approved):

PR #998, opened 2024-10-23
errorprone :: CatchingUnchecked

dev-mlb:


sambish5 (dismissed):

src/main/java/emissary/place/MultiFileUnixCommandPlace.java
```diff
@@ -605,6 +605,7 @@ protected List<IBaseDataObject> processData(IBaseDataObject tData) throws Resour
      * @param len length of data to use
      * @return attachments
      */
+    @SuppressWarnings("CatchingUnchecked")
```
DonResnik, line 608: why does this method still catch Exception and not RuntimeException?

src/main/java/emissary/pool/PayloadLauncher.java
```diff
@@ -21,6 +21,7 @@ public class PayloadLauncher {
     private static final Logger logger = LoggerFactory.getLogger(PayloadLauncher.class);
 
 
+    @SuppressWarnings("CatchingUnchecked")
```
DonResnik, line 24: why does this method catch Exception and not RuntimeException?

dev-mlb: `pool.borrowAgent()` throws Exception

sambish5 (approved):
changes lgtm

PR #1001, opened 2024-10-24
improve ByteUtil printable char validation of multi-byte charset

fbruton:
~~Planning to add additional unit-tests.~~

### DIFF
--- a/src/main/java/emissary/util/ByteUtil.java
+++ b/src/main/java/emissary/util/ByteUtil.java
@@ -383,7 +383,8 @@ public class ByteUtil {
      */
     private static boolean isPrintable(int codePoint) {
 
-        return codePoint >= 0x20 && codePoint <= 0x7E || // basic latin
+        return codePoint >= 0x09 && codePoint <= 0x0D ||
+                codePoint >= 0x20 && codePoint <= 0x7E || // basic latin
                 codePoint >= 0xA0 && codePoint <= 0xD7FF || // extended characters
                 codePoint >= 0xE000 && codePoint <= 0xFFFD ||
                 codePoint >= 0x10000 && codePoint <= 0x10FFFF;
--- a/src/main/java/emissary/util/ByteUtil.java
+++ b/src/main/java/emissary/util/ByteUtil.java
@@ -291,7 +291,7 @@ public class ByteUtil {
      * Scans a byte array looking for non-printable values.
      * 
      * @param utf8Bytes the bytes to be scanned.
-     * @return regardless of whether there were non-printable values.
+     * @return true if the byte array contains an invalid text character.
      */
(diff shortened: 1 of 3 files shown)


src/test/java/emissary/util/ByteUtilTest.java
```diff
@@ -208,4 +209,26 @@ void testGrabLine() {
         assertEquals("This is line three", ByteUtil.grabLine(data, 35), "Last line extraction");
     }
 
+    @Test
+    void testHasnonPrintableValues() {
+        String newLineCarriageTab = "\tThis is line one\r\nThis is line two\nThis is line three\n\nEnding with a tab";
+        assertFalse(ByteUtil.hasNonPrintableValues(newLineCarriageTab.getBytes(StandardCharsets.UTF_8)));
```
ldhardy, line 215: It's possible there is more than one definition, but I thought non-printing characters were tab, newline, etc.  Maybe I need more concise definition of what we're trying to identify.

ldhardy: Also, is it supposed to end with a tab?  \t?

fbruton: Maybe it should be called binary.

fbruton: The previous code allowed tabs, new lines and carriage returns, which makes sense.

src/main/java/emissary/util/ByteUtil.java
```diff
@@ -290,20 +290,104 @@ public static String grabLine(byte[] data, int pos) {
...
      * Scans a byte array looking for non-printable values.
      * 
-     * @param bytes the bytes to be scanned.
-     * @return whether or not there were non-printable values.
+     * @param utf8Bytes the bytes to be scanned.
+     * @return true if the byte array contains an invalid text character.
      */
-    public static boolean hasNonPrintableValues(final byte[] bytes) {
-        boolean badCharacters = false;
+    public static boolean hasNonPrintableValues(final byte[] utf8Bytes) {
```
ldhardy, line 296: Per our discussion in the office, Dave and I would like a chance to re-work this entire method and move it elsewhere.  We believe it should live in the test jar, maybe as a static member of RegressionTest or some other test utility.  I confirmed it's not used upstream in any production code.  But other members of this class are used in production code, so we should leave the class and it's other methods here.

ldhardy (approved):

dev-mlb:
I'm thinking we should merge the ideas from this pr and #1006

cfkoehler (approved):
My experience in this area is pretty minimal. But the code makes sense to me and seems to do what is needed.

PR #1004, opened 2024-10-30
VersionPlace: if version is release, do not add date

sambish5:
makes it so if version is release (i.e. 8.15.0, and not 8.15.0-SNAPSHOT), make it so even if `includeDate` is true, the version added to the payload is just the version numbering.


cfkoehler (approved):
Works and makes sense to me. Nice work

src/main/java/emissary/place/VersionPlace.java
```diff
@@ -63,6 +66,10 @@ private void configurePlace() {
 
         includeDate = configG.findBooleanEntry("INCLUDE_DATE", true);
         useAbbrevHash = configG.findBooleanEntry("USE_ABBREV_HASH", true);
+
+        String cfgRegex = configG.findStringEntry("VERSION_REGEX_FOR_NO_DATE", "^(\\d+\\.)?(\\d+\\.)?(\\d+)$");
```
DonResnik, line 70: Did you want to update VersionPlace.cfg with VERSION_REGEX_FOR_NO_DATE?

DonResnik: I see now that INCLUDE_DATE and USE_ABBREV_HASH are not in the cfg file either.  Is there a pattern for when config values are in the cfg and when they are just in the place?

jpdahlke: I think in general if the default it sufficient we leave it out of the .cfg file. Anyone is free to write their own .cfg file and use a different value which is what we like to allow for.

DonResnik: @jpdahlke - sounds, good, thx for the explanation

DonResnik (approved):

PR #1006, opened 2024-11-01
Update RegressionTest to assume UTF-8.

jdcove2:
This PR updates RegressionTest to assume all byte arrays are UTF-8 and not process them as ASCII (i.e. ISO_8859_1). The main benefit of this is that valid UTF-8 is output in the XML directly instead of being hashed or BASE64'd.

NOTE: This is a potentially "breaking" change that might require XML files to be regenerated.

### DIFF
--- a/src/main/java/emissary/core/IBaseDataObjectXmlCodecs.java
+++ b/src/main/java/emissary/core/IBaseDataObjectXmlCodecs.java
@@ -19,6 +19,7 @@ import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.io.PushbackReader;
 import java.io.Reader;
 import java.lang.reflect.Method;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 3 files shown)


src/main/java/emissary/core/IBaseDataObjectXmlCodecs.java
```diff
@@ -826,4 +831,55 @@ public static Method getIbdoMethod(final String methodName, final Class<?>... pa
             throws NoSuchMethodException {
         return IBaseDataObject.class.getMethod(methodName, parameterTypes);
     }
+
+    public static boolean requiresEncoding(final byte[] utf8Bytes) {
```
dev-mlb, line 835: Any reason not to update the ByteUtil or add these methods to a util class?? Also, since these are public methods, I'd like to see some javadocs.

dev-mlb (changes requested):

src/main/java/emissary/core/IBaseDataObjectXmlCodecs.java
fbruton, line 835: Recommend removing `requiresEncoding` method and updating any calls to `requiresEncoding` to call ByteUtil#containsNonIndexableBytes instead.

fbruton (changes requested):

dev-mlb (approved):

ldhardy (approved):

PR #1009, opened 2024-11-08
Add a BDO copy method to IBaseDataObjectHelper.

jdcove2:
This PR adds a method to IBaseDataObjectHelper that will shallow copy the attributes of a BaseDataObject to another BaseDataObject.


src/main/java/emissary/core/IBaseDataObjectHelper.java
```diff
@@ -97,6 +98,115 @@ public static IBaseDataObject clone(final IBaseDataObject iBaseDataObject, final
...
+     * call would not be necessary.
+     * 
+     * @param bdo the BaseDataObject to set the field on.
+     * @param fieldName the name of the field to be set.
+     * @param object the object that the field is to be set to.
+     * @throws IllegalAccessException if this {@code Field} object is enforcing Java language access control and the
+     *         underlying field is either inaccessible or final.
+     * @throws NoSuchFieldException if a field with the specified name is not found.
+     */
+    public static void setPrivateFieldValue(final BaseDataObject bdo, final String fieldName, @Nullable final Object object)
```
drivenflywheel, line 199: Why would this need to be `public`?

src/main/java/emissary/core/BaseDataObject.java
```diff
@@ -754,10 +757,14 @@ public String printMeta() {
 
     @Override
     public void addProcessingError(final String err) {
-        if (this.procError == null) {
-            this.procError = new StringBuilder();
+        if (err == null) {
+            this.procError = null;
```
drivenflywheel, line 761: Why would an "add" method remove existing entries?

drivenflywheel: > What suggestion do you have for resolving your example above?

Revert the `addProcessingError` change.  If your `copy` use case needs a `clearProcessingErrors` method, create it. Don't add unexpected new behavior to the existing method just to satisfy the new edge case.

jpdahlke:
closing for now, this can be reopened if this becomes a need in the future

PR #1014, opened 2024-11-15
 ibdo :: get parameter as string

dev-mlb:
Starting this old thread again

### DIFF
--- a/src/main/java/emissary/core/IBaseDataObject.java
+++ b/src/main/java/emissary/core/IBaseDataObject.java
@@ -3,15 +3,19 @@ package emissary.core;
 import emissary.core.channels.SeekableByteChannelFactory;
 import emissary.directory.DirectoryEntry;
 
+import org.apache.commons.lang3.StringUtils;
+
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
 import java.time.Instant;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
 public interface IBaseDataObject {
 
@@ -353,7 +357,9 @@ public interface IBaseDataObject {
      * @deprecated use {@link #getParameterAsConcatString(String)}
      */
     @Deprecated
-    String getStringParameter(String key);
+    default String getStringParameter(final String key) {
+        return getParameterAsConcatString(key);
+    }
 
     /**
      * Retrieve a specified metadata element as a string value
@@ -364,7 +370,9 @@ public interface IBaseDataObject {
      * @deprecated use {@link #getParameterAsConcatString(String, String)}
      */
     @Deprecated
-    String getStringParameter(String key, String sep);
+    default String getStringParameter(final String key, final String sep) {
+        return getParameterAsConcatString(key, sep);
+    }
 
     /**
      * Retrieve the Collection of metadata values identified by key where each element is converted to a string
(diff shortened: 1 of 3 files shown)


src/main/java/emissary/core/IBaseDataObject.java
```diff
@@ -359,9 +361,46 @@ enum MergePolicy {
...
+    String getParameterAsString(String key);
+
+
+    /**
+     * Retrieve a specified metadata element as a string of concatenated values
+     *
+     * @param key name of the metadata element
+     * @return the string value or null if no such element
+     */
+    String getParameterAsConcatString(String key);
```
drivenflywheel, line 393: Maybe?
```suggestion
    default String getParameterAsConcatString(String key) {
        return getParameterAsConcatString(key, DEFAULT_PARAM_SEPARATOR);
    }
```

dev-mlb: Yeah!

src/main/java/emissary/core/IBaseDataObject.java
```diff
@@ -350,17 +354,81 @@ enum MergePolicy {
...
+     * Retrieve the metadata value identified by key where the element is converted to a string
+     *
+     * @param key name of the metadata element
+     * @return parameter converted to strings
+     * @throws IllegalStateException if the parameter has more than one value
+     */
+    default String getParameterAsString(final String key) {
+        final var obj = getParameterAsStrings(key);
+        if (obj.size() > 1) {
+            throw new IllegalStateException("There is more than one value for parameter[" + key + " ]!!");
```
drivenflywheel, line 404: An `Exception` is probably too heavy-handed for an initial implementation, because it could cause failures that escape detection until they've been deployed.   For now, we should probably just log an error or warning so we can catalog "offending" fields.  It'll be burdensome downstream to wrap every `getParameterAsString` call with a `try...catch {log..}` block

fbruton:
I'm on board with the current changes. I recommend marking ready for review after adding unit-tests.

src/test/java/emissary/core/BaseDataObjectTest.java
```diff
@@ -1463,4 +1463,39 @@ void testNewInputStream() throws IOException {
             assertArrayEquals(bytes2, byteArrayOutputStream.toByteArray());
         }
     }
+
+    @Test
+    void testGetParameterAsString() {
+        this.b.putParameter("A", 1L);
+        this.b.appendParameter("A", "TWO");
+        this.b.appendParameter("A", "THREE");
+        assertEquals("1", this.b.getParameterAsString("A"));
```
fbruton, line 1472: Is it worth checking for the log message?
```suggestion
        try (LogbackTester logbackTester = new LogbackTester(BaseDataObject.clas
s.getName())){
            assertEquals("1", this.b.getParameterAsString("A"));
            LogbackTester.SimplifiedLogEvent logEvent = new LogbackTester.Simplifi
edLogEvent(Level.WARN, "Multiple values for parameter, returning the first - par
ameter:A, number of values:3",null);
            logbackTester.checkLogList(Collections.singletonList(logEvent));
        }
```

src/test/java/emissary/core/BaseDataObjectTest.java
```diff
@@ -1463,4 +1463,39 @@ void testNewInputStream() throws IOException {
             assertArrayEquals(bytes2, byteArrayOutputStream.toByteArray());
         }
     }
+
+    @Test
+    void testGetParameterAsString() {
+        this.b.putParameter("A", 1L);
```
fbruton, line 1469:
```suggestion
        assertNotNull(b);
        this.b.putParameter("A", 1L);
```

fbruton (changes requested):

fbruton (approved):

drivenflywheel (approved):

PR #1018, opened 2024-12-02
Deprecate IBaseDataObjectHelper.clone() that allows partial clones.

jdcove2:
This PR deprecates the IBaseDataObjectHelper.clone() method that allows for partial clones. Partial clones were implemented to allow cloning that matched the now deprecated BaseDataObject.clone() method. A search of the code bases did not show any instance where IBaseDataObjectHelper.clone() was called for a partial clone except for in the IBaseDataObjectHelperTest class. Therefore, only full clones seem to be needed and partial clones should be removed.

### DIFF
--- a/src/main/java/emissary/core/IBaseDataObjectHelper.java
+++ b/src/main/java/emissary/core/IBaseDataObjectHelper.java
@@ -52,12 +52,12 @@ public final class IBaseDataObjectHelper {
      * 
      * A "fullClone" duplicates all attributes.
      * 
-     * @deprecated
+     * @deprecated prefer {@link #clone(IBaseDataObject)}
      * @param iBaseDataObject the IBaseDataObject to be cloned.
      * @param fullClone specifies if all fields should be cloned.
      * @return the clone of the IBaseDataObject passed in.
      */
-    @Deprecated
+    @Deprecated(forRemoval = true)
     public static IBaseDataObject clone(final IBaseDataObject iBaseDataObject, final boolean fullClone) {
         Validate.notNull(iBaseDataObject, "Required: iBaseDataObject not null");
 
--- a/src/main/java/emissary/core/IBaseDataObjectHelper.java
+++ b/src/main/java/emissary/core/IBaseDataObjectHelper.java
@@ -38,14 +38,26 @@ public final class IBaseDataObjectHelper {
     private IBaseDataObjectHelper() {}
 
     /**
+     * Clones an IBaseDataObject.
+     * 
+     * @param iBaseDataObject the IBaseDataObject to be cloned.
+     * @return the clone of the IBaseDataObject passed in.
+     */
+    public static IBaseDataObject clone(final IBaseDataObject iBaseDataObject) {
+        return clone(iBaseDataObject, true);
+    }
+
+    /**
      * Clones an IBaseDataObject equivalently to emissary.core.BaseDataObject.clone(), which duplicates some attributes.
      * 
      * A "fullClone" duplicates all attributes.
      * 
+     * @deprecated
      * @param iBaseDataObject the IBaseDataObject to be cloned.
      * @param fullClone specifies if all fields should be cloned.
      * @return the clone of the IBaseDataObject passed in.
      */
+    @Deprecated
     public static IBaseDataObject clone(final IBaseDataObject iBaseDataObject, final boolean fullClone) {
         Validate.notNull(iBaseDataObject, "Required: iBaseDataObject not null");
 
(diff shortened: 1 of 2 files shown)


sambish5 (dismissed):

src/main/java/emissary/core/IBaseDataObjectHelper.java
```diff
@@ -37,15 +37,27 @@ private InternalIdBaseDataObject(final UUID internalId) {
...
     private IBaseDataObjectHelper() {}
 
+    /**
+     * Clones an IBaseDataObject.
+     * 
+     * @param iBaseDataObject the IBaseDataObject to be cloned.
+     * @return the clone of the IBaseDataObject passed in.
+     */
+    public static IBaseDataObject clone(final IBaseDataObject iBaseDataObject) {
+        return clone(iBaseDataObject, true);
```
drivenflywheel, line 47: I'm not a fan having the new overload reference the very method we're deprecating.  Maybe we should refactor the existing one by extracting two supporting methods that both overloads can use.  Something like the following signatures:
```java
private static copyBasicFields(final IBaseDataObject src, final IBaseDataObject dest) {
    // body would be current lines 66-87 
}
private static copyNonBasicFields(final IBaseDataObject src, final IBaseDataObject dest) {
    // body would be current lines 90-106
}
```
With that change, this overload would strip down to the following:
```java
 public static IBaseDataObject clone(final IBaseDataObject src) {
    Validate.notNull(src, "Required: src not null");
    IBaseDataObject dst = new InternalIdBaseDataObject(src.getInternalId());
    copyBasicFields(src, dst);
    copyNonBasicFields(src, dst);
    retrun dst;
}
```
The original now-deprecated overload would become this:
```java
 public static IBaseDataObject clone(final IBaseDataObject src, final boolean fullClone) {
    Validate.notNull(src, "Required: src not null");
    IBaseDataObject dst = fullClone ? new InternalIdBaseDataObject(src.getInternalId()) : DataObjectFactory.getInstance();
    copyBasicFields(src, dst);
    if (fullClone) {
        copyNonBasicFields(src, dst);
    }
    retrun dst;
}
```

After the deprecated overload is removed, we can collapse those private methods into the remaining `clone` method

src/main/java/emissary/core/IBaseDataObjectHelper.java
```diff
@@ -37,15 +37,27 @@ private InternalIdBaseDataObject(final UUID internalId) {
...
+    }
+
     /**
      * Clones an IBaseDataObject equivalently to emissary.core.BaseDataObject.clone(), which duplicates some attributes.
      * 
      * A "fullClone" duplicates all attributes.
      * 
+     * @deprecated
```
drivenflywheel, line 55:
```suggestion
     * @deprecated prefer {@link #clone(IBaseDataObject)}
```

src/main/java/emissary/core/IBaseDataObjectHelper.java
```diff
@@ -37,15 +37,27 @@ private InternalIdBaseDataObject(final UUID internalId) {
...
      * @param iBaseDataObject the IBaseDataObject to be cloned.
      * @param fullClone specifies if all fields should be cloned.
      * @return the clone of the IBaseDataObject passed in.
      */
+    @Deprecated
```
drivenflywheel, line 60:
```suggestion
    @Deprecated(forRemoval = true)
```

drivenflywheel (changes requested):

drivenflywheel (dismissed):

jpdahlke (approved):

src/main/java/emissary/core/IBaseDataObjectHelper.java
```diff
@@ -37,15 +37,27 @@ private InternalIdBaseDataObject(final UUID internalId) {
...
      * Clones an IBaseDataObject equivalently to emissary.core.BaseDataObject.clone(), which duplicates some attributes.
      * 
      * A "fullClone" duplicates all attributes.
      * 
+     * @deprecated prefer {@link #clone(IBaseDataObject)}
      * @param iBaseDataObject the IBaseDataObject to be cloned.
      * @param fullClone specifies if all fields should be cloned.
      * @return the clone of the IBaseDataObject passed in.
      */
+    @Deprecated(forRemoval = true)
```
jpdahlke, line 60:
```suggestion
    @Deprecated
```
This makes it so the build doesn't warn on itself for this change.

jpdahlke (changes requested):

drivenflywheel (approved):

sambish5 (approved):

jpdahlke (approved):

PR #1024, opened 2024-12-16
initialize static instance of EmissaryServer

jpdahlke:
- removes the need for node.mode system property (get from command or default value)
- no longer need to check namespace binding to determine state
- deprecate stop/status methods for future cleanup

Pulled out of https://github.com/NationalSecurityAgency/emissary/pull/900 but stops short of changing mode to enum and changing command validator.

### DIFF
--- a/src/main/java/emissary/pickup/WorkSpace.java
+++ b/src/main/java/emissary/pickup/WorkSpace.java
@@ -195,13 +195,13 @@ public class WorkSpace implements Runnable {
             this.outbound = new PriorityQueue<>(11, this.feedCommand.getSort());
         }
 
-        configure();
         startJetty();
+        register();
         initializeService();
     }
 
     protected void startJetty() {
-        if (!EmissaryServer.isStarted()) {
+        if (!EmissaryServer.isInitialized() || !EmissaryServer.getInstance().isServerRunning()) {
             // TODO investigate passing the feedCommand object directly to the serverCommand
             List<String> args = new ArrayList<>();
             args.add("-b");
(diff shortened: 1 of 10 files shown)


src/main/java/emissary/server/EmissaryServer.java
```diff
@@ -92,24 +92,39 @@ public class EmissaryServer {
 
     private final EmissaryNode emissaryNode;
 
-    public EmissaryServer(ServerCommand cmd) throws EmissaryException {
-        this(cmd, new EmissaryNode());
+    private static EmissaryServer emissaryServer;
```
dev-mlb, line 95: @SuppressWarnings("NonFinalStaticField")

src/main/java/emissary/pickup/WorkSpace.java
```diff
@@ -295,13 +295,10 @@ public void stop() {
...
-            final EmissaryNode node = new EmissaryNode();
+            final EmissaryNode node = EmissaryServer.getInstance().getNode();
             if (node.isValid()) {
                 try {
-                    final EmissaryServer s = EmissaryServer.lookup();
-                    s.getServer().stop();
-                } catch (NamespaceException ex) {
-                    logger.error("Cannot find jetty server", ex);
+                    EmissaryServer.getInstance().stop();
                 } catch (Exception ex) {
```
dev-mlb, line 302: Catch Runtime or add @SuppressWarnings("CatchingUnchecked")

dev-mlb (changes requested):
nits, but otherwise looks good

dev-mlb (approved):

fbruton (changes requested):
Recommend suggesting alternative methods in the Javadoc for deprecated methods where applicable.

jpdahlke:
> Recommend suggesting alternative methods in the Javadoc for deprecated methods where applicable.

Some of these have no other option and will be removed as part of future updates. I marked them as Deprecated vs removing them to reduce the overall scope of changes. I would be happy to address this in the next round of changes.

fbruton (approved):

PR #1026, opened 2024-12-17
unnecessary modifier :: variables declared in try-with-resources are …

dev-mlb:
…implicitly final


jpdahlke:
Is this a rule incompatibility between errorprone and sonarlint?

dev-mlb:
> Is this a rule incompatibility between errorprone and sonarlint?

Found these when fixing the warnings from pmd

drivenflywheel (approved):

fbruton (approved):

PR #1027, opened 2024-12-19
Improve performance of ShortNameComparator.

jdcove2:
This is an updated version of #135 .

### DIFF
--- a/src/main/java/emissary/util/ShortNameComparator.java
+++ b/src/main/java/emissary/util/ShortNameComparator.java
@@ -41,8 +41,8 @@ public class ShortNameComparator implements Comparator<IBaseDataObject>, Seriali
             // Get character length of level.
             final int nextIndex1 = s1.indexOf(Family.SEP, index1);
             final int nextIndex2 = s2.indexOf(Family.SEP, index2);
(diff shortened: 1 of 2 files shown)


drivenflywheel (changes requested):
Do we think the performance improvements will be primarily speed, memory utilization, or both?

ldhardy (commented):
Is there a trick to recreate the profile you describe in #135 ?  I've been trying to use the profiler on our main branch to observe the described behavior and I'm not having any luck.  I guess there have to be a lot of children extracted - I tried dumping in all kinds of archive input to see if I can see something similar on the profiler, but I'm not observing what you were.

ldhardy (approved):

ldhardy:
> Yes, this optimization is designed for objects with a large number of children (i.e. 100,000 or more). Let me know if you would like me to generate an object for you.

Nah, I believe you.  It's not that big of a change really.

ldhardy:
> Yes, this optimization is designed for objects with a large number of children (i.e. 100,000 or more). Let me know if you would like me to generate an object for you.

I meant the algorithm is basically the same, but yes, I can see how it's a less heavy handed implementation.

drivenflywheel (approved):

PR #1035, opened 2024-12-29
creates an abstract RefreshableServiceProviderPlace for config reloading

dev-mlb:
This is a first cut into handling place config refreshing -- this does not handle service key changes at this time

### DIFF
--- a/src/main/java/emissary/place/RefreshableServiceProviderPlace.java
+++ b/src/main/java/emissary/place/RefreshableServiceProviderPlace.java
@@ -4,6 +4,8 @@ import emissary.config.ConfigUtil;
 import emissary.config.Configurator;
 
 import org.apache.commons.collections4.CollectionUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -16,6 +18,9 @@ import javax.annotation.Nullable;
  */
 public abstract class RefreshableServiceProviderPlace extends ServiceProviderPlace {
 
+    private static final Logger logger = LoggerFactory.getLogger(RefreshableServiceProviderPlace.class);
+
+    private final Object allocatorLock = new Object();
     private final AtomicBoolean invalidated = new AtomicBoolean(false);
 
     public RefreshableServiceProviderPlace() throws IOException {}
@@ -60,6 +65,7 @@ public abstract class RefreshableServiceProviderPlace extends ServiceProviderPla
      * Invalidate a place that need to be refreshed.
      */
     public final void invalidate() {
+        logger.info("Place being marked as invalidated");
         setInvalidated(true);
     }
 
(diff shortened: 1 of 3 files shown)


jpdahlke:
😍

src/main/java/emissary/place/ServiceProviderPlace.java
```diff
@@ -925,6 +928,72 @@ protected void deregisterFromDirectory(List<String> keys) {
         }
     }
 
+    /**
+     * Reset the keys and reload the {@link Configurator}
+     *
+     * @throws IOException if there is an issue loading the config
+     */
+    protected void refresh() throws IOException {
```
drivenflywheel, line 936: It occurs to me that this might not be safe to call while the place is actively processing an IBDO, or while trying to identify the next place that should process a payload.

src/main/java/emissary/place/RefreshableServiceProviderPlace.java
```diff
@@ -0,0 +1,126 @@
...
+
+    protected RefreshableServiceProviderPlace(final InputStream configStream, final String placeLocation) throws IOException {
+        super(configStream, placeLocation);
+    }
+
+    public boolean isInvalid() {
+        return this.invalid.get();
+    }
+
+    public void invalidate() {
```
drivenflywheel, line 53: 1) Might want to add a Javadoc comment stating that `invalidate` must be called before calling `refresh`.
2) Might want to mark this method as `final` and do the same to the `refresh` overloads

src/main/java/emissary/place/RefreshableServiceProviderPlace.java
```diff
@@ -0,0 +1,126 @@
...
+     * Reinitialize the place by reloading the configurator and reconfiguring the place
+     *
+     * @param configStream the config data as an {@link InputStream}
+     */
+    public synchronized void refresh(final InputStream configStream) {
+        try {
+            if (this.invalid.get()) {
+                this.configG = reloadConfigurator(configStream);
+                reconfigurePlace();
+            }
```
drivenflywheel, line 82:
```suggestion
            } else {
                logger.warn("cannot refresh place configuration without first calling invalidate; no reconfiguration performed");
            }
```

src/main/java/emissary/place/RefreshableServiceProviderPlace.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * Reinitialize the place by reloading the configurator and reconfiguring the place
+     */
+    public synchronized void refresh() {
+        try {
+            if (this.invalid.get()) {
+                this.configG = reloadConfigurator();
+                reconfigurePlace();
+            }
```
drivenflywheel, line 65: Might even want this bumped up to ERROR level
```suggestion
            } else {
                logger.warn("cannot refresh place configuration without first calling invalidate; no reconfiguration performed");
            }
```

src/test/java/emissary/place/RefreshableServiceProviderPlaceTest.java
```diff
@@ -0,0 +1,96 @@
...
+    @AfterEach
+    public void tearDown() throws Exception {
+        super.tearDown();
+        assertNotNull(place);
+        place.shutDown();
+        place = null;
+    }
+
+    @Test
+    void testReconfigure() {
```
drivenflywheel, line 48: Consider adding another test (or block within this test) that calls `place.refresh()` without first calling `invalidate()`

src/main/java/emissary/place/RefreshableServiceProviderPlace.java
```diff
@@ -0,0 +1,126 @@
...
+import org.apache.commons.collections4.CollectionUtils;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
+import javax.annotation.Nullable;
+
+/**
+ *
```
drivenflywheel, line 15:
```suggestion
 * ServiceProviderPlace that supports on-demand refresh of its configuration
```

drivenflywheel (changes requested):
Looks like a good approach.   Minor changes suggested

drivenflywheel (dismissed):
LGTM.   The only potential change I see might be to also mark the new `isValid()` and `setInvalid()` methods as `final`

drivenflywheel (dismissed):

dev-mlb:
Sorry, the directory place change doesn't need to go out with this round of changes.

drivenflywheel (approved):

fbruton (approved):

PR #1037, opened 2025-01-06
Added cleaner Executrix temp file names API, deprecated array-based API

drivenflywheel:


rpg36:
This makes me happy! Anything to improve executrix!

dev-mlb (approved):
I like this change. The only thing I'd maybe change is possibly hold off on the *CommandPlaces -- get the changes in and then activate them in a follow up. I could be convinced either way.

rpg36 (approved):
I have no preference personally as to if this should be broken up into 2 phases or not as @dev-mlb suggested.

drivenflywheel:
> I like this change. The only thing I'd maybe change is possibly hold off on the *CommandPlaces -- get the changes in and then activate them in a follow up. I could be convinced either way.

I thought about that too, but these changes to both Places are method-local.  Transparent to downstream, which integrates this with zero code changes.

PR #1040, opened 2025-01-07
Implements an ArrowRootAllocationProvider SPI

drewfarris:
Other utilities that run within Emissary would benefit from a central source for getting access to an Arrow `RootAllocator`. This PR implements the `ArrowRootAllocationProvider` that creates a new `RootAllocator` and provides a static method for retrieving that instance. It also properly handles calling `close()` on the allocator at shutdown time. It can be initialized by including  it in the `META-INF/service/emissary.spi.InitializationProvider` resource file.

In the current state, we leave it up to the consumer to decide whether to create a child allocator using the reference to the root allocator, or to use the root allocator directly. We could decide to force consumers to create child allocators instead of providing references to the root allocator, but I'm not convinced that it is worth doing that at this point.

See also:
- Arrow Java [documentation on Memory Management](https://arrow.apache.org/docs/java/memory.html)
- Arrow [RootAllocator javadoc](https://arrow.apache.org/docs/dev/java/reference/org/apache/arrow/memory/RootAllocator.html)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -202,6 +202,13 @@
         <groupId>org.apache.arrow</groupId>
         <artifactId>arrow-memory-core</artifactId>
         <version>${dep.arrow.version}</version>
+        <exclusions>
+          <!-- conflicts with guava's dependency, which is an older version of the checker framework -->
+          <exclusion>
+            <groupId>org.checkerframework</groupId>
+            <artifactId>checker-qual</artifactId>
+          </exclusion>
+        </exclusions>
       </dependency>
       <dependency>
         <groupId>org.apache.arrow</groupId>
@@ -415,13 +422,12 @@
     <dependency>
       <groupId>org.apache.arrow</groupId>
       <artifactId>arrow-memory-core</artifactId>
-      <exclusions>
-        <!-- conflicts with guava's dependency, which is an older version of the checker framework -->
-        <exclusion>
-          <groupId>org.checkerframework</groupId>
-          <artifactId>checker-qual</artifactId>
-        </exclusion>
-      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.arrow</groupId>
+      <artifactId>arrow-memory-netty</artifactId>
+      <!-- consumers of emissary can use either arrow-memory-netty or arrow-memory-unsafe to provide an allocator -->
+      <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -401,6 +412,17 @@
...
     </dependency>
+    <dependency>
+      <groupId>org.apache.arrow</groupId>
+      <artifactId>arrow-memory-core</artifactId>
+      <exclusions>
+        <!-- conflicts with guava's dependency, which is an older version of the checker framework -->
+        <exclusion>
+          <groupId>org.checkerframework</groupId>
+          <artifactId>checker-qual</artifactId>
+        </exclusion>
```
drewfarris, line 423: I'm not totally certain this is the best way to approach this, but it's a good point for discussion

drewfarris:
This may need additional work. After doing some brief testing, if someone creates a child allocator from the root allocator, and that child allocator is not properly closed, calling shutdown on the provider seems to raise an IllegalStateException. Changing this to a draft PR in the meantime.

pom.xml
drewfarris: This is marked outdated because I moved the exclusion to the `dependencyManagement` section, but it's still relevant to discuss.

dev-mlb (approved):
This looks good to me, not really sure how to test it. I started up Emissary and got:
```
2025-01-15T19:51:03.308 [localhost-8001] INFO  org.apache.arrow.memory.BaseAllocator -   - Debug mode disabled. Enable with the VM option -Darrow.memory.debug.allocator=true.
2025-01-15T19:51:03.308 [localhost-8001] INFO  org.apache.arrow.memory.DefaultAllocationManagerOption -   - allocation manager type not specified, using netty as the default type
2025-01-15T19:51:03.309 [localhost-8001] INFO  org.apache.arrow.memory.CheckAllocator -   - Using DefaultAllocationManager at memory-netty/16.1.0/arrow-memory-netty-16.1.0.jar!/org/apache/arrow/memory/netty/DefaultAllocationManagerFactory.class
2025-01-15T19:51:03.343 [localhost-8001] INFO  emissary.spi.SPILoader -   - Initialized emissary.spi.ArrowRootAllocatorProvider
```
so it started up correctly. Not sure if we want to turn down the logging on `org.apache.arrow` to WARN?

dev-mlb:
```
[WARNING] /Users/matt.brake/IdeaProjects/nsa-emissary/src/main/java/emissary/spi/ArrowRootAllocatorProvider.java:[34,56] [FieldMissingNullable] Field is assigned (or compared against) a definitely null value but is not annotated @Nullable
    (see https://errorprone.info/bugpattern/FieldMissingNullable)
  Did you mean 'private static @Nullable BufferAllocator arrowRootAllocator = null;'?
[WARNING] /Users/matt.brake/IdeaProjects/nsa-emissary/src/main/java/emissary/spi/ArrowRootAllocatorProvider.java:[34,35] [NonFinalStaticField] Static fields should almost always be final.
    (see https://errorprone.info/bugpattern/NonFinalStaticField)
[WARNING] /Users/matt.brake/IdeaProjects/nsa-emissary/src/main/java/emissary/spi/ArrowRootAllocatorProvider.java:[82,33] [FieldMissingNullable] Field is assigned (or compared against) a definitely null value but is not annotated @Nullable
    (see https://errorprone.info/bugpattern/FieldMissingNullable)
  Did you mean 'private static @Nullable BufferAllocator arrowRootAllocator = null;'?
```

PR #1041, opened 2025-01-07
Add RegularExpresionTestUtil for testing regexes

ldhardy:
There are multiple places we're doing something similar to test regex, and plenty of places we need to, but are not yet.  It would be nice to clean it all up and use one utility class.  Needed for another effort.


src/test/java/emissary/test/util/RegularExpressionTestUtil.java
```diff
@@ -0,0 +1,82 @@
...
+                assertFalse(dmm.find(),
+                        String.format(" -- Pattern SHOULD NOT match the regex[%s], but did: %s", patternUnderTest.pattern(), dontMatchMe));
+                fineGrainTestCount++;
+            }
+        }
+        if (fineGrainTestCount == 0) {
+            fail("Didn't test anything.");
+        } else {
+            logger.debug("Successfully asserted {} regex test cases.", fineGrainTestCount);
+        }
```
drivenflywheel, line 78: (optional change)
```suggestion
        assertTrue(fineGrainTestCount > 0, "No regex assertions performed.");
        logger.debug("Successfully asserted {} regex test cases.", fineGrainTestCount);
```

src/test/java/emissary/test/util/RegularExpressionTestUtil.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * A method to test a list of values that should and should not match a particular regular expression. One of the two
+     * lists may be empty or null, but not both.
+     * 
+     * @param patternUnderTest - Required. The pre-compiled pattern used to test against the provided values. Must not be
+     *        null.
+     * @param shouldMatch - Optional. The list of strings that should match the regular expression.
+     * @param shouldNotMatch - Optional. The list of strings that should not match the regular expression.
```
drivenflywheel, line 50: (nit) - these dashes duplicate the endashes auto-inserted by Javadoc tooling, so the rendered Javadoc looks like this:
> `patternUnderTest` – - Required. The pre-compiled pattern used to test against the provided values. Must not be null.
> ` shouldMatch` – - Optional. The list of strings that should match the regular expression
> `shouldNotMatch` – - Optional. The list of strings that should not match the regular expression.
```suggestion
     * @param patternUnderTest Required. The pre-compiled pattern used to test against the provided values. Must not be
     *        null.
     * @param shouldMatch Optional. The list of strings that should match the regular expression.
     * @param shouldNotMatch Optional. The list of strings that should not match the regular expression.
```

drivenflywheel (dismissed):
Looks like a good addition.  We should add at least cursory test coverage now to ensure that future regressions that are accidentally introduced will get caught prior to downstream integration.

drivenflywheel:
This diff looks to have gone a bit wonky.  Lots of out-of-scope files now showing up as changed

ldhardy:
should be good now.  Just one file.

dev-mlb (approved):

dev-mlb (changes requested):
```
#14 67.84 /opt/emissary/src/test/java/emissary/test/util/RegularExpressionTestUtil.java:[19,7] [PrivateConstructorForUtilityClass] Classes which are not intended to be instantiated should be made non-instantiable with a private constructor. This includes utility classes (classes with only static members), and the main class.
#14 67.84     (see https://errorprone.info/bugpattern/PrivateConstructorForUtilityClass)
#14 67.84   Did you mean 'public final class RegularExpressionTestUtil {'?
#14 67.84 [WARNING] /opt/emissary/src/test/java/emissary/test/util/RegularExpressionTestUtil.java:[20,28] [NonFinalStaticField] Static fields should almost always be final.
#14 67.84     (see https://errorprone.info/bugpattern/NonFinalStaticField)
```

dev-mlb (approved):

PR #1042, opened 2025-01-07
change EmissaryServer to use an enum for mode

jpdahlke:


src/main/java/emissary/directory/EmissaryNode.java
```diff
@@ -64,9 +64,9 @@ public class EmissaryNode {
     /** Property that determines if server will shut down in the event a place fails to start */
     public static final String STRICT_STARTUP_MODE = "strict.mode";
 
-    // types are feeder, worker, standalone
-    // TODO: make an enum for these
-    private static final String DEFAULT_NODE_MODE = "standalone";
+    public enum EmissaryMode {
```
drivenflywheel, line 67: Should we consider naming this `ServerMode` instead?  More visually distinguishable from `EmissaryNode` and it's consistent with the `ServerModeConverter`

jpdahlke: Yeah, that's a good point. Mode and Node do look similar and consistency is good.

drivenflywheel (dismissed):

drivenflywheel (approved):

dev-mlb (approved):

PR #1044, opened 2025-01-09
Adding a utility to provide test data for specific unicode handling issues.

ldhardy:
There are a few places where we need more robust testing that we aren't making false assumptions about what the byte array contains and how (we attempt) to do clever things with it.  When our code or the 3rd party libraries we use are defaulting to baseless assumptions about how we can  move around in the bytes, we end up mangling the data in ways that isn't always obvious.  This helper class and the explaining test should allow us to test more edge cases, and avoid using dependencies that don't handle Unicode well.   I envision adding more samples as we come across more cases that keep tripping us up, but for now I need the XML map for one of my tickets, and I have plans for the emoji string for a couple of other tests. 

The entirety of this change should be test only, used from the emissary test jar.  I picked the icu4j version for consistency but I'd really like to always use the latest, it should be at the top of our list of dependencies to keep updated once we start using the icu4j utilities more often (as we should be).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,6 +64,7 @@
     <dep.guava.version>32.1.2-jre</dep.guava.version>
     <dep.httpclient.version>5.2.1</dep.httpclient.version>
     <dep.httpcore.version>5.2.1</dep.httpcore.version>
+    <dep.icu4j.version>73.2</dep.icu4j.version>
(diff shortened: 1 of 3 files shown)


dev-mlb (approved):

drivenflywheel (approved):

drivenflywheel:
Nice addition!

PR #1045, opened 2025-01-13
getStringParameter deprecation should suggest getParameterAsString

drivenflywheel:
Update `getStringParameter` Javadoc to suggest `getParameterAsString` when a single value is expected, as this is the most common situation and will log a warning is the expectation is violated


cfkoehler (approved):

sambish5 (approved):
:rocket:

PR #1047, opened 2025-01-17
fallback to original behavior for getParameterAsString

dev-mlb:


fbruton:
Should a method to retrieve the first string parameter also be included in this PR?

drivenflywheel:
> Should a method to retrieve the first string parameter also be included in this PR?

Can be a separate PR.  Can already use `getParameterAsStrings` and work from there.

drivenflywheel (approved):

PR #1061, opened 2025-02-10
pmd :: CollapsibleIfStatement

sambish5:
resolves the pmd warnings, as well as the sonar complaints about these collapsible if statements

### DIFF
--- a/src/main/java/emissary/kff/Ssdeep.java
+++ b/src/main/java/emissary/kff/Ssdeep.java
@@ -202,11 +202,9 @@ public final class Ssdeep {
                     // Therefore this secondary signature condition can
                     // only potentially be true if the main signature
                     // condition (which we've already checked) is true.
-                    if ((rollingHash % (this.blockSize * 2)) == ((this.blockSize * 2) - 1)) {
-                        if (this.fuzzLen2 < (SPAMSUM_LENGTH / 2 - 1)) {
-                            this.fuzzHash2[this.fuzzLen2++] = b64EncodeLowBits(this.sumHash2);
-                            this.sumHash2 = HASH_INIT;
-                        }
+                    if ((rollingHash % (this.blockSize * 2)) == ((this.blockSize * 2) - 1) && this.fuzzLen2 < (SPAMSUM_LENGTH / 2 - 1)) {
+                        this.fuzzHash2[this.fuzzLen2++] = b64EncodeLowBits(this.sumHash2);
+                        this.sumHash2 = HASH_INIT;
                     }
                 }
             }
--- a/src/main/java/emissary/roll/Roller.java
+++ b/src/main/java/emissary/roll/Roller.java
@@ -154,11 +154,9 @@ public class Roller implements Runnable {
     private boolean shouldRoll(long start) {
         // verify both time and progress are set
         // if we're below max we'll check the interval
-        if (period > 0 && max > 0 && progress.get() < max) {
+        if (period > 0 && max > 0 && progress.get() < max && ((start - lastRun + 100) < getIntervalInMillis())) {
             // we fired a progress run or delayed start due to starvation. add 100 for clock skew
-            if ((start - lastRun + 100) < getIntervalInMillis()) {
-                return false;
-            }
+            return false;
         }
         return true;
     }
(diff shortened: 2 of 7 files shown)


jpdahlke (changes requested):

src/main/java/emissary/roll/Roller.java
```diff
@@ -154,11 +154,9 @@ private long getIntervalInMillis() {
     private boolean shouldRoll(long start) {
         // verify both time and progress are set
         // if we're below max we'll check the interval
-        if (period > 0 && max > 0 && progress.get() < max) {
+        if (period > 0 && max > 0 && progress.get() < max && ((start - lastRun + 100) < getIntervalInMillis())) {
             // we fired a progress run or delayed start due to starvation. add 100 for clock skew
```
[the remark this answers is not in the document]
sambish5: moved the actions of this if to a separate method

src/main/java/emissary/kff/Ssdeep.java
```diff
@@ -202,11 +202,9 @@ private void applyBytes(final RollingState rollState, final byte[] buffer, final
...
                     // only potentially be true if the main signature
                     // condition (which we've already checked) is true.
-                    if ((rollingHash % (this.blockSize * 2)) == ((this.blockSize * 2) - 1)) {
-                        if (this.fuzzLen2 < (SPAMSUM_LENGTH / 2 - 1)) {
-                            this.fuzzHash2[this.fuzzLen2++] = b64EncodeLowBits(this.sumHash2);
-                            this.sumHash2 = HASH_INIT;
-                        }
+                    if ((rollingHash % (this.blockSize * 2)) == ((this.blockSize * 2) - 1) && this.fuzzLen2 < (SPAMSUM_LENGTH / 2 - 1)) {
+                        this.fuzzHash2[this.fuzzLen2++] = b64EncodeLowBits(this.sumHash2);
+                        this.sumHash2 = HASH_INIT;
```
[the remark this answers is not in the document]
sambish5: for now I just undid the changes in this class. The refactoring would be out of the scope of this PR, so for now just keeping it as the back-to-back if statements.

src/main/java/emissary/roll/Roller.java
```diff
@@ -156,11 +156,12 @@ private boolean shouldRoll(long start) {
...
-                return false;
-            }
+            return checkClockSkew(start);
         }
         return true;
     }
 
+    private boolean checkClockSkew(long start) {
+        return (start - lastRun + 100) >= getIntervalInMillis();
+    }
```
fbruton, line 166: Suggest grouping the method like this. The method names may need to be tweeked.

```
    /*
     * There are a couple of conditions where we would not want to execute when both a progress value and time schedule are
     * configured: - Time based roll happened and a progress run is scheduled - Progress roll happened between schduled runs
     */
    private boolean shouldRoll(long start) {
        // verify both time and progress are set
        // if we're below max we'll check the interval
        // we fired a progress run or delayed start due to starvation. add 100 for clock skew
        return ifProgressComplete() || isProgressCompleteWithCheckClockSkew(start);
    }

    private boolean ifProgressComplete() {
        return period <= 0 || max <= 0 || progress.get() >= max;
    }

    private boolean isProgressCompleteWithCheckClockSkew(long start) {
        return (start - lastRun + 100) >= getIntervalInMillis();
    }
    ```

Comment (jpdahlke) [reply src/main/java/emissary/roll/Roller.java:166]: I guess I was trying to think of it without inverting the logic too much, but it takes some extra negation to get there. I think the comment by @fbruton reads better (which was sorta the ultimate goal though we've really gone next level here) Maybe if I cleaned up the recommendation?
```
    private boolean shouldRoll(long start) {
        return isProgressComplete() || isIntervalExceeded(start);
    }
    
    // verify if both time and progress are set, then see if we have met the max
    private boolean isProgressComplete() {
        return period <= 0 || max <= 0 || progress.get() >= max;
    }
    
    // check the current interval, maybe we fired a progress run or delayed start due to starvation
    private boolean isIntervalExceeded(long start) {
        // add 100 for clock skew
        return (start - lastRun + 100) >= getIntervalInMillis();
    }
```

jpdahlke: Recommend reverting the change in this class so we can address later.

sambish5: has been reverted for now

jpdahlke (approved):

fbruton (approved):

PR #1062, opened 2025-02-10
Added Configurator findStringMatchMultiMap overload that takes a preserveOrder flag

drivenflywheel:
Useful for ensuring deterministic order of configured multi-map entries

### DIFF
--- a/src/main/java/emissary/config/ServiceConfigGuide.java
+++ b/src/main/java/emissary/config/ServiceConfigGuide.java
@@ -26,6 +26,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


sambish5 (dismissed):

src/main/java/emissary/config/ServiceConfigGuide.java
```diff
@@ -829,12 +830,44 @@ public Map<String, String> findStringMatchMap(@Nullable final String theParamete
...
+     * Example config entries
+     *    FOO_ONE: AAA
+     *    FOO_TWO: BBB
+     *    FOO_TWO: CCC
+     * Calling findStringMatchMap("FOO_",true)
+     * will yield a map with Sets
+     *     ONE -> {AAA}
+     *     TWO -> {BBB,CCC}
+     * }
+     * </pre>
```
ldhardy, line 856: Thank you for this example, it helps a lot.

src/test/java/emissary/config/ServiceConfigGuideTest.java
```diff
@@ -645,4 +648,52 @@ void testFindRequiredStringEntry() {
         assertNotNull(ex.getMessage(), "Exception message should not be null");
         assertTrue(ex.getMessage().contains(missingProp), "Exception message should name the missing config entry");
     }
+
+    @Test
+    void testFindStringMatchMultiMapOrdered() {
+        // this test fails if ORDERED is set to false due to results ordering
+        final boolean ORDERED = true;
```
ldhardy, line 655: confirmed this fails when false.

src/test/java/emissary/config/ServiceConfigGuideTest.java
```diff
@@ -645,4 +648,52 @@ void testFindRequiredStringEntry() {
...
+        while (expectedIterator.hasNext()) {
+            assertTrue(actualsIterator.hasNext());
+
+            i++;
+            var expectedEntry = expectedIterator.next();
+            var actualEntry = actualsIterator.next();
+            assertEquals(expectedEntry.getKey(), actualEntry.getKey(), "failed on key #" + i);
+            assertEquals(expectedEntry.getValue(), actualEntry.getValue(), "value mismatch for key '" + actualEntry.getKey() + "'");
+        }
+
```
ldhardy, line 696:
```suggestion
assertTrue(i>0);
```
I know it's pedantic but I've done this myself where I accidentally forgot to fix a test I was temporarily messing with and turns out the tests still pass because the test case list was empty and it never hit the asserts.

ldhardy (dismissed):

src/main/java/emissary/config/ServiceConfigGuide.java
drivenflywheel: > I assume that the default case (preserve order false) would mean that the TWO values could be in any arbitrary order depending on whether it was a Coordination place config laid over top.

Depends on how the key hashes get sorted, which isn't a sort order we can control with our config entries.

src/test/java/emissary/config/ServiceConfigGuideTest.java
drivenflywheel: Addressing this by adding an earlier `assertFalse(actual.isEmpty())`, but that's a good catch

ldhardy (approved):

sambish5 (approved):

PR #1064, opened 2025-02-12
Add optional RegressionTest encoders that always hash.

jdcove2:
This PR adds an optional encoder set that will always hash SeekableByteChannels and alternate view byte arrays without regard to whether the data is "printable". They can be used by any class that extends RegressionTest by overriding the getEncoders() method and returning IBaseDataObjectXmlCodecs.ALWAYS_SHA256_ELEMENT_ENCODERS.

### DIFF
--- a/src/test/java/emissary/test/core/junit5/RegressionTest.java
+++ b/src/test/java/emissary/test/core/junit5/RegressionTest.java
@@ -242,7 +242,7 @@ public abstract class RegressionTest extends ExtractionTest {
      * @return a value optionally containing the generated hash
      */
     protected Optional<String> hashBytesIfNonPrintable(byte[] bytes, final boolean alwaysHash) {
-        if (alwaysHash || (ArrayUtils.isNotEmpty(bytes) && ByteUtil.containsNonIndexableBytes(bytes))) {
+        if (ArrayUtils.isNotEmpty(bytes) && (alwaysHash || ByteUtil.containsNonIndexableBytes(bytes))) {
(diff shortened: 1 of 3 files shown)


fbruton (dismissed):

src/test/java/emissary/test/core/junit5/RegressionTest.java
```diff
@@ -230,10 +238,11 @@ protected void checkAnswersPreHookLogEvents(List<SimplifiedLogEvent> simplifiedL
      * Generates a SHA 256 hash of the provided bytes if they contain any non-printable characters
      * 
      * @param bytes the bytes to evaluate
+     * @param alwaysHash overrides the non-printable check and always hashes the bytes.
      * @return a value optionally containing the generated hash
      */
-    protected Optional<String> hashBytesIfNonPrintable(byte[] bytes) {
-        if (ArrayUtils.isNotEmpty(bytes) && ByteUtil.containsNonIndexableBytes(bytes)) {
+    protected Optional<String> hashBytesIfNonPrintable(byte[] bytes, final boolean alwaysHash) {
+        if (alwaysHash || (ArrayUtils.isNotEmpty(bytes) && ByteUtil.containsNonIndexableBytes(bytes))) {
```
drivenflywheel, line 245: Wouldn't we still want to return `Optional.empty()` if `bytes` is empty?

fbruton (approved):

SemperFidelisTyrannosaurus (approved):

PR #1072, opened 2025-02-25
Refresh IServiceProviderPlaces - invalidate, recreate, swap

dev-mlb:
Sample output -- 

```
2025-02-25T00:19:06.637 [localhost-8001] INFO  emissary.place.sample.RefreshablePlace - emissary-knight.png *.REFRESH_EXAMPLE.ANALYZE.http://localhost:8001/RefreshablePlace - SAMPLE_CONFIG 200
```
Updated emissary.place.sample.RefreshablePlace.cfg -- SAMPLE_CONFIG = 5000
http://localhost:8001/api/refresh
```
2025-02-25T00:19:20.427 [localhost-8001] INFO  emissary.server.EmissaryServer -   - Invalidating places that need to be reconfigured
2025-02-25T00:19:20.427 [localhost-8001] INFO  emissary.place.ServiceProviderRefreshablePlace -   - Place[emissary.place.sample.RefreshablePlace] being marked as invalidated
2025-02-25T00:19:26.644 [localhost-8001] INFO  emissary.pickup.file.FilePickUpPlace - emissary-knight.png  - **Deploying an agent for emissary-knight.png and object 80cb77c7-e054-4d59-a0d2-96e23bb97ec6 forms=[UNKNOWN] simple=simple
2025-02-25T00:19:26.644 [localhost-8001] INFO  emissary.pickup.file.FilePickUpPlace -   - emissary-knight.png processed and moved to done area /Users/matt.brake/IdeaProjects/emissary/target/data/DoneData
2025-02-25T00:19:26.645 [localhost-8001] INFO  emissary.directory.DirectoryPlace - emissary-knight.png  - *.REFRESH_EXAMPLE.ANALYZE.http://localhost:8001/RefreshablePlace$5040 has been invalidated, attempting refresh ....
2025-02-25T00:19:26.647 [localhost-8001] INFO  emissary.directory.DirectoryPlace - emissary-knight.png  - *.REFRESH_EXAMPLE.ANALYZE.http://localhost:8001/RefreshablePlace$5040 has been refreshed
2025-02-25T00:19:26.647 [localhost-8001] INFO  emissary.place.sample.RefreshablePlace - emissary-knight.png *.REFRESH_EXAMPLE.ANALYZE.http://localhost:8001/RefreshablePlace - SAMPLE_CONFIG 5000
```

PR #1070, opened 2025-02-22
Refresh IServiceProviderPlaces - pause, refresh, unpause

dev-mlb:
Brute force way to refresh services/places -- 

1. Invalidate services
2. Pause server
3. Wait for agents to drain
4. Refresh services
5. Unpause if all succeed, kill the server if failures

To force refresh, hit the api endpoint -- http://localhost:8001/api/refresh

Sample output: 
```
2025-02-22T23:33:29.280 [localhost-8001] INFO  emissary.server.EmissaryServer -   - Invalidating places that need to be reconfigured
2025-02-22T23:33:29.282 [localhost-8001] INFO  emissary.place.RefreshableServiceProviderPlace -   - Place[emissary.place.sample.RefreshablePlace] being marked as invalidated
2025-02-22T23:33:29.284 [localhost-8001] INFO  emissary.pickup.file.FilePickUpPlace -   - *** Pausing emissary.pickup.file.FileDataServer for emissary.pickup.file.FilePickUpPlace
2025-02-22T23:33:29.284 [localhost-8001] INFO  emissary.server.EmissaryServer -   - Waiting for server to drain
2025-02-22T23:33:33.732 [localhost-8001] INFO  emissary.core.Pausable -   - emissary.pickup.file.FileDataServer currently paused, sleeping for 60000
2025-02-22T23:33:34.294 [localhost-8001] INFO  emissary.place.RefreshableServiceProviderPlace -   - Place refresh performed successfully
2025-02-22T23:33:34.294 [localhost-8001] INFO  emissary.pickup.file.FilePickUpPlace -   - *** Unpausing emissary.pickup.file.FileDataServer for emissary.pickup.file.FilePickUpPlace

```

PR #1073, opened 2025-02-26
refresh :: "light" and "hard" refresh logic

dev-mlb:
This pr is a combination of #1070 and #1072 -- this has the ability to handle "light" vs "hard" refreshes. 
- **"light" refresh** -- this is a no-downtime approach that invalidates a ServiceProviderRefreshablePlace. When the place is then pulled from DirectoryPlace, the place is recreated using the same keys for DirectoryPlace and Namespace, but the configurator is reloaded and the place is able to reload a subset of its configs. 
- **"hard" refresh** -- the server is paused and there is a wait for the MoblieAgents to drain. Once the server is fully idle, all ServiceProviderRefreshablePlace existing keys are removed from DirectoyPlace and Namespace, and the places are fully recreated. This allows for changes to service names, proxies, deny lists, etc. The server is then unpaused to resume processing. Any failure in refresh would put the server in a bad state, so the server is shutdown. 

How to test:
- Curl POST request:
  - curl --anyauth -u emissary -X POST -H "X-Requested-By: emissary"  http://localhost:8001/api/invalidate
  - curl --anyauth -u emissary -X POST -H "X-Requested-By: emissary"  http://localhost:8001/api/refresh
- Command line
  - ./emissary server --invalidate
  - ./emissary server --refresh


src/main/java/emissary/directory/DirectoryPlace.java
```diff
@@ -1186,10 +1205,6 @@ public int removePlaces(final List<String> keys) {
      */
     @Override
     public int irdRemovePlaces(@Nullable final List<String> keys, final boolean propagating) {
-        if (this.emissaryNode.isStandalone()) {
```
dev-mlb, line 1189: This could have some integration implications...in standalone mode, if a place fails to start the keys should be removed but they're not.

src/main/java/emissary/place/ServiceProviderPlace.java
```diff
@@ -218,6 +222,15 @@ protected ServiceProviderPlace(InputStream configStream, String placeLocation) t
         this(configStream, null, placeLocation);
     }
 
+    /**
+     * Create the place based on another ServiceProviderPlace
+     *
+     * @param place the ServiceProviderRefreshablePlace to clone
+     */
+    protected ServiceProviderPlace(final ServiceProviderPlace place) {
```
jpdahlke, line 230: Should this take a ServiceProviderRefreshablePlace as an argument?

src/main/java/emissary/place/ServiceProviderPlace.java
```diff
@@ -273,15 +285,26 @@ protected Configurator loadConfigurator(@Nullable String placeLocation) throws I
...
      */
     protected void setupPlace(@Nullable String theDir, String placeLocation) throws IOException {
+        setupPlace(theDir, placeLocation, true);
+    }
+
+    /**
+     * Help the constructor get the place running
+     *
+     * @param theDir name of our directory
+     */
```
jpdahlke, line 295: Add remaining param for this javadoc

src/main/java/emissary/place/ServiceProviderPlace.java
```diff
@@ -313,10 +336,12 @@ protected void setupPlace(@Nullable String theDir, String placeLocation) throws
...
         }
 
-        // Register with the directory
-        // This pushes all our keys out to the directory which
-        // sends them on in turn to peers, &c. in the p2p network
-        register();
+        if (register) {
+            // Register with the directory
+            // This pushes all our keys out to the directory which
+            // sends them on in turn to peers, &c. in the p2p network
```
jpdahlke, line 342:
```suggestion
```

src/main/java/emissary/server/api/Refresh.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(Refresh.class);
+
+    public static final String INVALIDATE = "invalidate";
+    public static final String REFRESH = "refresh";
+
+    @POST
+    @Path("/" + INVALIDATE)
+    @Produces(MediaType.TEXT_HTML)
+    public Response shutdownNow(@Context HttpServletRequest request) {
```
jpdahlke, line 28:
```suggestion
    public Response invalidatePlaces(@Context HttpServletRequest request) {
```

src/main/java/emissary/server/api/Refresh.java
```diff
@@ -0,0 +1,50 @@
...
+        } catch (Exception e) {
+            LOG.warn("Exception trying to invalidate places", e);
+            return Response.serverError().entity("error trying to invalidate places").build();
+        }
+    }
+
+    @POST
+    @Path("/" + REFRESH)
+    @Produces(MediaType.TEXT_HTML)
+    public Response forceShutdown(@Context HttpServletRequest request) {
```
jpdahlke, line 41:
```suggestion
    public Response refreshPlaces(@Context HttpServletRequest request) {
```

jpdahlke (changes requested):
Nits on first read through.

src/main/java/emissary/directory/DirectoryPlace.java
jpdahlke: If we end up spending too much time on this review we can break this out and get it integrated ahead of time.

src/main/java/emissary/place/ServiceProviderPlace.java
dev-mlb: ServiceProviderPlace doesn't know anything about ServiceProviderRefreshablePlace. I was having issues with the `ServiceProviderRefreshablePlace(ServiceProviderRefreshablePlace place)` constructor and java compiler adding `super()` which was calling `ServiceProviderPlace()` and causing unintended side-effects. I had to add this little ditty to get past it. I ❤️  Java.

src/main/java/emissary/server/api/Refresh.java
dev-mlb: Whoops

src/main/java/emissary/server/api/Refresh.java
dev-mlb: D'oh

src/main/java/emissary/place/ServiceProviderPlace.java
dev-mlb: On it

src/main/java/emissary/place/ServiceProviderRefreshablePlace.java
```diff
@@ -0,0 +1,169 @@
...
+/**
+ * ServiceProviderPlace that supports on-demand refresh of its configuration
+ */
+public abstract class ServiceProviderRefreshablePlace extends ServiceProviderPlace {
+
+    private static final Logger logger = LoggerFactory.getLogger(ServiceProviderRefreshablePlace.class);
+
+    private final Object allocatorLock = new Object();
+    private final AtomicBoolean invalidated = new AtomicBoolean(false);
+    private final AtomicBoolean defunct = new AtomicBoolean(false);
```
drivenflywheel, line 27: I'm not sure I understand the intended role of `defunct`, but it seems to limit the place to at-most one refresh, even if `invalidate` is called again.  Is that intentional?

drivenflywheel (commented):
I like this approach - just trying to make sure we haven't accidentally prevented repeated periodic refreshes by making `defunct` a one-way gate

src/main/java/emissary/place/ServiceProviderRefreshablePlace.java
dev-mlb: Yes this is intentional. You shouldn't be able to get in the state where you can pull a defunct place, but I was trying to future proof the code. When a place is refreshed, a new object is created and swaps out with the old one in the namespace. I didn't want the off chance that we refresh a place more than once.

drivenflywheel (dismissed):

jpdahlke (changes requested):
Super exciting!

I was able to perform both operations using curl and commands. I used ImageMetadataJavaPlace to extend the new abstract class as an example and (with some edits) was able to get two different CSV behaviors and two different BUD behaviors based on what config entries I changed in ImageMetadataJavaPlace.cfg.

One thing I noticed was that when using just "invalidate", the configuration was applied and any file I copied into the InputData location was immediately picked up once the place was available. When using the full "refresh", it seems that the server does not immediately unpause even though the log claims it has. It will wait until exactly 60s after the pause message to resume pick-up. It does log the unpause, but that must just set the state to unpaused. Not a deal breaker, and that might be how it's worked in the past, it was just unexpected when comparing invalidate to refresh.

src/main/java/emissary/place/ServiceProviderRefreshablePlace.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /**
+     * Refresh specific constructor, basically clones a place and swaps out the Namespace ref
+     *
+     * @param place the ServiceProviderRefreshablePlace to clone
+     * @param register set true to register with directory place, false otherwise
+     * @throws IOException if there is an issue trying to create the place
+     */
+    public ServiceProviderRefreshablePlace(final ServiceProviderRefreshablePlace place, final boolean register) throws IOException {
```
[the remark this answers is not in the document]
dev-mlb: It's a constructor, so I cannot mark it abstract. That would be awesome tho. 

Good catch, Factory actually throws and AssertionError and I was only looking for Exception.

jpdahlke (approved):

fbruton (approved):
1. ./emissary server
2. vi ${emissary.config.dr}/emissary.place.sample.RefreshablePlace.cfg
3. ./emissary server --invalidate
4. ./emissary server --refresh
5. ./emissary config --place=emissary.place.sample.RefreshablePlace -p 8001 -h localhost
6. Verified that the modified configs are present.

PR #1074, opened 2025-02-26
adding ability to ExtractionTest to have OS specific answers in test answer files

sambish5:
This adds the ability to have specific `os-release` attributes in answer file elements so they will only be checked on that specified OS. If not, that element will not be checked in ExtractionTest's `checkAnswers()` method.

The current elements this `verifyOS()` check is applied to are:
- dataLength
- meta
- nometa
- data
- view
- noview

(Had to alter the original `dataLength` check, because the old check did not assume multiple dataLength tags in the answer file. With this we are assuming data lengths could be different across different OS, meaning multiple tags _could_ be possible)

### DIFF
--- a/src/test/java/emissary/test/core/junit5/ExtractionTest.java
+++ b/src/test/java/emissary/test/core/junit5/ExtractionTest.java
@@ -19,7 +19,6 @@ import org.jdom2.DataConversionException;
 import org.jdom2.Document;
 import org.jdom2.Element;
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
@@ -55,17 +54,18 @@ public abstract class ExtractionTest extends UnitTest {
                     KffDataObjectHandler.SET_FILE_TYPE);
     @Nullable
     protected IServiceProviderPlace place = null;
-    @SuppressWarnings("NonFinalStaticField")
-    private static String systemOsRelease = "";
+    @Nullable
+    private static final String SYSTEM_OS_RELEASE;
 
-    @BeforeAll
-    public static void setSystemOsInfo() {
+    static {
         if (OSReleaseUtil.isUbuntu()) {
-            systemOsRelease = "ubuntu";
+            SYSTEM_OS_RELEASE = "ubuntu";
         } else if (OSReleaseUtil.isCentOs()) {
-            systemOsRelease = "centos";
+            SYSTEM_OS_RELEASE = "centos";
         } else if (OSReleaseUtil.isRhel()) {
-            systemOsRelease = "rhel";
+            SYSTEM_OS_RELEASE = "rhel";
+        } else {
+            SYSTEM_OS_RELEASE = null;
         }
     }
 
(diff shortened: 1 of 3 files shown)


fbruton:
As a future enhancement, we should check the OS major version ID, which can be optional.

src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -391,6 +411,25 @@ protected void checkStringValue(Element meta, String data, String tname) {
         }
     }
 
+    protected boolean verifyOs(Element element) {
+        Attribute specifiedOs = element.getAttribute("os-release");
+        if (specifiedOs != null) {
+            String os = specifiedOs.getValue();
+            switch (os) {
+                case "ubuntu":
+                    return OSReleaseUtil.isUbuntu();
```
fbruton, line 420: I have concerns with reading a file each time an OS check is performed. Consider storing the OS release as a static field.

fbruton (changes requested):

src/test/java/emissary/test/core/junit5/ExtractionTest.java
sambish5: added in a static field to store the found OS on the first check, then compare the attribute tags to that field on other checks.

jpdahlke (approved):

cfkoehler (approved):

src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -54,6 +54,8 @@ public abstract class ExtractionTest extends UnitTest {
                     KffDataObjectHandler.SET_FILE_TYPE);
     @Nullable
     protected IServiceProviderPlace place = null;
+    @SuppressWarnings("NonFinalStaticField")
+    private static String systemOsRelease = "";
```
fbruton, line 58: Will this cause a thread safety issue when tests run in parallel?

jpdahlke: Since this is a value based on the machine running the suite, once set I don't believe this value would change during the entirety of the test run. It would be either the value or `null` and even in a race condition it would race to the same value.

It seemed like a `BeforeAll` to me, but then there would be multiple switch statements needed while we await a schema with a set of valid values.

sambish5: My thought was what Jason said. It would be null or the value, since it's based on the environment, and that would be the same, and then in a race condition, go to that same value.

Using a `BeforeAll` would require multiple switch statements, due to the existing one checking the answer file attribute as well as the OS environment at once the first check. We could split out the check to be `BeforeAll` if we wanted to be safe, would just add the OS env check before tests.

sambish5: due to the fact that we're _probably_ going to add other checks for OS info, I figured I'd move it to a `BeforeAll` for ease in the future.

src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -53,6 +55,19 @@ public abstract class ExtractionTest extends UnitTest {
...
+    @BeforeAll
+    public static void setSystemOsInfo() {
+        if (OSReleaseUtil.isUbuntu()) {
+            systemOsRelease = "ubuntu";
+        } else if (OSReleaseUtil.isCentOs()) {
+            systemOsRelease = "centos";
+        } else if (OSReleaseUtil.isRhel()) {
+            systemOsRelease = "rhel";
+        }
+    }
```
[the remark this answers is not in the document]
sambish5: made this change.

cfkoehler (approved):

fbruton (approved):

PR #1075, opened 2025-02-28
Update PlaceComparisonHelper to compare "dispose runnables".

jdcove2:
PlaceComparisonHelper was originally written such that it would not compare "dispose runnables". I now think we should compare "dispose runnables" so this PR makes that change.


SemperFidelisTyrannosaurus (approved):

drivenflywheel:
I thought the reason we weren't comparing the DisposeRunnables is that they're object references that don't serialize well (the `toString()` included the memory reference). Was something done to address that?

drivenflywheel (approved):

PR #1079, opened 2025-03-01
add undeclared deps, remove unused deps, tidy up management

jpdahlke:
```
[INFO] --- maven-dependency-plugin:3.8.1:analyze-only (analyze) @ emissary ---
[WARNING] Used undeclared dependencies found:
[WARNING]    com.google.code.findbugs:jsr305:jar:3.0.2:compile
[WARNING]    org.glassfish.jersey.test-framework:jersey-test-framework-core:jar:3.1.3:test
[WARNING]    org.junit.jupiter:junit-jupiter-api:jar:5.11.4:test
[WARNING]    org.junit.platform:junit-platform-commons:jar:1.11.4:test
[WARNING]    ch.qos.logback:logback-core:jar:1.5.13:compile
[WARNING]    jakarta.ws.rs:jakarta.ws.rs-api:jar:3.1.0:compile
[WARNING]    org.apache.httpcomponents.core5:httpcore5:jar:5.2.1:compile
[WARNING]    org.eclipse.jetty:jetty-servlet:jar:11.0.24:compile
[WARNING]    org.opentest4j:opentest4j:jar:1.3.0:test
[WARNING]    com.fasterxml.jackson.core:jackson-annotations:jar:2.15.2:compile
[WARNING]    com.fasterxml.jackson.core:jackson-core:jar:2.15.2:compile
[WARNING]    org.junit.jupiter:junit-jupiter-params:jar:5.11.4:test
[WARNING]    org.eclipse.jetty:jetty-http:jar:11.0.24:compile
[WARNING]    org.eclipse.jetty.toolchain:jetty-jakarta-servlet-api:jar:5.0.2:compile
[WARNING]    io.dropwizard.metrics:metrics-core:jar:4.2.25:compile
[WARNING] Unused declared dependencies found:
[WARNING]    com.github.spullara.mustache.java:compiler:jar:0.9.10:compile
[WARNING]    io.dropwizard.metrics:metrics-servlets:jar:4.2.25:compile
[WARNING]    org.glassfish.jersey.inject:jersey-hk2:jar:3.1.3:compile
[WARNING]    org.eclipse.jetty:jetty-xml:jar:11.0.24:compile
[WARNING]    org.glassfish.jersey.core:jersey-common:jar:3.1.3:compile
[WARNING]    org.codehaus.janino:janino:jar:3.1.12:compile
[WARNING]    org.glassfish.jersey.containers:jersey-container-jetty-http:jar:3.1.3:compile
[WARNING]    org.glassfish.jaxb:jaxb-runtime:jar:3.0.1:compile
[WARNING]    org.eclipse.jetty:jetty-webapp:jar:11.0.24:compile
[WARNING]    org.eclipse.jetty:jetty-plus:jar:11.0.24:compile
```
This does not solve jersey-common/hk2 usage (need to dig into that later) or findbugs (#1080)

Change to the classpath is:
```
jakarta.transaction-api-2.0.0.jar			      <
jetty-jndi-11.0.24.jar					      <
jetty-plus-11.0.24.jar					      <
jetty-webapp-11.0.24.jar				      <
jetty-xml-11.0.24.jar					      <
metrics-json-4.2.25.jar					      <
metrics-servlets-4.2.25.jar				      <
profiler-1.1.1.jar					      <
```
Removed some unreferenced jetty libraries and their deps (transaction-api) as well as some dropwizard deps (metrics) and their deps (profiler)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -448,6 +448,10 @@
       <artifactId>metrics-jvm</artifactId>
     </dependency>
     <dependency>
+      <groupId>jakarta.annotation</groupId>
+      <artifactId>jakarta.annotation-api</artifactId>
+    </dependency>
+    <dependency>
       <groupId>jakarta.ws.rs</groupId>
       <artifactId>jakarta.ws.rs-api</artifactId>
     </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -66,6 +66,7 @@
     <dep.httpcore.version>5.2.1</dep.httpcore.version>
     <dep.icu4j.version>73.2</dep.icu4j.version>
     <dep.jackson.version>2.15.2</dep.jackson.version>
+    <dep.jakarta.annotation-api.version>2.1.1</dep.jakarta.annotation-api.version>
     <dep.jakarta.ws.rs-api.version>3.1.0</dep.jakarta.ws.rs-api.version>
     <dep.jakarta.xml.bind-api.version>3.0.1</dep.jakarta.xml.bind-api.version>
     <dep.janino.version>3.1.12</dep.janino.version>
@@ -197,6 +198,11 @@
         <groupId>info.picocli</groupId>
         <artifactId>picocli</artifactId>
         <version>${dep.picocli.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>jakarta.annotation</groupId>
+        <artifactId>jakarta.annotation-api</artifactId>
+        <version>${dep.jakarta.annotation-api.version}</version>
       </dependency>
       <dependency>
         <groupId>jakarta.ws.rs</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -66,18 +66,22 @@
     <dep.httpcore.version>5.2.1</dep.httpcore.version>
     <dep.icu4j.version>73.2</dep.icu4j.version>
     <dep.jackson.version>2.15.2</dep.jackson.version>
+    <dep.jakarta.ws.rs-api.version>3.1.0</dep.jakarta.ws.rs-api.version>
     <dep.jakarta.xml.bind-api.version>3.0.1</dep.jakarta.xml.bind-api.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -551,6 +614,21 @@
...
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-params</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.platform</groupId>
+      <artifactId>junit-platform-commons</artifactId>
+      <scope>test</scope>
+    </dependency>
```
sambish5, line 631: For some of these (I'm only explicitly calling out these JUnit ones because I know we've switched to using the BOM), I don't know if we explicitly want to add them into the pom or use one the additional configs that can be added to the `maven-dependency-plugin`.
https://maven.apache.org/plugins/maven-dependency-plugin/examples/exclude-dependencies-from-dependency-analysis.html
Using something like `<ignoredUsedUndeclaredDependencies>` and setting these in there may be good since they are all included under the umbrella of the BOM, and _technically_ then aren't declared but used.

I'm not 100% if these other config tabs would be accurate to use for others of theses dependencies or not, but wanted to at lease throw this idea out.

jpdahlke: I like where you are going with that, and I agree we might want to clean this up a bit. I figured we would do this step first and then we can follow-up with configuring the tool better?

sambish5 (approved):

pom.xml
```diff
@@ -198,6 +199,11 @@
         <artifactId>picocli</artifactId>
         <version>${dep.picocli.version}</version>
       </dependency>
+      <dependency>
```
fbruton, line 202: This dependency also needs to be included in the `dependencies` section of the pom.

fbruton (changes requested):

sambish5 (approved):

PR #1085, opened 2025-03-07
refresh :: add monitoring code to detect config changes

dev-mlb:
This mechanism, similar to the Apache Commons FileAlterationMonitor, triggers registered FileAlterationObservers when files are changed. However, it operates on-demand, only checking for alterations when explicitly invoked at specified intervals and when the place is invoked, as opposed to the continuous monitoring employed by the FileAlterationMonitor.

Polling the file system for changes at predetermined intervals, this mechanism does not depend on the operating system's native file change notification events. This on-demand approach can be more resource-efficient, especially when file changes are infrequent, and can lead to reduced overhead and potentially better performance due to the absence of continuous monitoring and event handling.

Important Note: This on-demand mechanism may introduce a slight delay in detecting file changes, dependent on the chosen polling interval and when the place is invoked, compared to continuous monitoring solutions.

### DIFF
--- a/src/main/java/emissary/place/ServiceProviderRefreshablePlace.java
+++ b/src/main/java/emissary/place/ServiceProviderRefreshablePlace.java
@@ -220,8 +220,10 @@ public abstract class ServiceProviderRefreshablePlace extends ServiceProviderPla
             Preconditions.checkArgument(StringUtils.isNotBlank(path), "Path cannot be blank");
             Preconditions.checkArgument(intervalMinutes > 0, "Monitoring interval is not greater than 0");
 
-            final FileAlterationObserver observer;
             final Path file = Paths.get(path);
+            Preconditions.checkArgument(Files.exists(file), "Path does not exist");
+
+            final FileAlterationObserver observer;
             if (Files.isDirectory(file)) {
                 logger.debug("Monitoring directory {} for changes", file);
                 observer = new FileAlterationObserver(path);
--- a/src/main/java/emissary/place/ServiceProviderRefreshablePlace.java
+++ b/src/main/java/emissary/place/ServiceProviderRefreshablePlace.java
@@ -4,15 +4,25 @@ import emissary.config.ConfigUtil;
 import emissary.config.Configurator;
 import emissary.core.Factory;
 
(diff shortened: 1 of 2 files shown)


src/main/java/emissary/place/ServiceProviderRefreshablePlace.java
```diff
@@ -78,21 +90,49 @@ public ServiceProviderRefreshablePlace(final ServiceProviderRefreshablePlace pla
...
     }
 
+    @Override
+    protected void setupPlace(@Nullable final String theDir, final String placeLocation, final boolean register) throws IOException {
+        super.setupPlace(theDir, placeLocation, register);
+
+        try {
+            Preconditions.checkNotNull(this.configG, "The configurator is null");
+            final var path = configG.findStringEntry("MONITORING_PATH");
+            if (StringUtils.isNotBlank(path)) {
```
drivenflywheel, line 100: Should we also add a `Files.exists(file)` check here or in the `Monitor` constructor?

drivenflywheel (commented):
Verified that I could get RefreshablePlace to reload new config values if I had it configured to monitor its own .cfg file, but I was also able to demonstrate that we don't currently log a warning if we specify a non-existent path, and that silent misconfiguration might be hard to troubleshoot

drivenflywheel (approved):

src/main/java/emissary/place/ServiceProviderRefreshablePlace.java
```diff
@@ -78,21 +90,49 @@ public ServiceProviderRefreshablePlace(final ServiceProviderRefreshablePlace pla
...
+                final var intervalMinutes = configG.findLongEntry("MONITORING_INTERVAL_MINUTES", 15);
+                this.monitor = new Monitor(this, path, intervalMinutes);
```
jpdahlke, line 102: I think to get started, having an info log message here would be nice so we know it wired correctly. Once we convert things over we can change to debug.

jpdahlke: I'm good with a follow-on for this addition.

jpdahlke (approved):

PR #1088, opened 2025-03-18
regression :: updated IbdoMethodNames to use BiConsumers

dev-mlb:
The current implementation relies on reflection to access IBDO methods. This approach, however, is being replaced with direct method calls to improve robustness and maintainability. Reflection, while flexible, introduces fragility due to its reliance on string-based method lookups. These string representations can easily break during refactoring, leading to unexpected errors and runtime failures. By shifting to direct method calls, we eliminate this risk and ensure that method invocations remain intact even as the codebase evolves. This transition not only enhances the reliability of the system but also improves its overall performance by bypassing the overhead associated with reflection.


cfkoehler (approved):
This is my first time seeing BiConsumer's used and think it all makes sense. Behavior is as I expect.

sambish5 (approved):
nice :fire: think switching to direct calls will be helpful in the long-term

PR #1095, opened 2025-04-02
pmd :: UnnecessaryFullyQualifiedNames

sambish5:
Commented out UnnecessaryFullyQualified rule.
- in very few cases, actually removing the fully qualified name, only if it seemed useless/Assertions in testing classes.


jpdahlke:
The recommendation for this was to disable the rule in the PMD config file.

jpdahlke (approved):

PR #1101, opened 2025-04-04
Output 0-length files in MultiFileUnixCommandPlace

nixon124:
Added configurable ability to output 0-length files in MultiFileUnixCommandPlace


src/main/java/emissary/place/MultiFileUnixCommandPlace.java
```diff
@@ -285,7 +288,7 @@ protected void getFileList(File tmpDir, String inputFileName, List<File> outFile
                     logger.debug("Ignoring file '{}' because it is the input file.", fname);
                     continue;
                 }
-                if (file.length() == 0) {
+                if (file.length() == 0 && ignoreEmptyFile) {
```
fbruton, line 291: Skip unnecessary file length check if the feature flag is false.
```suggestion
                if (ignoreEmptyFile && file.length() == 0) {
```

fbruton (changes requested):

fbruton (approved):

PR #1106, opened 2025-04-08
removing obsolete myKey value from SessionProducer constructor

sambish5:
this was noticed through the pmd UnusedFormalParameter warnings, as `myKey` was not was not used from the method signature with all uses in the class being commented out.

### DIFF
--- a/src/main/java/emissary/parser/SessionProducer.java
+++ b/src/main/java/emissary/parser/SessionProducer.java
@@ -19,7 +19,6 @@ public class SessionProducer {
     public static final String PARSER_ERROR = "This session could not be completely parsed. The format was bad.";
 
     protected SessionParser sp;
-    // protected String myKey;
     protected int numSessions = 0;
     protected List<String> initialForms;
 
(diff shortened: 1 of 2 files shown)


src/main/java/emissary/parser/SessionProducer.java
```diff
@@ -29,12 +28,10 @@ public class SessionProducer {
      * Creates a SessionProducer
      *
      * @param sp The SessionParser, which has parsed theContent.
-     * @param myKey Value to stick into the transform history, obsolete
      * @param initialForms Forms to be preloaded onto the form stack.
      */
-    public SessionProducer(SessionParser sp, String myKey, List<String> initialForms) {
```
dev-mlb, line 35: We probably should deprecate this constructor first, as it will have downstream implications, but still move to the new constructor.

dev-mlb (changes requested):

src/main/java/emissary/parser/SessionProducer.java
sambish5: good idea, you're right should probably do that. I'll implement it

dev-mlb (approved):

drivenflywheel (approved):

PR #1107, opened 2025-04-09
[DRAFT] extraction-test :: add support for answer schemas

dev-mlb:
This initiative proposes the implementation of schema validation for extraction test answer files utilizing XSD version 1.0, necessitated by the limited availability of libraries supporting version 1.1. Consequently, certain workarounds, such as explicit enumeration of attachments and extracts, are required due to constraints inherent in version 1.0. It is imperative that sequence items adhere to their defined order, which may pose challenges for integration. In cases where validation parsing errors occur, it is suggested that instead of failing the test, a warning or error be logged. This alternative approach would facilitate the incremental resolution of all warnings and enable the activation of XSD validation in a future major release. The immediate failure of the test was executed to assess the potential integration impact. Input and recommendations are welcome. The primary aim is to streamline the validation code through this process.


src/test/resources/emissary/test/core/schemas/answers.xsd
```diff
@@ -0,0 +1,604 @@
...
+            <xs:element ref="numAttachments"/>
+            <xs:element ref="extractCount"/>
+            <xs:group ref="children"/>
+        </xs:choice>
+        <xs:attribute ref="os-release"/>
+        <xs:attribute ref="os-version"/>
+    </xs:complexType>
+    <xs:group name="children">
+        <xs:sequence>
+            <xs:element name="att1" type="child" minOccurs="0"/>
```
jpdahlke, line 202: Could we add another element here named `att` that's a different type that could be validated independently much like the current child; that way we can have a `maxOccurs="unbounded"` and eventually move to that strategy over time?

dev-mlb: Added the ability to specify the elements as:
```
<att index="1">
<extract index="1">
```
Updated ExtractionTest class to support.

jpdahlke (approved):

src/test/resources/emissary/test/core/schemas/answers.xsd
```diff
@@ -0,0 +1,586 @@
...
+            <xs:attribute ref="os-version"/>
+            <xs:attribute ref="matchMode"/>
+            <xs:attribute ref="collectionSeparator"/>
+        </xs:complexType>
+    </xs:element>
+    <xs:element name="nometa">
+        <xs:complexType>
+            <xs:sequence>
+                <xs:element ref="name"/>
+                <xs:element ref="value"/>
```
drivenflywheel, line 97: Should we have a `value` child for `nometa`?  Could change in a follow-on PR if we don't want it

drivenflywheel (approved):

PR #1109, opened 2025-04-09
code clean-up in MagicNumberFactory

sambish5:
noticed this class could use some TLC while going through pmd stuff.
Generally made simple changes. There are a lot of `null` checks in this class on things that explicitly cannot be null, so further clean-up could be done, but I left _most_ of those for now.


jpdahlke (approved):

drivenflywheel (approved):
Thanks for cleaning this up

PR #1137, opened 2025-05-05
tests :: add schema for regression test answers

dev-mlb:
Introducing a regression test answer schema to provide a basis for comparison with the extraction test answer schema.

### DIFF
--- a/src/test/java/emissary/test/core/junit5/UnitTest.java
+++ b/src/test/java/emissary/test/core/junit5/UnitTest.java
@@ -100,6 +100,10 @@ public abstract class UnitTest {
         assertMaxNonSystemThreadCount(1);
     }
 
+    public String getAnswerXsd() {
+        return answerXsd;
+    }
+
     /**
      * Configure the test stuff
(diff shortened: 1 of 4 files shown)


src/test/java/emissary/test/core/junit5/UnitTest.java
```diff
@@ -267,6 +271,7 @@ protected Document getAnswerDocumentFor(String resource) {
     @Nullable
     protected Document getAnswerDocumentValidated(String aname) {
         SchemaFactory schemaFactory = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI);
+        String answerXsd = getAnswerXsd();
```
fbruton, line 274: The local variable `answerXsd` has the same name as the field. Suggest calling `getAnswerXsd()` multiple times in this method or calling the field `answerXsd`

fbruton (changes requested):

drivenflywheel (approved):

PR #1145, opened 2025-06-07
bug fix :: initialize FileAlterationObserver in ServiceProviderRefreshablePlace

dev-mlb:
Found this while testing #1144


src/main/java/emissary/place/ServiceProviderRefreshablePlace.java
```diff
@@ -235,9 +235,14 @@ protected Monitor(final ServiceProviderRefreshablePlace place, final String path
                 observer = new FileAlterationObserver(parent.toFile(), new NameFileFilter(fileName));
             }
 
-            final var listener = new RefreshListener(place);
-            observer.addListener(listener);
-            this.observers.add(observer);
+            try {
+                final var listener = new RefreshListener(place);
+                observer.addListener(listener);
+                observer.initialize();
```
dev-mlb, line 241: I was missing this initialize method, so all fired events were seen as creation events.

jpdahlke (approved):

PR #1149, opened 2025-06-09
New abstract place for gRPC connections

rupert-griffin:
Adds a new place that handles gRPC connections.

Reviewers should pay particular attention to the pom.xml changes. It took a lot to get everything to install without triggering errorprone warnings and the like. The proto test file was especially difficult to work with.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1254,7 +1254,7 @@
               <phase>generate-sources</phase>
               <configuration>
                 <sources>
-                  <source>${project.build.directory}/target/generated-sources/protobuf</source>
+                  <source>${project.build.directory}/generated-sources/protobuf</source>
                 </sources>
               </configuration>
             </execution>
--- a/src/main/java/emissary/grpc/GrpcConnectionPlace.java
+++ b/src/main/java/emissary/grpc/GrpcConnectionPlace.java
@@ -133,19 +133,18 @@ public abstract class GrpcConnectionPlace extends ServiceProviderPlace implement
      * @param callLogic function that performs the actual gRPC call using the stub and request
      * @param request the protobuf request message to send
      * @return the response returned by the gRPC call
-     * @param <StubT> the gRPC stub type
-     * @param <ReqT> the protobuf request type
-     * @param <RespT> the protobuf response type
+     * @param <Q> the protobuf request type
+     * @param <R> the protobuf response type
+     * @param <S> the gRPC stub type
      */
-    protected <StubT extends AbstractBlockingStub<StubT>, ReqT extends GeneratedMessageV3, RespT extends GeneratedMessageV3> RespT invokeGrpc(
-            Function<ManagedChannel, StubT> stubFactory,
-            BiFunction<StubT, ReqT, RespT> callLogic, ReqT request) {
+    protected <Q extends GeneratedMessageV3, R extends GeneratedMessageV3, S extends AbstractBlockingStub<S>> R invokeGrpc(
+            Function<ManagedChannel, S> stubFactory, BiFunction<S, Q, R> callLogic, Q request) {
 
         return retryHandler.execute(() -> {
             ManagedChannel channel = ConnectionFactory.acquireChannel(channelPool);
-            RespT response = null;
+            R response = null;
             try {
-                StubT stub = stubFactory.apply(channel);
+                S stub = stubFactory.apply(channel);
                 response = callLogic.apply(stub, request);
                 ConnectionFactory.returnChannel(channel, channelPool);
             } catch (StatusRuntimeException e) {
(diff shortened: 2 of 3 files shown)


src/main/java/emissary/grpc/RetryHandler.java
```diff
@@ -0,0 +1,95 @@
...
+                .build());
+
+        retry.getEventPublisher()
+                .onRetry(this::logMessageOnRetry)
+                .onError(this::logMessageOnError);
+    }
+
+    private void logMessageOnRetry(RetryOnRetryEvent event) {
+        int attemptNumber = event.getNumberOfRetryAttempts();
+        Level level = attemptNumber <= numFailsBeforeWarn ? Level.INFO : Level.WARN;
```
cfkoehler, line 84: I like this way of doing the logging! Great work!

pom.xml
cfkoehler, line 1: @jpdahlke Can you check the pom changes.
Here is what I got for dependency tree differences:

## New Dependencies Added:

### Compile Dependencies:
•  com.google.protobuf:protobuf-java:jar:3.25.5:compile
•  io.github.resilience4j:resilience4j-core:jar:1.7.0:compile
•  With transitive dependency: io.vavr:vavr:jar:0.10.2:compile
◦  And its transitive: io.vavr:vavr-match:jar:0.10.2:compile
•  io.github.resilience4j:resilience4j-retry:jar:1.7.0:compile
•  io.grpc:grpc-stub:jar:1.45.1:compile

### Test Dependencies:
•  io.grpc:grpc-protobuf:jar:1.45.1:test
•  With transitive dependency: com.google.api.grpc:proto-google-common-protos:jar:2.26.0:test
•  And: io.grpc:grpc-protobuf-lite:jar:1.45.1:test
•  javax.annotation:javax.annotation-api:jar:1.3.2:test

rupert-griffin:
> @rupert-griffin Are these build messages any concern?
> 
> ```
> [INFO] --- protobuf:0.6.1:compile (default) @ emissary ---
> [INFO] /Users/runner/work/emissary/emissary/src/main/proto does not exist. Review the configuration or consider disabling the plugin.
> ```

> src/main/proto does not exist. Review the configuration or consider disabling the plugin

I'll need to confirm first, but I'm thinking this is due to having all the Protobuf dependencies in the pom.xml while only having Protobuf files in the test directory and not in the src directory. Seems like a point in favor of just creating an arbitrary gRPC subclass in src for testing purposes vs instantiating one in the GrpcConnectionPlaceTest file itself and doing everything there.

src/main/java/emissary/grpc/GrpcConnectionPlace.java
```diff
@@ -0,0 +1,250 @@
...
+            }
+            return response;
+        });
+    }
+
+    /**
+     * Executes multiple unary gRPC calls in parallel using a shared {@link AbstractFutureStub}.
+     * <p>
+     * TODO: Determine channel handling strategy when some calls succeed and others fail <br>
+     * TODO: Clarify expected blocking behavior for response collection
```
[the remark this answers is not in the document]
rupert-griffin: I figured this could be optional future work. Not essential, but I think future stubs can be nice to have.

DEVELOPING.md
```diff
@@ -791,3 +791,12 @@ nslookup somehost
 ```
 
 If you have this situation, then it is recommended to use [Google's DNS servers](https://developers.google.com/speed/public-dns/docs/using).
+
+### Generated Sources
+
+Intellij sometimes doesn't see code generated from other files (e.g. Protobuf).
+To resolve this, right-click:
+* `/target/generated-sources/${subfolder}` > Mark Directory as > Sources Root
```
fbruton, line 799: By adding the generated gRPC source to the sources, this patch eliminates the need for additional IntelliJ instructions.
[0001-add-generated-grpc-code-to-sources.patch.txt](https://github.com/user-attachments/files/20746029/0001-add-generated-grpc-code-to-sources.patch.txt)

fbruton (changes requested):

DEVELOPING.md
rupert-griffin: @fbruton Okay, checks are passing again

src/main/java/emissary/grpc/sample/DataReversePlace.java
cfkoehler, line 1: Thank you for this example place. Good idea on the reverse functionality

cfkoehler (approved):

jpdahlke (dismissed):
Only reviewed the pom which was close to what I expected.

cfkoehler (approved):

cfkoehler (approved):

PR #1159, opened 2025-07-05
add murmur hash 'HASH_ID' to KffDataObjectHandler

cfkoehler:
close [3487](https://github.com/NationalSecurityAgency/burrito-grande/issues/3487)

### DIFF
--- a/src/main/java/emissary/kff/KffDataObjectHandler.java
+++ b/src/main/java/emissary/kff/KffDataObjectHandler.java
@@ -199,10 +199,10 @@ public class KffDataObjectHandler {
      * 
      * @param d the data object
      * @param useSbc use the {@link SeekableByteChannel} interface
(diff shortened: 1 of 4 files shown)


src/main/java/emissary/kff/KffDataObjectHandler.java
drivenflywheel, line 1: (Globally) Replace `murmer` with `murmur`

cfkoehler: :face_palm:

cfkoehler:
Need to think through what happens if the payload changes and hashing get's called again. Don't think we want duplicate of this field in the output

cfkoehler:
Close in favor of: https://github.com/NationalSecurityAgency/emissary/pull/1163

PR #1171, opened 2025-08-18
New CoordinationPlace configs to deny places or put them at end of ref list

rcameronm:
All `SERVICE_COORDINATION` parameters defined in a coordination place's flavored config file will end up at the beginning of the reference list. Added two new parameters for when this behavior isn't wanted:
1. `APPEND_SERVICE_COORDINATION`: Appends a place to the *end* of the reference list
2. `DENY_SERVICE_COORDINATION`: Remove a place from the reference list (e.g. it was defined in the base config file but you don't want it in a given flavor)

These changes haven't been discussed anywhere prior to this PR as far as I'm aware. It just seemed like it would be useful. These code changes only took a few minutes so if we decide we don't wanna move forward with this it's not a big deal.

### DIFF
--- a/src/main/java/emissary/place/CoordinationPlace.java
+++ b/src/main/java/emissary/place/CoordinationPlace.java
@@ -108,6 +108,8 @@ public class CoordinationPlace extends ServiceProviderPlace {
         updateTransformHistory = configG.findBooleanEntry("UPDATE_TRANSFORM_HISTORY", false);
 
         placeKeys = configG.findEntries("SERVICE_COORDINATION");
+        placeKeys.addAll(configG.findEntries("APPEND_SERVICE_COORDINATION"));
+        placeKeys.removeAll(configG.findEntries("DENY_SERVICE_COORDINATION"));
         logger.debug("We got {} entries to coordinate", placeKeys.size());
 
         placeRefs = new ArrayList<>();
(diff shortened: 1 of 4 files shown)


cfkoehler (dismissed):
I threw a bunch of different cases's in the flavor cfg file and the coordination place got loaded in the order that I would expect. So think this is good. But @drivenflywheel might know more gotchas to check

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -108,6 +108,8 @@ protected void configurePlace() {
         updateTransformHistory = configG.findBooleanEntry("UPDATE_TRANSFORM_HISTORY", false);
 
         placeKeys = configG.findEntries("SERVICE_COORDINATION");
+        placeKeys.addAll(configG.findEntries("APPEND_SERVICE_COORDINATION"));
+        placeKeys.removeAll(configG.findEntries("DENY_SERVICE_COORDINATION"));
```
drivenflywheel, line 112: As implemented, "DENY_SERVICE_COORDINATION" is final;  we can't override it in another flavor file.  Is that what we want?

Are we not implementing a "REMOVE_SERVICE_COORDINATION" option since we could just use `SERVICE_COORDINATION != "PLACE_TO_DENY"?

drivenflywheel (changes requested):

drivenflywheel (approved):

cfkoehler (approved):

PR #1172, opened 2025-08-20
adding OS specific checks to attachment elements in ExtractionTest

sambish5:
adding ability to have OS specific `<numAttachments>` & `<att#>` in test answer files.

### DIFF
--- a/src/test/java/emissary/test/core/junit5/ExtractionTest.java
+++ b/src/test/java/emissary/test/core/junit5/ExtractionTest.java
@@ -382,28 +382,40 @@ public abstract class ExtractionTest extends UnitTest {
                 osSpecificNumAtt = true;
             }
         }
+
+        // check attachments answer file count against payload count
         if (!osSpecificNumAtt) {
             numAtt = JDOMUtil.getChildIntValue(el, "numAttachments");
             numAttElements = el.getChildren().stream().filter(c -> c.getName().startsWith(ATTACHMENT_ELEMENT_PREFIX)).count();
-        }
 
-        // check attachments answer file count against payload count
-        if (numAtt > -1) {
-            if (osSpecificNumAtt && attachments != null && !attachments.isEmpty()) {
-                assertTrue(numAtt <= attachments.size(), String.format(
-                        "Expected <numAttachments> in %s for specific OS not less than or equal to number of att in payload. ==> expected: <%d> but was: <%d>",
-                        tname, numAtt, attachments.size()));
-            } else {
+            if (numAtt > -1) {
                 assertEquals(numAtt, attachments != null ? attachments.size() : 0,
                         String.format("Expected <numAttachments> in %s not equal to number of att in payload.", tname));
+            } else if (numAtt == -1 && numAttElements > 0) {
+                assertEquals(numAttElements, attachments != null ? attachments.size() : 0,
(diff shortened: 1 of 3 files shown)


src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -368,19 +368,54 @@ protected void checkAnswers(Document answers, IBaseDataObject payload, List<IBas
     protected void checkAnswers(Element el, IBaseDataObject payload, @Nullable List<IBaseDataObject> attachments, String tname)
             throws DataConversionException {
 
-        int numAtt = JDOMUtil.getChildIntValue(el, "numAttachments");
```
dev-mlb, line 371: I believe we might be overcomplicating this slightly. The logic that verifies the attachment number against the provided attachments could remain as is. Our main adjustment would involve altering how we extract answers from the XML. Maybe if we did something like:
```
        int numAtt = -1;
        long numAttElements = -1;
        List<Element> numAttachments = el.getChildren("numAttachments");
        for (Element numAttEl : numAttachments) {
            if (verifyOs(numAttEl)) {
                numAtt = Integer.parseInt(numAttEl.getValue());
                numAttElements = el.getChildren().stream().filter(c -> c.getName().startsWith(ATTACHMENT_ELEMENT_PREFIX) && verifyOs(c)).count();
                break;
            }
        }
```
Thoughts?

dev-mlb (changes requested):

src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -368,19 +368,53 @@ protected void checkAnswers(Document answers, IBaseDataObject payload, List<IBas
...
+                    fail(String.format("%d attachments in payload with no count in answer xml, add matching <numAttachments> count for %s",
+                            attachments.size(), tname));
+                }
+            }
         } else {
-            if (attachments != null && !attachments.isEmpty()) {
-                fail(String.format("%d attachments in payload with no count in answer xml, add matching <numAttachments> count for %s",
-                        attachments.size(), tname));
+            int attInPayload = attachments != null ? attachments.size() : 0;
+            assertTrue(numAtt <= attInPayload, String.format(
```
dev-mlb, line 404: Should this check be equals?

src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -368,19 +368,53 @@ protected void checkAnswers(Document answers, IBaseDataObject payload, List<IBas
...
+                    "Expected <numAttachments> in %s for specific OS not less than or equal to number of att in payload. ==> expected: <%d> but was: <%d>",
+                    tname, numAtt, attInPayload));
+            assertTrue(numAttElements <= attInPayload, String.format(
```
dev-mlb, line 407: Should this check be equals?

dev-mlb (changes requested):
I'm thinking we do not need error messages for "specific OS" test failures. Maybe that could condense the test statements??

src/test/java/emissary/test/core/junit5/ExtractionTest.java
sambish5: removed and now just compares att# and numAttachments counts with the old comparisons

cfkoehler (approved):

PR #1190, opened 2025-09-26
change version command output to include git hash

memny:
output changes from
`Emissary Version: 8.33.0-SNAPSHOT 2025-09-26T14:22:15Z`
to
`Emissary Version: 8.33.0-SNAPSHOT - built on 2025-09-26T14:34:44Z - git hash: de96edc`

### DIFF
--- a/src/main/java/emissary/Emissary.java
+++ b/src/main/java/emissary/Emissary.java
@@ -146,7 +146,9 @@ public class Emissary {
     }
 
     protected void dumpVersionInfo() {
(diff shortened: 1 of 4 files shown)


cfkoehler:
@memny do you think there is any benefit in expanding the version details on the endpoint as well? 
https://github.com/NationalSecurityAgency/emissary/blob/main/src/main/java/emissary/server/api/Version.java

memny:
yes, totally agree. i'll update that as well.  for background I was having trouble figuring out exactly what was running on dev.

PR #1196, opened 2025-10-10
Added ConfigUtil functionality to instantiate new objects from configuration file

rcameronm:
This is a helper for some of my future work. I don't believe this functionality already exists in Emissary, but let me know if it does and I can take this down.


drewfarris (approved):

cfkoehler (approved):
Integrated upstreams and simplifies a test. Looks good to me!

PR #1290, opened 2026-01-12
executrix :: validate placename with an allowlist

dev-mlb:


memny (changes requested):
Any thoughts on adding a `getPlaceName()` and some tests?

memny (approved):

fbruton (approved):

PR #1296, opened 2026-01-25
Fix Issue #997: Preserve ConfigEntry order during merge operations

### LINKED ISSUE #997: ServiceConfigGuide re-orders ConfigEntries when merging
Given a configuration file like:
```
TEST_KEY = testValueA
TEST_KEY = testValueB
```

There are potentially places that are sensitive to the ordering of the ConfigEntries that are generated, such as coordination places. When running with a plain configuration file, the order is preserved, but when merging in a flavored configuration (or whenever else the merge flag is set to true in `ServiceConfigGuide.handleNewEntry(...)`, the new ConfigEntries generated are inserted at position 0 in the `serviceParameters` list. As a result, the entries wind up in the list in the opposite order in which they were listed in the configuration file. This means it may be difficult to preserve/predict the order a series of configuration entries are returned by something like `List<String> configG.findEntries("TEST_KEY")`

In the case of the example above, a non-merge operation generates the order 'testValueA, testValueB' whereas a merged configuration operation produces the result of 'testValueB, testValueA'. The issue on merge may indeed be more complex than a simple re-ordering and deserves further investigation.

One portion of the code in question is from the `ServiceConfigGuide.handleNewEntry(...)` method https://github.com/NationalSecurityAgency/emissary/blob/03c58c0a79727249407146c134fd897c58ddc1bd/src/main/java/emissary/config/ServiceConfigGuide.java#L289-L293

Alternately, we could just make the statement that `findEntries` always returns the matching configuration options in non-deterministic order and rely on other methods for precisely ordering configuration entries.

It's not clear how pervasive the ordering assumption is throughout configuration files used in systems that depend on Emissary, so any changes related to this will require investigation.

ravisastryk:
The `merge()` method was reversing entry order by inserting at position 0. Fixed by collecting entries first and using addAll(0, collection) which preserves iteration order.

Fixes #997 - ConfigEntry order is now preserved during merge operations.

## Background and Context

When merging configurations, entries were inserted at position 0 individually. This was causing the final order to be reversed from the source file order.

```properties
# Config file
KEY = valueA
KEY = valueB
KEY = valueC
```

- Direct load: `[valueA, valueB, valueC]`
- After merge: `[valueC, valueB, valueA]` 

### Why This Works

`List.addAll(int index, Collection c)` inserts **all elements in iteration order** starting at the specified index. The existing elements shift right to accommodate.

```
Before: serviceParameters = [X, Y, Z]
addAll(0, [A, B, C])
After:  serviceParameters = [A, B, C, X, Y, Z]  <----- Order is preserved
```

### Why This Is Better

| Aspect | `add(0, x)` in loop | `addAll(0, collection)` |
|--------|---------------------|-------------------------|
| **Correctness** | Reverses order | Preserves order |
| **Performance** | O(n²) - shifts array n times | O(n) - single shift operation |
| **Atomicity** | Multiple mutations | Single list operation |

## Fix

Replace the loop of `add(0, entry)` calls with a single `addAll(0, entries)`  which preserves iteration order while still placing merged entries at the front.

## Changes

- `ServiceConfigGuide.merge()`: Use `addAll(0, ...)` instead of individual insertions
- Added regression tests for entry ordering

## Tests

- All existing tests pass
- New tests verify order preservation on merge

### DIFF
--- a/src/main/java/emissary/config/ServiceConfigGuide.java
+++ b/src/main/java/emissary/config/ServiceConfigGuide.java
@@ -1192,19 +1192,43 @@ public class ServiceConfigGuide implements Configurator, Serializable {
      */
     @Override
     public void merge(final Configurator other) throws IOException {
-        int i = 1;
-
         // First handle the remove entries from "other"
         if (other instanceof ServiceConfigGuide) {
             for (final ConfigEntry entry : ((ServiceConfigGuide) other).getRemoveEntries()) {
-                handleNewEntry(entry.getKey(), entry.getValue(), "!=", "<merge>", i++, true);
+                // Process remove entries with the != operator
+                final String parmName = handleReplacements(entry.getKey(), "<merge>", 0);
+                final String sval = handleReplacements(entry.getValue(), "<merge>", 0);
+                final ConfigEntry anEntry = new ConfigEntry(parmName, sval);
+
+                if ("*".equals(sval)) {
+                    removeAllEntries(parmName);
+                    this.values.remove(parmName);
+                } else {
+                    removeEntry(anEntry);
+                    if (sval.equals(this.values.get(parmName))) {
+                        this.values.remove(parmName);
+                    }
+                }
+                this.removeParameters.add(anEntry);
             }
         }
 
-        // Add in new entries from "other" at the top of the list
+        // Collect all entries to merge first, then insert in batch
+        // This preserves the order of entries from the source configuration
+        // Using addAll(0, collection) inserts all elements at position 0 in iteration order
+        final List<ConfigEntry> entriesToMerge = new ArrayList<>();
         for (final ConfigEntry entry : other.getEntries()) {
-            handleNewEntry(entry.getKey(), entry.getValue(), "=", "<merge>", i++, true);
+            final String parmName = handleReplacements(entry.getKey(), "<merge>", 0);
+            final String sval = handleReplacements(entry.getValue(), "<merge>", 0);
+            final ConfigEntry anEntry = new ConfigEntry(parmName, sval);
+            entriesToMerge.add(anEntry);
+
+            // Update the values map with the last value for substitution
+            this.values.put(parmName, sval);
         }
+
+        // Insert all merged entries at position 0, preserving their relative order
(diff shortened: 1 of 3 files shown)


drivenflywheel (dismissed):
Code and tests look good.  Will have to evaluate downstream impacts prior to downstream integration in case any existing configs deliberately leverage the existing behavior.  Those downstream configs could be reworked, but they'll need to be identified.

ravisastryk:
> Code and tests look good. Will have to evaluate downstream impacts prior to downstream integration in case any existing configs deliberately leverage the existing behavior. Those downstream configs could be reworked, but they'll need to be identified.

That's a good point. However, if downstream systems are affected, they were **relying on buggy behavior** and need fixing anyway and we never documented this behavior and it is actually fixing the original bug https://github.com/NationalSecurityAgency/emissary/issues/997

ravisastryk:
@rupert-griffin @jpdahlke @drivenflywheel Can you please re-review?

ravisastryk:
> @ravisastryk - I understand that you're eager to get this merged, but it's not likely to happen in the near term. This PR changes heavily-leveraged behavior that's been in place for years - and in a good way - but once this is merged downstream integrators won't be able to accept any other emissary changes unless they also ensure that this change won't impact their runtime behavior. That can be a heavy lift, and we don't want this change to impede any urgent needs that may come up.

Thank you for the thoughtful explanation @drivenflywheel I appreciate the transparency about the downstream concerns and the care taken to protect integrators from unexpected disruptions. That said, I'd love to keep this conversation going so we can find a path forward when the timing is right. A few thoughts that might help frame that discussion:

1. **Backward-compatible bridging**: @rupert-griffin's suggestion of overloading `handleNewEntry` and `merge` with a prepend/append parameter is a great middle ground — it keeps existing behavior as the default while making the correct behavior opt-in. Happy to implement that if it would lower the integration risk.

2. **Feature flag / config knob**: If it helps, the behavior change could be gated behind a config flag so integrators can adopt it on their own timeline.

No pressure on timing. I just want to make sure the fix for #997 stays on the radar and that there's a clear path to eventually landing it. Happy to take direction on whichever approach the team prefers.

PR #1301, opened 2026-01-26
sentinel :: log stack trace and/or thread dump

dev-mlb:
closes: https://github.com/NationalSecurityAgency/burrito-grande/issues/3752

### DIFF
--- a/src/main/config/logback.xml
+++ b/src/main/config/logback.xml
@@ -21,6 +21,26 @@
     </encoder>
   </appender>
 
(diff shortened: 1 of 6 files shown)


drivenflywheel (approved):
Excellent work.   We may want to consider a future enhancement that captures info only for the specific thread(s) responsible for triggering the action so the log output doesn't include "innocent bystanders", but this is already very useful as-implemented.

cfkoehler (approved):
This was cool to see in action. Excited for how it can help us! 
Thanks for the fast work!

PR #1306, opened 2026-01-29
sentinel :: run actions on affected threads

dev-mlb:


memny (commented):
tests pass and the code looks good to me but I don't think I'm familiar enough with what it's doing for my opinion to matter.

fbruton (approved):

drivenflywheel (approved):
Sorry for the delayed review.  This is working exactly as I'd want.  Excellent work 🔥

PR #1350, opened 2026-04-10
Increased gRPC class modularization and updated exception handling

rcameronm:


src/main/java/emissary/grpc/invoker/GrpcInvoker.java
```diff
@@ -123,28 +96,17 @@ public <Q extends GeneratedMessageV3, R extends GeneratedMessageV3, S extends Ab
...
-                    if (cause instanceof RuntimeException) {
-                        throw (RuntimeException) cause;
-                    }
-                    throw new IllegalStateException(cause);
-                })
-                .exceptionally(throwable -> {
-                    Throwable cause = unwrapThrowable(throwable);
-                    logger.error(cause.getMessage(), cause);
-                    return null; // Return null instead of throwing exception and ruining entire batch
+                    throw GrpcExceptionUtils.toContextualRuntimeException(
```
cfkoehler: Just want to confirm the desired behavior now is to "fail fast" on the first exception rather than logging-and-returning-null for failed futures. The new behavior is arguably more correct (silent null returns are dangerous), but any downstream callers that relied on the old "swallow and continue" semantics will break silently.

rcameronm: Yes since futures aren't used downstream yet I think this change should be safe to make.

Also, since it's possible for whatever reason that we may want something to swallow and continue, we can pass in that behavior to the `exceptionally` argument in the `CompletableFutureFinalizers#awaitAllAndGet(Collection, Supplier, Function)` method

drivenflywheel (approved):

cfkoehler (approved):

PR #1355, opened 2026-04-16
Add unlimited retry support for gRPC calls

cfkoehler:
Downstream had this feature to block processing gRPC is not responsive. Match the same capability here. 

This pull request adds support for an "unlimited retry" mode to the `RetryHandler`, allowing gRPC calls to be retried indefinitely until they succeed, rather than failing after a maximum number of attempts. It introduces a new configuration option, updates the retry logic for both synchronous and asynchronous calls, and adds comprehensive tests for the new behavior.

**Unlimited Retry Feature:**

* Added a new configuration property `GRPC_RETRY_UNLIMITED` to enable unlimited retry cycles after exhausting the configured maximum attempts. When enabled, the retry cycle restarts indefinitely until the call succeeds. [[1]](diffhunk://#diff-6793610d295357d2736a5a4b64c64a48f66a95801cfae07880ccae2a98e11eedR39-R41) [[2]](diffhunk://#diff-6793610d295357d2736a5a4b64c64a48f66a95801cfae07880ccae2a98e11eedR68-R76) [[3]](diffhunk://#diff-6793610d295357d2736a5a4b64c64a48f66a95801cfae07880ccae2a98e11eedR91)
* Updated the `execute` and `executeAsync` methods in `RetryHandler` to support unlimited retry cycles, including custom logic for both synchronous and asynchronous retries.
* Added helper methods for unwrapping exceptions and handling retry cycles in asynchronous execution.

**Testing:**

* Added a new nested test class `UnlimitedRetryTests` in `GrpcSampleServicePlaceTest.java` to verify the unlimited retry behavior, including tests for both recoverable and non-recoverable gRPC error codes.

**Dependency and Import Updates:**

* Added imports for `CompletableFuture` and `CompletionException` to support the new async retry logic.

### DIFF
--- a/src/main/java/emissary/grpc/retry/RetryHandler.java
+++ b/src/main/java/emissary/grpc/retry/RetryHandler.java
@@ -13,6 +13,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.event.Level;
 
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -34,6 +36,9 @@ import java.util.function.Supplier;
  * default={@code 2.0}</li>
  * <li>{@code GRPC_RETRY_NUM_FAILS_BEFORE_WARN} - Determines the number of execution failures before logger should start
  * sending warnings, default={@code 3}</li>
+ * <li>{@code GRPC_RETRY_UNLIMITED} - When {@code true}, the retry cycle restarts indefinitely after exhausting
+ * {@code GRPC_RETRY_MAX_ATTEMPTS} until the call succeeds. Use this when downstream systems must block until the gRPC
+ * request completes and a permanent failure is not acceptable, default={@code false}</li>
  * </ul>
  * To calculate exponential backoff wait time, let:
  * <ul>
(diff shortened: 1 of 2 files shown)


src/main/java/emissary/grpc/retry/RetryHandler.java
```diff
@@ -107,10 +115,52 @@ private void logMessageOnError(RetryOnErrorEvent event) {
     }
 
     public <T> T execute(Supplier<T> supplier) {
-        return Retry.decorateSupplier(retry, supplier).get();
+        if (!retryUnlimited) {
+            return Retry.decorateSupplier(retry, supplier).get();
+        }
+        int cycleCount = 0;
+        while (true) {
```
cfkoehler, line 122: I think this can be simpler. just first try at it

rupert-griffin:
I looked into it a bit, hoping there was something in Resilience4j that internally allowed for unlimited retries, but unfortunately that isn't the case.

I ran some numbers though, and if we just set the maximum retry attempts to `Integer.MAX_VALUE` and keep all the other settings as default, we can keep retrying continuously for years. 

The below result is in milliseconds, where `2.147 * 10^12` milliseconds is roughly 68 years.

retry

* `N` is the total retries (`Integer.MAX_VALUE`, or `1<<31`)
* `M` is the retry back off mulitiplier (2)
* `w_0` is the initial wait time (64 milliseconds)
* `w` is the maximum wait time (1000 milliseconds)

PR #1360, opened 2026-04-21
extraction-test :: add support for strict (regression) mode

dev-mlb:


src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -122,8 +126,57 @@ public abstract class ExtractionTest extends UnitTest {
...
+        STRICT, LENIENT
+    }
+
+    /**
+     * Get the test mode
+     *
+     * @return the test mode
+     */
+    public TestMode getTestMode() {
+        return TestMode.LENIENT;
```
drivenflywheel: I'd expect this to return the value of a private or protected field, not a constant value.  Is this the intended implementation?

src/test/java/emissary/test/core/junit5/RegressionTest.java
```diff
@@ -47,79 +22,12 @@
  * <li>Once the tests pass, you can remove the overridden method(s) added above.</li>
  * </ol>
  */
+@Deprecated
```
drivenflywheel: I'm tempted to defer this annotation until a follow-on PR, because this might cause an awful lot of warnings in downstream projects.

```suggestion
```

src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -122,8 +126,57 @@ public abstract class ExtractionTest extends UnitTest {
...
+     */
+    public boolean isLenient() {
+        return getTestMode() == TestMode.LENIENT;
+    }
+
+    /**
+     * Regression mode -- This test mode compares the entire BDO instead of just what is defined in the XML. In other words,
+     * the XML must define exactly the output of the Place, no more and no less. There are methods provided to generate the
+     * XML required.
+     *
```
drivenflywheel, line 133: (nit) It seems like this Javadoc comment should be on the `TestMode` enum or its `STRICT` value rather than on the `isStrict()` method.   We also may want to be consistent on the "Regression/STRICT" and "Extraction/LENIENT" terminology.  (I happen to prefer "Strict" and "Lenient" myself.)

drivenflywheel (changes requested):

src/test/java/emissary/test/core/junit5/RegressionTest.java
dev-mlb: removed

src/test/java/emissary/test/core/junit5/ExtractionTest.java
dev-mlb: Removed TestMode, simplified it down to a simple isStrict method. No reason to over complicate things.

drivenflywheel (approved):
🚀

cfkoehler (approved):

PR #1363, opened 2026-04-24
remove use of getStringParameter() in DropOffUtil

cfkoehler:
Tested with next upstream project but not validated all the way through yet. This change makes me nervous and can make it even smaller if we think best.


sambish5 (approved):

drivenflywheel (approved):
LGTM.  If we see a lot of warnings at runtime from `getParameterAsString`, it's a reflection of the data, not these changes.

PR #1366, opened 2026-04-28
add support for dataFile element in XML configuration and update test resource handling

cfkoehler:


src/test/java/emissary/test/core/junit5/UnitTest.java
cfkoehler, line 165: Do we think this should be depreciated? Or we can move the logic from `getMyXmlTestParameterFiles()` into this so no downstream API changes

dev-mlb: I think we should not deprecate it yet. I think the way RegressionTest works is it still needs to find the dat files first. My plan is to update the logic for discovery for RegressionTest (i.e. removing the initial form from the filename), but until then I think we need to keep this. I believe with a little more work, we'll be able to deprecate this method in the near future.

dev-mlb:
We should probably take this first, over #1372. Once we merge, I can rebase and account for the new datFile tag (I'll need to add it to generateAnswers logic).

dev-mlb (approved):

memny (approved):

PR #1368, opened 2026-04-30
Make gRPC futures thread interruptible

rcameronm:
`CompletableFuture#get()` and `CompletableFuture#join()` both have the same functionality. Both block the calling thread while waiting for a future to complete and then return the result. The difference between them is that:
* `CompletableFuture#get()`
  * Throws checked and unchecked exceptions and needs to be wrapped in a `try` block
  * Throws an `InterruptedException` if the thread is interrupted and stops waiting for the future to complete
* `CompletableFuture#join()`
  * Only throws unchecked exceptions
  * Ignores interrupted threads

Switching the finalizer method to `get` because if a place times out, we want the thread interrupt to cancel the RPC. Note that sequential RPCs using gRPC blocking stubs are already thread interruptible.

### DIFF
--- a/src/main/java/emissary/grpc/future/CompletableFutureFinalizers.java
+++ b/src/main/java/emissary/grpc/future/CompletableFutureFinalizers.java
@@ -2,11 +2,13 @@ package emissary.grpc.future;
 
 import emissary.grpc.exceptions.GrpcExceptionUtils;
 
+import io.grpc.Status;
 import jakarta.annotation.Nullable;
 
 import java.util.Collection;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
@@ -17,8 +19,9 @@ public class CompletableFutureFinalizers {
     /**
      * Waits for a provided asynchronous computation to complete and then return its result. This method blocks the calling
      * thread until the {@link CompletableFuture} input has completed. If the future completes exceptionally, the
-     * corresponding {@link CompletableFuture#join()} call will throw a {@link RuntimeException}, unless the future has been
-     * preconfigured with explicit exception handling behavior.
+     * corresponding {@link CompletableFuture#get()} call will throw a {@link RuntimeException}, unless the future has been
+     * preconfigured with explicit exception handling behavior. If the thread is interrupted while the future is still
+     * completing, the call will be canceled and a {@link io.grpc.StatusRuntimeException} will be thrown.
      *
      * @param future future representing an asynchronous computation
      * @return the result returned by an asynchronous computation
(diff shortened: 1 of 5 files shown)


src/main/java/emissary/grpc/future/CompletableFutureFinalizers.java
```diff
@@ -17,17 +19,22 @@ private CompletableFutureFinalizers() {}
...
      * @return the result returned by an asynchronous computation
      * @param <R> result type
      */
     public static <R> R awaitAndGet(CompletableFuture<R> future) {
         try {
-            return future.join();
-        } catch (RuntimeException e) {
+            return future.get();
+        } catch (InterruptedException e) {
+            future.cancel(false);
```
cfkoehler, line 37: Per the Java spec, `CompletableFuture` doesn't use a managed thread so it can't be interrupted via this flag. The call marks the future exceptionally with `CancellationException` (if it hasn't completed), which prevents any other code from consuming its result, but the underlying gRPC call on the network/executor thread is unaffected. 

I think having this here is good and harmless but we just might want to clarify in the doc above:
> Cancel the client-side future; the in-flight RPC may still complete on the server

src/test/java/emissary/grpc/sample/GrpcSampleServicePlaceTest.java
```diff
@@ -426,6 +476,62 @@ void testBlockedResponsesAreProcessedSequentially() throws IOException, Interrup
...
+                    }
+                });
+
+                clientThread.start();
+
+                assertTrue(startedLatch.await(2, TimeUnit.SECONDS), "Not all servers received requests");
+                assertNull(errorRef.get());
+
+                clientThread.interrupt();
+                clientThread.join(2000);
```
cfkoehler: Is `2000`ms too long for a test here? Do you think `500` would be enough?

```suggestion
                clientThread.join(500);
```

rcameronm: Good point

src/main/java/emissary/grpc/future/CompletableFutureFinalizers.java
rcameronm: This is a good point, you're right. The original `ListenableFuture` object returned from the gRPC invocation sends a cancellation message back to the server when you call `future.cancel(true)`, but `CompletableFuture` doesn't.

Even though my main goal with this PR was to just not block Burrito if the place times out, this is still a pretty big oversight on my part. 

Tomorrow I'll push an update that will send a cancellation message to the server when we cancel in the client. It's up to the server how it wants to handle that, but at least then it isn't any of Burrito's business anymore.

rcameronm: New test changes should now indicate that if a gRPC server is prepared to handle a canceled RPC, it will handle them upon client-side thread interrupt.

drivenflywheel (approved):

cfkoehler (approved):

PR #1372, opened 2026-05-03
extraction-test :: generate answers

dev-mlb:
Alternate pr #1387, if we take #1366

Notes:
- Retains setup blocks in the resulting answer files.
- OS-specific variations are not currently handled.
- Targeted Updates Only: -DgenerateAnswers=true requires specific test class overrides rather than a bulk update.
- This iteration streamlines roughly 99% of the generation process. However, some manual "finishing touches" may be required, such as adjusting for OS differences or stripping non-idempotent (volatile) fields.

To test, you can update this method in ExtractionTest to look like:
```
        return /*isStrict() &&*/ Boolean.getBoolean("generateAnswers") && createAnswerGenerator() != null;
    }
```

And then run `mvn clean package -DgenerateAnswers=true`

Or, in a specific extraction test add:
```
        return true;
    }
```


src/test/java/emissary/core/IBaseDataObjectXmlCodecs.java
```diff
@@ -692,6 +714,18 @@ public void encode(final List<Map<String, byte[]>> values, final Element parentE
             DEFAULT_STRING_ENCODER,
             DEFAULT_STRING_OBJECT_ENCODER);
 
+    /**
+     * The default set of XML element encoders.
```
drivenflywheel: `ALT_DEFAULT_ELEMENT_ENCODERS` should probably have a different description that `DEFAULT_ELEMENT_ENCODERS`.   I can't distinguish why to use one vs the other.

src/test/java/emissary/test/core/junit5/ExtractionTestAnswerGenerator.java
```diff
@@ -0,0 +1,95 @@
...
+import org.jdom2.Document;
+import org.jdom2.Element;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class ExtractionTestAnswerGenerator extends RegressionTestAnswerGenerator {
```
drivenflywheel: (this role reversal is cosmetically weird, given that RegressionTest extends ExtactionTest)

dev-mlb: RegressionTestAnswerGenerator was a fallback generator, but the code doesn't make it that far because the answer file is required to exist.

drivenflywheel (dismissed):
🚀

PR #1373, opened 2026-05-04
update use of deprecated ChecksumCalculator constructor in testing classes

sambish5:
update from deprecated constructor that takes `String[] algs` to the updated constructor that accepts `Collection<String> algs`


dev-mlb (approved):

drivenflywheel (approved):
🚀

PR #1374, opened 2026-05-04
update use of deprecated Factory.create() in FactoryTest

sambish5:
move from 
`
create(final String className, final Object[] args, final String location)
`
to
`
create(final String className, final List<Object> args, final String location)
`


drivenflywheel (approved):
🚀

cfkoehler (approved):

PR #1385, opened 2026-05-12
Stop killing gRPC channels after RPC failure

rcameronm:
`ManagedChannel` objects are expensive to create and are meant to be long-lived and reused. RPC failure typically indicates an issue on the server and not with the channel itself. If it is due to an underlying connection issue, the `ManagedChannel` handles it under the hood. 

See:
https://grpc.io/docs/guides/performance/
https://github.com/grpc/grpc-java/issues/7442
https://github.com/grpc/grpc-java/issues/11020

### DIFF
--- a/src/main/java/emissary/grpc/invoker/GrpcInvoker.java
+++ b/src/main/java/emissary/grpc/invoker/GrpcInvoker.java
@@ -47,12 +47,11 @@ public class GrpcInvoker {
             ManagedChannel channel = ConnectionFactory.acquireChannel(channelPool);
             try {
                 S stub = stubFactory.apply(channel);
-                R response = callLogic.apply(stub, request);
-                ConnectionFactory.returnChannel(channel, channelPool);
-                return response;
+                return callLogic.apply(stub, request);
             } catch (RuntimeException e) {
-                ConnectionFactory.invalidateChannel(channel, channelPool);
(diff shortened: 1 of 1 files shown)


fbruton (approved):

src/main/java/emissary/grpc/invoker/GrpcInvoker.java
```diff
@@ -47,12 +47,11 @@ public <Q extends Message, R extends Message, S extends AbstractBlockingStub<S>>
             ManagedChannel channel = ConnectionFactory.acquireChannel(channelPool);
             try {
                 S stub = stubFactory.apply(channel);
-                R response = callLogic.apply(stub, request);
-                ConnectionFactory.returnChannel(channel, channelPool);
-                return response;
+                return callLogic.apply(stub, request);
             } catch (RuntimeException e) {
```
memny, line 51: is there any case where the runtime exception would indicate that the channel has gone bad and that it should be invalidated??

rcameronm: Not as far as I can tell. Everything I read says that a managed channel should be alive for the app's entire lifecycle. I haven't found anything to indicate that we should be invalidating it at any point other than at app shutdown.

memny (approved):

drewfarris (approved):

PR #1386, opened 2026-05-12
httpclient5 :: migrate HTTP connection handling to HttpClient 5.6

sambish5:
**Upgrading all deprecated HttpClient/HttpCore 5 pieces to use the new TLS/SSL protocol pieces.**
_Some important details for this PR:_  
  
`Registry<ConnectionSocketFactory>` has been mainly replaced by `setTlsSocketStrategy`.
`setTlsSocketStrategy` method in the builder automatically applies your custom TLS/SSL strategy to the https protocol handler. The `PoolingHttpClientConnectionManagerBuilder` creates a default registry that includes standard "http" and "https" support.
  
Was getting errors originally with these updates in `EmissaryServerIT.testSSLWorks()`. This test was failing at line 47 for `client.send(new HttpGet(endpoint)).getContent(MapResponseEntity.class)`, and the error was:
```
Bad request -> status: 500 message: class javax.net.ssl.SSLHandshakeException: No name matching localhost found
```
This required me to add the `HostnameVerificationPolicy` to the `setTlsSocketStrategy`. The different strategies for this are as follows:
- BUILTIN: Delegated to the JSSE provider, typically executed during the TLS handshake.
- CLIENT: Executed by the HTTP client after the TLS handshake.
- BOTH: verifies both above policies to ensure the server is authorized for that hostname.
  
In order for it to work by default, or without `HostnameVerificationPolicy.CLIENT` set, we'd have to add `localhost` to the keystore.jks. This is something we can do if we feel is necessary, but for now felt setting it to the CLIENT policy works fine.

`SSLContext.getInstance()` and `init()` is recommended to be replaced by `SSLContexts.custom().build()`.
This is also applicable to `HttpClientBuilder.create()` to `HttpClients.custom().build()`, so I implemented both of these where applicable (_though not necessarily complained about or marked as deprecated_).

### DIFF
--- a/src/main/java/emissary/client/HTTPConnectionFactory.java
+++ b/src/main/java/emissary/client/HTTPConnectionFactory.java
@@ -15,6 +15,7 @@ import org.apache.hc.client5.http.ssl.DefaultClientTlsStrategy;
 import org.apache.hc.client5.http.ssl.DefaultHostnameVerifier;
 import org.apache.hc.client5.http.ssl.HostnameVerificationPolicy;
 import org.apache.hc.client5.http.ssl.NoopHostnameVerifier;
+import org.apache.hc.client5.http.ssl.TrustAllStrategy;
 import org.apache.hc.core5.http.ConnectionReuseStrategy;
 import org.apache.hc.core5.ssl.SSLContexts;
 import org.apache.log4j.Logger;
@@ -133,7 +134,7 @@ public class HTTPConnectionFactory {
 
         // custom ssl context populated with our trust and key materials
         return SSLContexts.custom()
-                .loadTrustMaterial(trustStore, (chain, authType) -> true)
+                .loadTrustMaterial(trustStore, TrustAllStrategy.INSTANCE)
                 .loadKeyMaterial(keyStore, kpChar)
                 .setProtocol(cfg.findStringEntry(CFG_SSLCONTEXT_TYPE, DFLT_CONTEXT_TYPE))
                 .setSecureRandom(new SecureRandom())
--- a/src/main/java/emissary/client/HTTPConnectionFactory.java
+++ b/src/main/java/emissary/client/HTTPConnectionFactory.java
@@ -8,22 +8,22 @@ import com.google.common.annotations.VisibleForTesting;
 import jakarta.annotation.Nullable;
 import org.apache.hc.client5.http.impl.DefaultClientConnectionReuseStrategy;
 import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
-import org.apache.hc.client5.http.impl.classic.HttpClientBuilder;
+import org.apache.hc.client5.http.impl.classic.HttpClients;
 import org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager;
-import org.apache.hc.client5.http.socket.ConnectionSocketFactory;
-import org.apache.hc.client5.http.socket.PlainConnectionSocketFactory;
+import org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManagerBuilder;
+import org.apache.hc.client5.http.ssl.DefaultClientTlsStrategy;
 import org.apache.hc.client5.http.ssl.DefaultHostnameVerifier;
+import org.apache.hc.client5.http.ssl.HostnameVerificationPolicy;
 import org.apache.hc.client5.http.ssl.NoopHostnameVerifier;
-import org.apache.hc.client5.http.ssl.SSLConnectionSocketFactory;
 import org.apache.hc.core5.http.ConnectionReuseStrategy;
-import org.apache.hc.core5.http.config.Registry;
-import org.apache.hc.core5.http.config.RegistryBuilder;
+import org.apache.hc.core5.ssl.SSLContexts;
 import org.apache.log4j.Logger;
 
 import java.io.IOException;
 import java.security.GeneralSecurityException;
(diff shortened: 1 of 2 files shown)


src/main/java/emissary/client/HTTPConnectionFactory.java
dev-mlb: Since you're using the `.loadTrustMaterial` method below, I believe you can get rid of this line now?

From what I'm reading

> The SSLContexts.custom().loadTrustMaterial(...) method acts as a high-level wrapper. Internally, it handles the creation of the TrustManagerFactory, initializes it with your KeyStore, and extracts the TrustManager for you.
> 
> Manually initializing the TrustManagerFactory and KeyManagerFactory was the standard "Java way" for years, but it requires several lines of boilerplate and extra exception handling. The HttpClient 5 builder reduces this to a single method call

src/main/java/emissary/client/HTTPConnectionFactory.java
```diff
@@ -145,10 +137,12 @@ SSLContext build(final Configurator cfg) throws IOException, GeneralSecurityExce
         final KeyManagerFactory kmf = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
```
dev-mlb: Same as TrustManagerFactory, I believe you can get rid of this line.

src/main/java/emissary/client/HTTPConnectionFactory.java
```diff
@@ -145,10 +137,12 @@ SSLContext build(final Configurator cfg) throws IOException, GeneralSecurityExce
...
         kmf.init(keyStore, kpChar);
 
-        final SSLContext sc = SSLContext.getInstance(cfg.findStringEntry(CFG_SSLCONTEXT_TYPE, DFLT_CONTEXT_TYPE));
-        sc.init(kmf.getKeyManagers(), tmf.getTrustManagers(), new SecureRandom());
-
-        return sc;
+        // custom ssl context populated with our trust and key materials
+        return SSLContexts.custom()
```
dev-mlb, line 136: Do we need to add this to the builder?
```.setProtocol(cfg.findStringEntry(CFG_SSLCONTEXT_TYPE, DFLT_CONTEXT_TYPE))```

Not sure if we need it or not, but it was defined above.

File: src/main/java/emissary/client/HTTPConnectionFactory.java [hunk 3]
```diff
@@ -102,24 +97,21 @@ private HTTPConnectionFactory() {
...
         } catch (IOException | GeneralSecurityException ex) {
             log.error("Error configuring HTTPConnectionFactory. The connection factory will use HTTP Client default settings", ex);
         }
-        if (registry == null) {
-            this.connMan = new PoolingHttpClientConnectionManager();
-        } else {
-            this.connMan = new PoolingHttpClientConnectionManager(registry);
-        }
-
+        this.connMan = Objects.requireNonNullElseGet(connectionManager, PoolingHttpClientConnectionManager::new);
```
dev-mlb:
```suggestion
       this.connMan = Objects.requireNonNullElseGet(connectionManager, () -> PoolingHttpClientConnectionManagerBuilder.create().build());
```
Delegate to the builder maybe?

---

Or move this up into the catch block and simplify it to 
```
this.connMan = PoolingHttpClientConnectionManagerBuilder.create().build();
```
and reduce the need for `PoolingHttpClientConnectionManager connectionManager = null;`

dev-mlb (changes requested):
Overall looks good. Just some questions...

src/main/java/emissary/client/HTTPConnectionFactory.java
sambish5: I read that as well, but saw a bunch of examples online keeping the ManagerFactories. I agree, I think removing them is the correct call.

src/main/java/emissary/client/HTTPConnectionFactory.java
sambish5: Like and applied delegating it to the builder. Moving it into the catch created java complaints for me of "Defer assignment to 'connMan' using temp variable" due to them saying the variable may have already been assigned.

dev-mlb (dismissed):

src/main/java/emissary/client/HTTPConnectionFactory.java
```diff
@@ -139,16 +131,13 @@ SSLContext build(final Configurator cfg) throws IOException, GeneralSecurityExce
...
-        final KeyManagerFactory kmf = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
-        kmf.init(keyStore, kpChar);
-
-        final SSLContext sc = SSLContext.getInstance(cfg.findStringEntry(CFG_SSLCONTEXT_TYPE, DFLT_CONTEXT_TYPE));
-        sc.init(kmf.getKeyManagers(), tmf.getTrustManagers(), new SecureRandom());
-
-        return sc;
+        // custom ssl context populated with our trust and key materials
+        return SSLContexts.custom()
+                .loadTrustMaterial(trustStore, (chain, authType) -> true)
```
memny: I think it's cleaner to use the implementation that comes packaged with httpclient.

```suggestion
                .loadTrustMaterial(trustStore, TrustAllStrategy.INSTANCE)
```

memny (changes requested):

memny (approved):

dev-mlb (approved):

PR #1387, opened 2026-05-14
extraction-tests :: generate answers (adds dataFile logic)

dev-mlb:
- [x] Builds on #1366 

~~Take this or #1372~~


drivenflywheel (dismissed):

src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -163,17 +169,21 @@ public String getAnswerXsd() {
      *         automatically
      */
     protected boolean generateAnswers() {
-        return Boolean.getBoolean("generateAnswers") && createAnswerGenerator() != null;
+        return isStrict() && Boolean.getBoolean("generateAnswers") && createAnswerGenerator() != null;
```
fbruton, line 172: `isStrict()` defaults to false, causing `generateAnswers()` to return always return false for extraction tests.

fbruton (changes requested):

src/test/java/emissary/test/core/junit5/ExtractionTest.java
dev-mlb: That's correct. I wanted it to be off for ExtractionTests so it doesn't interfere with RegressionTest answer generation. Eventually I'll update it to remove the isStrict() check. For now, you'll need to update the specific extraction test that you want to generate answers.

src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -890,46 +943,40 @@ protected void setupPayload(IBaseDataObject payload, Document doc) {
         }
 
         Element root = (doc != null) ? doc.getRootElement() : null;
-        if (root == null) {
+        Element setup = (root != null) ? root.getChild(SETUP) : null;
+        if (setup == null) {
```
cfkoehler: With this, any test whose answer file has a <root> but no <setup> will now have a stale fileType. right? since it used to call `payload.setFileType(payload.currentForm());` Is that what we want?

dev-mlb: fixed

cfkoehler (approved):

fbruton (approved):

PR #1388, opened 2026-05-14
removing getStringParameter() usage across emissary

sambish5:
removing use of `getStringParameter()` across Emissary.
This pretty much only touches testing code, besides changes in `PayloadUtil`.

Places where `getStringParameter()` was not removed:
- HtmlEscapePlace & test
  - both of these classes are fully deprecated as is, felt no need to go in and change.
- BaseDataObjectTest, KffDataObjectHandlerTest, DropOffUtilTest, & MultiFileUnixCommandPlaceTest.
  - Each one of these test against `getStringParameter()`, `getParameterAsConcatString()`, & `getParameterAsString()`. I don't know if we want to maintain tests against deprecated code, but figured I wouldn't touch them for now. If it is desired for these to be removed, can easily be done in this PR or in the future.

### DIFF
--- a/src/main/java/emissary/util/PayloadUtil.java
+++ b/src/main/java/emissary/util/PayloadUtil.java
@@ -232,7 +232,7 @@ public class PayloadUtil {
         }
         final Element meta = new Element("metadata");
         for (final String key : d.getParameters().keySet()) {
-            final Element m = JDOMUtil.protectedElement("param", d.getParameterAsString(key));
+            final Element m = JDOMUtil.protectedElement("param", d.getStringParameter(key));
             m.setAttribute("name", key);
             meta.addContent(m);
         }
--- a/src/main/java/emissary/util/PayloadUtil.java
+++ b/src/main/java/emissary/util/PayloadUtil.java
@@ -232,7 +232,7 @@ public class PayloadUtil {
         }
         final Element meta = new Element("metadata");
         for (final String key : d.getParameters().keySet()) {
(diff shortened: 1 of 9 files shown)


src/test/java/emissary/test/core/junit5/ExtractionTest.java
```diff
@@ -610,7 +610,7 @@ protected void checkAnswers(Element el, IBaseDataObject payload, @Nullable List<
             if (verifyOs(meta)) {
                 String key = meta.getChildTextTrim(NAME);
                 checkForMissingNameElement(PARAMETER, key, tname);
-                checkStringValue(meta, payload.getStringParameter(key), tname);
+                checkStringValue(meta, payload.getParameterAsString(key), tname);
```
jpdahlke: I think these calls could return multiple values based on where they are generically used. Should the underlying code here do something different rather than just do string comparisons?

src/main/java/emissary/util/PayloadUtil.java
```diff
@@ -232,7 +232,7 @@ public static Document toXml(final IBaseDataObject d) {
         }
         final Element meta = new Element("metadata");
         for (final String key : d.getParameters().keySet()) {
-            final Element m = JDOMUtil.protectedElement("param", d.getStringParameter(key));
+            final Element m = JDOMUtil.protectedElement("param", d.getParameterAsString(key));
```
jpdahlke: Is there a different way we should be doing this to handle multiple values?

memny (approved):

jpdahlke (approved):

PR #1396, opened 2026-05-22
feat: allow custom base resource path for emissary server

ianpgifford:
Changes:
- Added base resource path config and loader utility class
- Updated server context paths, action classes, and templates to reference base resource when set
- Modified link validator to allow relative links
  - Updated link validator tests 

To test:
- Change base resource path value ( i.e. `BASE_RESOURCE_PATH = "/mybaseresource"`) and verify the emissary server webapp operates under that resource, including static resources, shutdown, api, etc.


src/main/java/emissary/server/util/BaseResourcePathUtil.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.io.IOException;
+
+/**
+ * Helper for retrieving the base resource path for templates.
+ *
+ * The base resource path is loaded once on class initialization, and made available for the main server config and any
+ * action classes (and therefore templates) that need it.
+ */
+public final class BaseResourcePathUtil {
```
ianpgifford, line 18: I was unsure about the necessity of this class specifically; creating it let me have a config class for it which contains the base resource path config, rather than having to choose which of the others gets it in their config. I also wasn't sure about the overhead of re-loading the same config value many times, so this just puts it in static and solves that. Open to reworking this, though.

src/main/java/emissary/server/mvc/NavAction.java
```diff
@@ -44,7 +44,7 @@ public EmissaryNav nav() {
 
     public static class EmissaryNav {
 
-        private static final Pattern VALID_LINK = Pattern.compile("^(https?:/)?/.*");
+        private static final Pattern VALID_LINK = Pattern.compile("^(https?:/)?\\.?/.*");
```
ianpgifford, line 47: Needed relative links to keep the nav bar simple; technically relative/path works, but open relative paths make the regex harder to keep strict without getting overcomplicated, so I decided to enforce ./ notation for relative paths (./relative/path)

src/main/java/emissary/server/EmissaryServer.java
```diff
@@ -149,19 +150,21 @@ public Server startServer() {
             // Resource.setDefaultUseCaches(false);
 
             // needs to be loaded first into the server as it setups up Emissary stuff
+            String baseResourcePath = BaseResourcePathUtil.getBaseResourcePath();
+
             ContextHandler emissaryHandler = buildEmissaryHandler();
             // TODO: rework this, no need for it be set with a context path but if this
             // is left out, it matches / and nothing works correctly
-            emissaryHandler.setContextPath("/idontreallyservecontentnowdoi");
+            emissaryHandler.setContextPath(baseResourcePath + "/idontreallyservecontentnowdoi");
```
ianpgifford, line 162: Based on my research and testing how jetty handles context paths, there isn't a good way to "nest" paths with handlers, and this was the best way to add a parent path to everything. Still smells a little funny, though 🤨 would not be surprised to find out there was a better way, but I could not find it.

src/main/java/emissary/server/util/BaseResourcePathUtil.java
```diff
@@ -0,0 +1,34 @@
...
+    private static final String BASE_RESOURCE_PATH = initBaseResourcePath();
+
+    private BaseResourcePathUtil() {}
+
+    private static String initBaseResourcePath() {
+        try {
+            Configurator config = ConfigUtil.getConfigInfo(BaseResourcePathUtil.class);
+            return config.findStringEntry(CONFIG_KEY, DEFAULT_BASE_RESOURCE_PATH);
+        } catch (IOException e) {
+            throw new EmissaryRuntimeException("Unable to load configuration for BaseResourcePathUtil", e);
```
drivenflywheel: Since we're using this to optionally override a historical default value, in the case of a missing config file we should consider logging a message but then returning the default value.  
 (Would need to add a logger for this class)

```suggestion
            logger.info("Unable to load configuration for BaseResourcePathUtil; using default values", e);
            return DEFAULT_BASE_RESOURCE_PATH;
```

src/main/java/emissary/server/util/BaseResourcePathUtil.java
```diff
@@ -0,0 +1,34 @@
...
+    private static final String CONFIG_KEY = "BASE_RESOURCE_PATH";
+    private static final String DEFAULT_BASE_RESOURCE_PATH = "";
+    private static final String BASE_RESOURCE_PATH = initBaseResourcePath();
+
+    private BaseResourcePathUtil() {}
+
+    private static String initBaseResourcePath() {
+        try {
+            Configurator config = ConfigUtil.getConfigInfo(BaseResourcePathUtil.class);
+            return config.findStringEntry(CONFIG_KEY, DEFAULT_BASE_RESOURCE_PATH);
```
drivenflywheel: Rather than returning this directly, we may want to do some basic validation that the configured value is "safe" - i.e., it shouldn't facilitate directory traversal.

src/main/java/emissary/server/EmissaryServer.java
```diff
@@ -193,7 +197,7 @@ public Server startServer() {
             configuredServer.start();
             // server.join(); // don't join so we can shutdown
 
-            String serverLocation = cmd.getScheme() + "://" + cmd.getHost() + ":" + cmd.getPort();
+            String serverLocation = cmd.getScheme() + "://" + cmd.getHost() + ":" + cmd.getPort() + BaseResourcePathUtil.getBaseResourcePath();
```
drivenflywheel, line 204: Hmm.  Are we sure we need this added both here _and_ at the start of the context paths? (I haven't tested it.)

If we do need this here, we'll need to ensure that the first character after the port is "/" so we don't end up with something like "http://localhost:8080mycustompath"

drivenflywheel:
Good stuff.  Some minor feedback that may or may not need to be addressed

src/main/java/emissary/server/util/BaseResourcePathUtil.java
ianpgifford: Good point; thoughts on matching against the following regex?
`^(?:\/[\w-]+)*$`
Check for any number of "/ + characters" with valid characters in a-z, A-Z, 0-9, _, -
Good on `/route`, `/myroute/route`, `/my_route1/my-subroute2`
Fail on `route`, `./route`, `http://www.myroute.com`, `myroute!` 
Should be restrictive enough without preventing generally valid uses for a base path, assuming my regex is correct 😁

src/main/java/emissary/server/EmissaryServer.java
ianpgifford: The value created here is just displayed in the console when you start the server so you can click on it to open it in browser, etc., so this is just for convenience. Re: ensuring /, see other comment about regex, it should cover that.

ianpgifford:
There are comments I left about certain parts of the code which I resolved so as to not block the MR.

src/main/java/emissary/server/util/BaseResourcePathUtil.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.regex.Pattern;
+
+/**
+ * Helper for retrieving the base resource path for templates.
+ *
+ * The base resource path is loaded once on class initialization, and made available for the main server config and any
+ * action classes (and therefore templates) that need it.
+ */
+public final class BaseResourcePathUtil {
+    private static final Logger logger = LoggerFactory.getLogger(NavAction.class);
```
memny:
```suggestion
    private static final Logger logger = LoggerFactory.getLogger(BaseResourcePathUtil.class);
```

ianpgifford: Good catch, ty

src/main/java/emissary/server/util/BaseResourcePathUtil.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private static final Pattern VALID_BASE_RESOURCE_PATH = Pattern.compile("^(?:\\/[\\w-]+)*$");
```
memny: Adds the tilde since it's a valid URL character and addresses SonarQube complaints about redundant escape of the forward slash and infinite recursion (max of 5 repetitions is arbitrary).

```suggestion
    private static final Pattern VALID_BASE_RESOURCE_PATH = Pattern.compile("^(?:/[\w-~]+){1,5}$");
```

src/main/java/emissary/server/mvc/NavAction.java
memny: it allows some squirreliness like `http:/./blah` but I don't think it's that important.

how about

```suggestion
        private static final Pattern VALID_LINK = Pattern.compile("^(https?://.+|\\.{1,2}/.*");
```

src/main/java/emissary/server/util/BaseResourcePathUtil.java
ianpgifford: Tilde in URL makes me scared but according to the internet it is technically valid, so fair enough. Have to tweak the regex a little because /w-~ is considered a range according to the tester I'm using.

drivenflywheel (approved):

src/main/java/emissary/server/mvc/NavAction.java
memny: thoughts on this?

ianpgifford: Sorry, missed this as it isn't showing on the main comments list. You're missing a closing parenthesis in the regex; is this supposed to be `"^(https?:/)/.+|\\.{1,2}\.*"`? Also, I don't think we want to match on ../; there shouldn't be any case where links are navigating higher than the base route that I can think of. Frankly this regex should just be `"^\..*"` and live in a config validation test because it only serves to check against internally configured values anyway (there are no real security concerns around the safety or validity of this regex that I am aware of). The current one only accepts starting with http or https as well simply because it did before, and there's no reason to drop it in case there's some downstream use case where a nav link needs to go externally. I'm happy to just leave it as "starts with http(s):// or ./".

memny (approved):

PR #1404, opened 2026-05-28
add weekly schedule for GitHub Actions dependency updates

cfkoehler:


memny (dismissed):
This looks fine to me but I'm not versed in github actions. Would there be any interest in adding a [github actions linter](https://github.com/rhysd/actionlint)? I've experimented a little at home with precommit hooks I don't have a good feel for if that's going too far.

cfkoehler:
> This looks fine to me but I'm not versed in github actions. Would there be any interest in adding a [github actions linter](https://github.com/rhysd/actionlint)? I've experimented a little at home with precommit hooks I don't have a good feel for if that's going too far.

We run actionlint as part of mega linter: https://github.com/NationalSecurityAgency/emissary/blob/main/.github/linters/.mega-linter.yml#L8
See it flagging a different change here: https://github.com/NationalSecurityAgency/emissary/actions/runs/26566709985/job/78263130118

.github/dependabot.yml
```diff
@@ -15,3 +15,7 @@ updates:
       - dependency-name: "*"
         update-types:
           - "version-update:semver-major"
+  - package-ecosystem: "github-actions"
+    directory: "./.github/workflows"
```
dev-mlb: should this be `directory: "/"`?

dev-mlb (changes requested):

.github/dependabot.yml
dev-mlb: From what I'm reading:

If you use (directory: "./.github/workflows"), Dependabot will fail and throw an error similar to:

> "Dependabot can't resolve your directory configuration." Dependabot has strict, built-in rules for the github-actions ecosystem. It expects the directory parameter to be the root of the repository (/), because it already inherently knows exactly where GitHub Actions workflow files live (always in .github/workflows).

If you try to explicitly point it to ./.github/workflows, Dependabot evaluates that path relative to your repository root. It then looks for a path like .github/workflows/.github/workflows/, finds nothing, and stops working.

cfkoehler:
```suggestion
    directory: "/"
```

dev-mlb (approved):

jpdahlke (approved):

PR #1418, opened 2026-06-02
refactor: remove deprecated setFileTypeIfEmpty method

ianpgifford:
Changes:
- Removed deprecated setFileTypeIfEmpty overload and move its code into the remaining implementation of the method
- Remove tests for removed function
To test:
- Tests pass

### DIFF
--- a/src/main/java/emissary/core/BaseDataObject.java
+++ b/src/main/java/emissary/core/BaseDataObject.java
@@ -1136,18 +1136,12 @@ public class BaseDataObject implements Serializable, Cloneable, Remote, IBaseDat
     }
 
     @Override
-    @Deprecated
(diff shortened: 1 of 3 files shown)


ianpgifford:
I could not find any instances where the String[] overload of the function was in use, so no changes are necessary here or elsewhere.

cfkoehler:
I believe we were not planning to remove depreacted functions until the next major release. Is that right @jpdahlke

ianpgifford:
> I believe we were not planning to remove depreacted functions until the next major release. Is that right @jpdahlke

According to the ticket, this is not supposed to be removed until Emissary 9, but I figured the branch and/or PR can stand until then. Wasn't sure how long that would be, though. This was done before you had sent me the approved work chart so also feel free to close this if it's out of scope.

PR #1423, opened 2026-06-04
update GitHub Actions workflows to address lint findings

cfkoehler:
## Zizmor Security Fixes

Resolves 119 zizmor findings across 9 workflow files.

### Changes

**`template-injection` (44)** — Replaced `${{ ... }}` inside `run:` blocks with shell env vars. Actions interpolates `${{ }}` before the shell runs, so attacker-controllable values (branch names, `github.actor`, etc.) could inject shell code. Env vars deliver the value as data, not code.

**`artipacked` (14)** — Added `persist-credentials: false` to every `actions/checkout`. Stops `GITHUB_TOKEN` from being written into `.git/config`, where later steps or artifacts could leak it.

**`excessive-permissions` (17)** — Added workflow-level `permissions: contents: read` to 4 files. Jobs that need write still declare their own elevated `permissions:` block.

**`cache-poisoning` (2)** — Removed `cache: 'maven'` from `setup-java` on `macos-build` and `site-build`. These jobs run on both PRs and pushes to protected branches; a fork PR could poison a cache entry that a later main-branch push restores. Trade-off: slower builds on these two jobs.

**`archived-uses` (1)** — Replaced archived `actions/create-release@v1` with `gh release create`.

### Supporting changes

- `persist-credentials: false` breaks `git push`, so every push step now runs `gh auth setup-git` with `GH_TOKEN` in env — equivalent auth without writing the token to disk.
- Quoted bare `${VAR}` references the zizmor auto-fix introduced, to clear `shellcheck SC2086` warnings.

### Review focus

The four push workflows (`cherrypick`, `create-patch-branch`, `maven-release`, `maven-version`) are `workflow_dispatch`-only and aren't exercised by PR CI — worth a careful read of the push steps before the next release.

### Verification

- `zizmor` → No findings (41 still suppressed by config, unchanged)
- `actionlint` → clean

### DIFF
--- a/.github/workflows/maven-release.yml
+++ b/.github/workflows/maven-release.yml
@@ -25,6 +25,9 @@ env:
   JAVA_DISTRIBUTION: "corretto"
   MAVEN_OPTS: "-Dhttps.protocols=TLSv1.2 -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=WARN -Djava.awt.headless=true"
 
+permissions:
+  contents: read
+
 jobs:
   verify:
     runs-on: ubuntu-latest
@@ -73,6 +76,8 @@ jobs:
     steps:
       - name: Checkout repo
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - name: Get the version
         run: echo "RELEASE_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout | grep -v '\[.*' | awk -v release_suffix="$IN_RELEASE_SUFFIX" -F- '{print ""$1""release_suffix}')" >> "$GITHUB_ENV"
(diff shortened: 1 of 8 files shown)


.github/workflows/maven-release.yml
```diff
@@ -106,42 +115,40 @@ jobs:
...
         continue-on-error: true
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          tag_name: ${{ env.RELEASE_VERSION }}
-          release_name: ${{ env.RELEASE_VERSION }}
-          draft: false
-          prerelease: false
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: gh release create "${RELEASE_VERSION}" --title "${RELEASE_VERSION}" --target "release/${RELEASE_VERSION}"
```
cfkoehler, line 151: I still need to test/validate this fully

jpdahlke (approved):

dev-mlb (approved):

PR #1424, opened 2026-06-05
Validate gRPC endpoints at startup

rcameronm:


src/main/java/emissary/grpc/pool/ConnectionFactory.java
```diff
@@ -85,9 +89,9 @@ public class ConnectionFactory extends BasePooledObjectFactory<ManagedChannel> {
      * @param configG configuration provider for channel and pool parameters
      */
     public ConnectionFactory(String host, int port, Configurator configG) {
-        this.host = host;
-        this.port = port;
-        this.target = host + ":" + port; // target may be a host or dns service
+        this.host = validateHostName(host);
+        this.port = validatePortNumber(port);
+        this.target = createTarget(host, port);
```
cfkoehler: This is a nit and we can talk it through more. Could go either way on this.
This is creating the target with the unvalidated value. I know currently there is no normalization or changes done in `validateHostName` or `validatePortNumber` but could set us up in the future of making a mistake. 

so should this be:

```suggestion
        this.target = createTarget(this.host, this.port);
```

src/main/java/emissary/grpc/pool/ConnectionFactory.java
```diff
@@ -133,6 +137,26 @@ public ConnectionFactory(String host, int port, Configurator configG) {
         this.poolConfig.setLifo(retrievalOrdering.equals(PoolRetrievalOrdering.LIFO));
     }
 
+    private static String validateHostName(String host) {
+        if (host.equals(LOCALHOST) || host.startsWith(DNS_PREFIX)) {
```
cfkoehler: @jpdahlke how doe you feel on this narrow scope? Might be too opinionated  for the framework?

The allowed set — only "localhost" or "dns:///" prefix — is quite narrow. The gRPC ManagedChannelBuilder.forTarget() natively accepts:
- IPv4/IPv6 addresses ("10.0.0.1", "[::1]")
- Plain hostnames ("mygrpc-service", "myservice.svc.cluster.local")
- `passthrough:///` for direct addresses

src/main/java/emissary/grpc/pool/ConnectionFactory.java
```diff
@@ -133,6 +137,26 @@ public ConnectionFactory(String host, int port, Configurator configG) {
...
+            return host;
+        }
+        throw new IllegalArgumentException(
```
cfkoehler: There is no null guard here in `validateHostName`
If null is passed, host.equals(LOCALHOST) throws NullPointerException rather than IllegalArgumentException. Since the public constructor accepts any String, a null check with a clear message would be consistent with how the port validation behaves.

cfkoehler (commented):
Few comments. Nothing critical and happy to talk through them and be convinced they are not needed.

src/main/java/emissary/grpc/pool/ConnectionFactory.java
memny: I wouldn't expect a method called `validate...` to return something.

src/main/java/emissary/grpc/pool/ConnectionFactory.java
memny: I didn't realize intellij was [inferring @NotNull](https://blog.jetbrains.com/idea/2014/10/automatic-notnullnullablecontract-inference-in-intellij-idea-14/)

rcameronm: Ah good point, will address

src/main/java/emissary/grpc/pool/ConnectionFactory.java
rcameronm: Bad pattern to have them return? I can change both methods to void, then don't need to worry about which values are getting assigned

src/main/java/emissary/grpc/pool/ConnectionFactory.java
rcameronm: Plain hostnames are automatically inferred to be `dns:///`. My thought here was that implementing classes could define it differently. E.g. an `XdsChannelManager` will expect the `xds:///` prefix. But I agree that for a base case this may just be way too narrow.

rcameronm: Looking into it more, I think I should probably just remove the hostname validation stuff altogether. The managed channel builder checks if it's a valid URI or not.

cfkoehler (approved):

memny (approved):

PR #1400, opened 2026-05-26
New gRPC channel manager

rcameronm:
This pull request refactors the gRPC connection management system by replacing the old connection pooling approach with a new `ChannelManager` abstraction. The main goal is to provide a more flexible and configurable way to manage gRPC channels, making it easier to customize channel behavior and support different channel management strategies. The most significant changes are outlined below.

**Core Refactor: Channel Management Abstraction**

* Introduced a new abstract base class `ChannelManager` in `emissary.grpc.channel`, which centralizes configuration and management of gRPC `ManagedChannel` connections. This class supports various channel options (keep-alive, load balancing, message sizes) and provides a factory method for instantiating subclasses via reflection. (`src/main/java/emissary/grpc/channel/ChannelManager.java`)

* Updated `GrpcRoutingPlace` to use `ChannelManager` instead of the previous `ConnectionFactory` and pooling approach. Each target now has a dedicated `GrpcInvoker` backed by a `ChannelManager`, improving configurability and resource management. (`src/main/java/emissary/grpc/GrpcRoutingPlace.java`) [[1]](diffhunk://#diff-1415a03090ce56650769b0765593db672352d7c6ca0e5837c5e68b600e0dc278R4-L6) [[2]](diffhunk://#diff-1415a03090ce56650769b0765593db672352d7c6ca0e5837c5e68b600e0dc278L41-R42) [[3]](diffhunk://#diff-1415a03090ce56650769b0765593db672352d7c6ca0e5837c5e68b600e0dc278R54-R56) [[4]](diffhunk://#diff-1415a03090ce56650769b0765593db672352d7c6ca0e5837c5e68b600e0dc278L101-L122) [[5]](diffhunk://#diff-1415a03090ce56650769b0765593db672352d7c6ca0e5837c5e68b600e0dc278L191-R204)

**Configuration and API Updates**

* Added support for specifying the channel manager class via the `GRPC_CHANNEL_MANAGER_CLASS_NAME` configuration key, defaulting to `PooledChannelManager`. This allows for easy swapping or customization of channel management strategies. (`src/main/java/emissary/grpc/GrpcRoutingPlace.java`) [[1]](diffhunk://#diff-1415a03090ce56650769b0765593db672352d7c6ca0e5837c5e68b600e0dc278L41-R42) [[2]](diffhunk://#diff-1415a03090ce56650769b0765593db672352d7c6ca0e5837c5e68b600e0dc278R54-R56)

[...]

PR #1446, opened 2026-06-25
reorganized gRPC channel logic

rcameronm:
Preliminary work to move away from gRPC channel pools. The only expected behavioral change in this PR is that gRPC resources will be closed upon Emissary shutdown. No other changes should be apparent to users. 

See #1400 for planned follow-on work.


src/main/java/emissary/grpc/GrpcRoutingPlace.java
```diff
@@ -98,28 +93,27 @@ protected GrpcRoutingPlace(@Nullable Configurator configs) throws IOException {
     }
 
     private void configureGrpc() {
-        if (configG == null) {
-            throw new IllegalStateException("gRPC configurations not found for " + this.getPlaceName());
-        }
+        Objects.requireNonNull(configG);
```
cfkoehler, line 96: nit: losing context now with null config. now the error is just an NPE without the context. Any specific reason to remove the `IllegalStateException` and message?

rcameronm: No nothing important, I can revert this if you'd prefer. This line has just always bugged me. My understanding is that even though `configG` is technically nullable, it isn't actually possible for it to be null at this point. The null check was needed to remove a warning, but the throw line was practically unreachable.

Alternatively I think I can do 
```
Objects.requireNonNull(configG, "gRPC configurations not found for " + this.getPlaceName());
```

fbruton (approved):

cfkoehler (approved):

PR #1484, opened 2026-07-30
Magic number repeated bytes

ianpgifford:
Changes:
- Reworked existing magic byte pickup to enable repeated byte count format of `/(X)` ( followed by the byte to repeat )
- Added handler for \( escape character to initiate a repeated byte char

To test:
- Tests pass
- ...


dev-mlb:
In standard Unix `file` magic files that support the regex type, you can use standard regular expression quantifiers (like {n}, +, or *) to match repeating characters. So something like this:
```
0   regex   ^-{5,}    ASCII divider line
```
My suggestion is to align our magic code with the updated Unix file magic code and add regex support. 

Thoughts??

PR #1495, opened 2026-08-07
errorprone :: default charset utils

dev-mlb:


sambish5 (dismissed):

src/main/java/emissary/util/CharsetUtil.java
cfkoehler: Should the comment be updated since this is now `UTF_8` and not necessarily platform default?

cfkoehler: This is still `Latin-1` right? Could it cause problems having both methods be different?

dev-mlb: I think this is intentional...I updated the method signature (it's private) to point it out as a fallback method and updated the javadocs. Let me know if you think this works.

dev-mlb: Updated javadoc

sambish5 (approved):

cfkoehler (approved):

PR #1500, opened 2026-08-07
errorprone :: default charset output

dev-mlb:


src/main/java/emissary/output/roller/journal/JournalEntry.java
```diff
@@ -24,7 +26,7 @@ public JournalEntry(final String val, final long offset) {
     void serialize(final ByteBuffer b) {
         b.putInt(this.val.length());
         b.put(SEP);
-        b.put(this.val.getBytes());
+        b.put(this.val.getBytes(UTF_8));
```
cfkoehler: Think this might be a pre existing bug? Not sure if it is a problem though.

`putInt(val.length())` writes a char count, `put(val.getBytes(UTF_8))` writes bytes. Any non-ASCII value makes deserialize read short and throw on the separator check — journal unreadable, which defeats crash recovery.

dev-mlb: Yeah, this looks to be an existing bug. I updated the code here: #1514

cfkoehler (dismissed):

drivenflywheel (commented):
🚀

drivenflywheel (dismissed):

src/main/java/emissary/output/roller/journal/JournalEntry.java
```diff
@@ -27,7 +27,7 @@ void serialize(final ByteBuffer b) {
         final byte[] valBytes = this.val.getBytes(UTF_8);
         b.putInt(valBytes.length);
         b.put(SEP);
-        b.put(valBytes);
+        b.put(valBytes.getBytes(UTF_8));
```
jpdahlke:
```suggestion
        b.put(valBytes);
```

jpdahlke (approved):

cfkoehler (approved):

PR #1520, opened 2026-08-18
Extend gRPC channel manager with pooled channel manager

rcameronm:
This PR serves as another step towards using a single `ManagedChannel` per endpoint.

It splits all of the code from the `ChannelPoolFactory` into two places:
* `ManagedChannel` creation goes into the now abstract `ChannelManager` 
* Channel pooling goes into a dedicated `PooledChannelManager` subclass

There are no intended changes to Emissary outputs. 

Let me know if there's too much going on here and I can split this up into smaller PRs.


src/main/java/emissary/grpc/channel/ChannelManager.java
```diff
@@ -2,45 +2,204 @@
...
+ * <li>{@code GRPC_CHANNEL_MAX_INBOUND_MESSAGE_BYTE_SIZE} - Max inbound gRPC message size, default={@code 4194304}</li>
+ * <li>{@code GRPC_CHANNEL_MAX_INBOUND_METADATA_BYTE_SIZE} - Max inbound gRPC metadata size, default={@code 8192}</li>
+ * </ul>
  */
-public class ChannelManager implements AutoCloseable {
-    private final ObjectPool<ManagedChannel> channelPool;
-    private final String host;
-    private final int port;
+public abstract class ChannelManager implements AutoCloseable {
+    public static final String GRPC_CHANNEL_PREFIX = "GRPC_CHANNEL_";
```
cfkoehler: Will integration be needed adding this prefix? `GRPC_POOL_MAX_SIZE → GRPC_CHANNEL_POOL_MAX_SIZE`, etc.

src/main/java/emissary/grpc/channel/impl/PooledChannelManager.java
```diff
@@ -0,0 +1,245 @@
...
+     * Borrows a {@link ManagedChannel} from the connection pool.
+     *
+     * @return a managed gRPC channel
+     */
+    @Override
+    public ManagedChannel acquire() {
+        try {
+            return channelPool.borrowObject();
+        } catch (Exception e) {
+            throw new PoolException("Unable to borrow channel from pool: " + e.getMessage(), e);
```
cfkoehler, line 110: Am I saying this right?

The old returnChannel caught everything and invalidated. Now a PoolException from returnObject replaces the real gRPC exception. A non-retryable `StatusRuntimeException(INVALID_ARGUMENT)` gets swapped for a PoolException, which extends ChannelException and so makes retryOnException return true.

src/main/java/emissary/grpc/channel/impl/PooledChannelManager.java
```diff
@@ -0,0 +1,245 @@
...
+     * invalidate the channel to avoid pool corruption.
+     *
+     * @param channel the channel to return
+     */
+    @Override
+    public void release(ManagedChannel channel) {
+        try {
+            channelPool.returnObject(channel);
+        } catch (Exception e) {
+            shutdown(channel);
```
cfkoehler, line 126: The catch calls `shutdown(channel)`, which wraps `invalidateObject` failures in its own `PoolException`. If the channel isn't in the pool at all, you get "Invalidated object not currently part of this pool" instead of the real cause. `testReleaseUnavailableChannel` only covers the still-idle case. Wants an invalidateQuietly.

src/main/java/emissary/grpc/channel/ChannelManager.java
```diff
@@ -2,45 +2,204 @@
...
+    /**
+     * For use with {@link ChannelManager} implementations that expect channel exclusivity. Tells the manager that a
+     * {@link ManagedChannel} is no longer in use.
+     *
+     * @param channel the channel to release
+     */
+    public abstract void release(ManagedChannel channel);
+
+    /**
+     * Shuts down a {@link ManagedChannel} and frees its resources. Must be called on each channel that is acquired.
```
cfkoehler: This changes with pools right? 
`Must be called on each channel that is acquired` -> doing that would destroy every pooled channel on every RPC

rcameronm: Must be called at the end of its lifecycle, will adjust comment

src/main/java/emissary/grpc/channel/ChannelManager.java
rcameronm: ah true yeah that is a pain, I'll undo that

src/main/java/emissary/grpc/channel/impl/PooledChannelManager.java
rcameronm: I don't believe a `StatusRuntimeException` would have ever been thrown in the old `ChannelPoolFactory` class either. This exception type is thrown specifically when an RPC fails. The exceptions being thrown here are due to issues with either the pool or the gRPC channel itself. 

This comment and the other one left on R125 both seem to be related to how the new `shutdown(ManagedChannel)` class throws a PoolException on failure while the equivalent old `static invalidateChannel(ManagedChannel, ObjectPool<ManagedChannel>)` failed with only a log message. 

My thinking there was that, while I've never seen a channel actually fail to get returned to the pool or shutdown, it probably indicates other issues that are worth triggering a retry when applicable. I'm cool with changing it back to only logging a message though.

PR #1524, opened 2026-08-23
magic :: code clean up and documentation

dev-mlb:
- [x] Builds on #1536

This update continues modernizing our magic file processing logic, focusing on code cleanup, immutability, and documentation. Key improvements include:
- Immutability & Safety: Converted mutable rule objects (MagicNumber) into read-only, parser-built structures.
- Cleaner Data Types: Replaced legacy magic numbers with the new MagicDataType enum for self-documenting names, sizes, and byte orders.
- Streamlined Logic: Consolidated comparison logic into a single location and refactored the escape-sequence decoder (replacing a 256-row lookup table with a switch statement and removing 3 unused methods).
- Bug Fix: Fixed an issue where two UnixFile constructors accidentally loaded rules into a throwaway object, leaving zero detection rules (now properly chained).
- Documentation: Thoroughly updated Javadocs, comments, and the package-info comparison matrix.

### DIFF
--- a/src/main/java/emissary/util/UnixFile.java
+++ b/src/main/java/emissary/util/UnixFile.java
@@ -9,17 +9,17 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.List;
 
+/**
+ * Behaves like the UNIX file command. Magic number entries are loaded from one or more magic configuration files and
+ * samples are evaluated against them, falling back to a simple ASCII or binary test when nothing matches.
+ */
 public class UnixFile {
 
     private static final Logger log = LoggerFactory.getLogger(UnixFile.class);
 
-    /** The magic number configuration file. The file which contains all magic number entries */
-    private final List<File> magicFiles = new ArrayList<>();
-
-    /** The Magic number helper class */
+    /** The magic number helper class */
     private final MagicNumberUtil util = new MagicNumberUtil();
 
     /** The Binary file type description */
@@ -38,7 +38,7 @@ public class UnixFile {
      * @param magicFile the <code>File</code> containing magic number entries
      */
     public UnixFile(File magicFile) throws IOException {
-        new UnixFile(magicFile, false);
+        this(magicFile, false);
     }
 
     /**
@@ -53,7 +53,6 @@ public class UnixFile {
             throw new IllegalArgumentException("Magic file not found at: " + magicFile.getAbsolutePath());
         }
 
-        this.magicFiles.add(magicFile);
         util.load(magicFile, swallowParseException);
     }
 
@@ -63,12 +62,12 @@ public class UnixFile {
      * @param magicPaths the String names of magic files to load
      */
     public UnixFile(List<String> magicPaths) throws IOException {
-        new UnixFile(magicPaths, false);
+        this(magicPaths, false);
     }
 
     /**
      * Load multiple magic files into one identification engine
-     * 
+     *
      * @param magicPaths the String names of magic files to load
      * @param swallowParseException should we swallow Ignorable ParseException or bubble them up
      */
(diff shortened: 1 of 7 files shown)


src/main/java/emissary/util/magic/MagicDataType.java
```diff
@@ -0,0 +1,153 @@
...
+    DATE("DATE", 4, 4, false, false),
+
+    /** A 2-byte big-endian number. */
+    BESHORT("BESHORT", 5, 2),
+
+    /** A 4-byte big-endian number. */
+    BELONG("BELONG", 6, 4),
+
+    /** A 4-byte big-endian date stamp, skipped by the parser. */
+    BEDATE("BEDATE", 7, 4),
```
cfkoehler: With this change BEDATE becomes a supported type right? Just double checking
```
main:  '0 bedate 0x41424344 FOO' -> ParseException: Unsupported Data Type: bedate
1424:  '0 bedate 0x41424344 FOO' -> OK: 0     BEDATE  =0x41424344     FOO
```

src/main/java/emissary/util/magic/MagicNumber.java
cfkoehler, line 1: making note of the protected -> private scope changes. Not sure the impact of that downstream.

src/main/java/emissary/util/magic/MagicNumber.java
```diff
@@ -92,41 +129,78 @@ public class MagicNumber {
...
+    private final String description;
 
-    // Magic Number Properties
-    protected List<MagicNumber[]> dependencies;
+    private final List<List<MagicNumber>> dependencyLayers = new ArrayList<>();
+
+    /**
+     * Creates a fully specified rule. Intended to be called exclusively by the factory while parsing config files.
+     *
+     * @param depth how deeply nested this rule is (from column A) 5 *
```
cfkoehler:
```suggestion
     * @param depth how deeply nested this rule is (from column A)
```

src/main/java/emissary/util/magic/MagicNumber.java
```diff
@@ -92,41 +129,78 @@ public class MagicNumber {
...
+        this.unaryOperator = unaryOperator;
+        this.value = value;
+        this.substitute = substitute;
+        this.description = description;
+    }
 
     /**
-     * Recreates the string entry for this magic number plus its child continuations under new lines preceded by a '&gt;'
-     * character at the appropriate depth.
+     * Recreates the full config text for this rule plus all of its matching sub-rules indented underneath it.
```
cfkoehler: is it actually indenting the children here? I might be missing it and need to look at the output

src/main/java/emissary/util/magic/MagicDataType.java
dev-mlb: No, this change is just moving some things around and attempting to start code clean up. All behavior should be as before. Follow on changes will actually add support for more types. These types were moved to this enum from MagicNumber.

dev-mlb: @cfkoehler Instead of this code cleanup pr, maybe I should add more testing first to exercise what's there? That way we can sift thru what's implemented fully and what errors out. This can serve as a follow on pr to that. Thoughts?

cfkoehler: That will help. Thanks

sylvesterkaczmarek (commented):
Hi, `fromKey()` uses `key.toUpperCase(Locale.getDefault())`, which makes magic-file parsing depend on the host locale. On a Turkish locale, for example, `string` becomes `STRİNG` and will not match the enum key `STRING`, so a valid entry can stop parsing. `Locale.ROOT` (or `equalsIgnoreCase`) would keep this deterministic across hosts.

src/main/java/emissary/util/UnixFile.java
```diff
@@ -63,12 +62,12 @@ public UnixFile(File magicFile, boolean swallowParseException) throws IOExceptio
      * @param magicPaths the String names of magic files to load
      */
     public UnixFile(List<String> magicPaths) throws IOException {
-        new UnixFile(magicPaths, false);
```
drivenflywheel, line 66: 🤦‍♂️

PR #1531, opened 2026-08-25
IBDO batch processing implementation

rcameronm:
Followed existing patterns as closely as possible. The intention was to not be disruptive towards existing processes.

### DIFF
--- a/src/main/java/emissary/place/CoordinationPlace.java
+++ b/src/main/java/emissary/place/CoordinationPlace.java
@@ -314,7 +314,14 @@ public class CoordinationPlace extends ServiceProviderPlace {
                 }
             }
 
-            selected.stream().filter(d -> Form.ERROR.equals(d.currentForm())).forEach(errors::add);
+            selected.stream()
+                    .filter(d -> Form.ERROR.equals(d.currentForm()))
+                    .forEach(errors::add);
+
+            active.removeIf(d -> !shouldContinue(p, errors.contains(d)));
+            if (active.isEmpty()) {
+                break;
+            }
         }
 
         for (IBaseDataObject d : dList) {
--- a/src/main/java/emissary/place/CoordinationPlace.java
+++ b/src/main/java/emissary/place/CoordinationPlace.java
@@ -319,7 +319,6 @@ public class CoordinationPlace extends ServiceProviderPlace {
 
         for (IBaseDataObject d : dList) {
             applyForm(d, errors.contains(d));
-            sproutHook(sprouts, d);
             cleanUpHook(d);
         }
 
@@ -341,7 +340,9 @@ public class CoordinationPlace extends ServiceProviderPlace {
             for (IBaseDataObject d : selected) {
                 updateTransformHistory(d, p);
                 try {
-                    sprouts.addAll(p.agentProcessHeavyDuty(d));
+                    List<IBaseDataObject> s = p.agentProcessHeavyDuty(d);
+                    sprouts.addAll(s);
+                    sproutHook(s, d); // only calls hook when we know the sprout's parent
                 } catch (Exception e) {
                     errors.add(d);
                     handlePlaceException(p, true, e);
--- a/src/main/java/emissary/place/CoordinationPlace.java
+++ b/src/main/java/emissary/place/CoordinationPlace.java
@@ -18,9 +18,12 @@ import org.apache.commons.collections4.CollectionUtils;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.IdentityHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import static emissary.core.constants.Configurations.OUTPUT_FORM;
 
@@ -50,6 +53,8 @@ public class CoordinationPlace extends ServiceProviderPlace {
 
     // set of coordination places that failed to be created/did not exist
     protected static final Set<String> failedCoordPlaceCreation = new LinkedHashSet<>();
+
+    protected boolean batchProcessingEnabled;
 
     /**
      * Create the place using the supplied configuration and location
@@ -143,6 +148,13 @@ public class CoordinationPlace extends ServiceProviderPlace {
                 }
             }
         }
+
+        batchProcessingEnabled = placeRefs.stream().anyMatch(IServiceProviderPlace::isBatchProcessingEnabled);
+    }
+
+    @Override
+    public boolean isBatchProcessingEnabled() {
+        return batchProcessingEnabled;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -267,6 +279,78 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
         return sproutCollection;
     }
 
+    /**
+     * Consume a list of data objects and coordinates their processing
+     *
+     * @param dList the payload list to process
+     * @return the list of sprouted data objects
+     */
+    protected List<IBaseDataObject> coordinateBatch(List<IBaseDataObject> dList) {
```
cfkoehler, line 288: Does this need to have a concept of https://github.com/NationalSecurityAgency/emissary/pull/1531/changes#diff-684067cf302064ee522782eeff48e8f1a8b4b8390128e5a2c6d83cc8b4de697aR262

For example
```
coord.isBatchProcessingEnabled=true continueOnError=false
BATCH  doc0 VISITED_P2=yes forms=[ERROR]     <-- should never have reached P2
BATCH  doc1 VISITED_P2=yes forms=[DONE]
BATCH  doc2 VISITED_P2=yes forms=[DONE]
```

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -267,6 +279,78 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
...
+                    logger.warn("Place {} was interrupted during execution", p);
+                }
+            }
+
+            selected.stream().filter(d -> Form.ERROR.equals(d.currentForm())).forEach(errors::add);
+        }
+
+        for (IBaseDataObject d : dList) {
+            applyForm(d, errors.contains(d));
+            sproutHook(sprouts, d);
```
cfkoehler: This assuming all sprouts have the same parent right?

src/main/java/emissary/place/ServiceProviderPlace.java
cfkoehler, line 419: Think this can throw an NPE. 
only inspects `params.length == 1 && params[0].isAssignableFrom(IBaseDataObject.class)`, so the flags describe the single-payload methods. But the new List overloads use those same flags to proxy to each other.

Also nothing validates that a batch-enabled place actually implements a List overload. Maybe worth adding?

src/main/java/emissary/place/CoordinationPlace.java
rcameronm: Ah true, that is a good point. The base implementation of `sproutHook` just logs a message, but I don't know what every inheritor does with it. Probably don't wanna make this assumption, will change.

src/main/java/emissary/place/ServiceProviderPlace.java
rcameronm: my thinking was that the multi-payload methods would only be applicable if the corresponding single-payload method was enabled... but this isn't very intuitive. I'll make some changes.

rcameronm:
@cfkoehler Okay I think I've addressed each of your comments. Lmk if you want me to go over any of the changes with you or if you don't think I addressed anything sufficiently.

src/main/java/emissary/place/IServiceProviderPlace.java
```diff
@@ -110,13 +110,37 @@ public interface IServiceProviderPlace {
      */
     List<IBaseDataObject> processHeavyDuty(IBaseDataObject payload) throws ResourceException;
 
+    /**
+     * Override point for HD Agent calls
+     *
+     * @param payloadList list of sibling payloads to be processed
+     * @return list of IBaseDataObject result attachments
+     */
+    List<IBaseDataObject> processHeavyDuty(List<IBaseDataObject> payloadList) throws ResourceException;
```
dev-mlb, line 119: This is a breaking change for any class that implements this interface that does not extend ServiceProviderPlace. Maybe default this to throw UnsupportedOperationException?

src/main/java/emissary/place/IServiceProviderPlace.java
```diff
@@ -110,13 +110,37 @@ public interface IServiceProviderPlace {
...
      * @param payload the payload to be processed
      */
     void process(IBaseDataObject payload) throws ResourceException;
 
+    /**
+     * Override point for non-HD agent calls
+     *
+     * @param payloadList list of sibling payloads to be processed
+     */
+    void process(List<IBaseDataObject> payloadList) throws ResourceException;
```
dev-mlb, line 142: Same here -- this is a breaking change for any class that implements this interface that does not extend ServiceProviderPlace. Maybe default this to throw UnsupportedOperationException?

src/main/java/emissary/place/ServiceProviderPlace.java
```diff
@@ -263,29 +266,37 @@ public void agentProcessCall(IBaseDataObject payload) throws ResourceException {
...
-                List<IBaseDataObject> l = agentProcessHeavyDuty(dataObject);
-                if (!l.isEmpty()) {
-                    dataObject.setNumChildren(dataObject.getNumChildren() + l.size());
-                }
-
-                // Accumulate results in a list to return
-                list.addAll(l);
+                MDC.put(MDCConstants.SERVICE_LOCATION, this.getKey());
+                list = processHeavyDuty(payloadList);
+                payloadList.forEach(this::rehash);
```
dev-mlb, line 277: There is no setNumChildren in this block, is this the correct behavior? I don't think that gets set elsewhere, no?

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -267,6 +279,86 @@ protected List<IBaseDataObject> coordinate(IBaseDataObject d, boolean hd) {
...
+    }
+
+    private List<IBaseDataObject> coordinateBatch(IServiceProviderPlace p, List<IBaseDataObject> selected, Set<IBaseDataObject> errors) {
+        if (p.isBatchProcessingEnabled()) {
+            selected.forEach(d -> updateTransformHistory(d, p));
+            try {
+                return p.agentProcessHeavyDuty(selected);
+            } catch (Exception e) {
+                errors.addAll(selected);
+                handlePlaceException(p, true, e);
```
dev-mlb, line 342: `sproutHook` is never called, is that expected? We may not know the parent, but are we losing functionality?

src/main/java/emissary/place/ServiceProviderPlace.java
```diff
@@ -263,29 +266,37 @@ public void agentProcessCall(IBaseDataObject payload) throws ResourceException {
...
             } catch (Exception e) {
-                logger.error("Place.process exception", e);
-                dataObject.addProcessingError("agentProcessHD(" + keys.get(0) + "): " + e);
-                dataObject.replaceCurrentForm(Form.ERROR);
+                payloadList.forEach(d -> handleProcessingError(d, e));
```
dev-mlb, line 279: If we spent a ton of time processing docs, and one errors out, we error out all of it and throw away what we've extracted? I'm wondering if this needs to be config driven. I think I can see a case where we'd want to keep the processed docs, but maybe we wouldn't??

src/main/java/emissary/place/CoordinationPlace.java
```diff
@@ -143,6 +148,13 @@ protected void configurePlace() {
                 }
             }
         }
+
+        batchProcessingEnabled = placeRefs.stream().anyMatch(IServiceProviderPlace::isBatchProcessingEnabled);
```
dev-mlb, line 152: This is a little confusing. The coordinator doesn't own this. I'm wondering if this should be a config to enable batch processing for a full set of places that are being coordinated?

PR #1536, opened 2026-09-01
magic :: unit testing

dev-mlb:


cfkoehler (approved):

sambish5 (approved):

sylvesterkaczmarek (commented):
This test turns a signed-byte formatting bug into expected behaviour: `0xBB` is asserted as `-45`, so a hex string becomes `0x41-45`. Format bytes as `b & 0xff` with two hex digits and make the test expect `0x41bb` instead of locking in the bug.

cfkoehler:
> This test turns a signed-byte formatting bug into expected behaviour: `0xBB` is asserted as `-45`, so a hex string becomes `0x41-45`. Format bytes as `b & 0xff` with two hex digits and make the test expect `0x41bb` instead of locking in the bug.

@sylvesterkaczmarek 
Thanks for the comment. The purpose of this pr is to demonstrate the current behavior in unit test form. There are follow on prs to correct the behavior.

PR #1539, opened 2026-09-03
magic :: bug fixes

dev-mlb:
- [ ]  builds on #1524 

Fixes
1. A byte like 0xBB was printing as -45 instead of bb. The fix treats each byte as the 0-255 value it really is and prints the proper two-digit hex.
2. When a magic rule's description pulls actual bytes out of the file (e.g. %d for a version number), it was interpreting the bytes as a signed number. So FF FF FF FF (max unsigned value) printed as -1. Now it treats them as unsigned, so it prints the big positive number.
3. The logic checked "is there 3+ bytes left?" but only actually needed 1 byte. So a 2-byte file never got its value substituted at all. Now it only requires the 1 byte it actually reads.
4. For a string rule with a %d placeholder, it was outputting the raw character (e.g. A) instead of its numeric value (65). Now %d/%ld on a string yields the number.
5. A rule like >-5 (look 5 bytes before the current spot) would try to read from a negative position and throw an error. Now it just treats that as "no match" instead of crashing.
6. When asked to produce a specific-length array, the hex path ignored the size (returned 0x41 0x42) while the decimal path respected it (padded to 00 00 41 42). Now both pad consistently to the requested size.

### DIFF
--- a/src/main/java/emissary/util/UnixFile.java
+++ b/src/main/java/emissary/util/UnixFile.java
@@ -9,17 +9,17 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.List;
 
+/**
+ * Behaves like the UNIX file command. Magic number entries are loaded from one or more magic configuration files and
+ * samples are evaluated against them, falling back to a simple ASCII or binary test when nothing matches.
+ */
 public class UnixFile {
 
     private static final Logger log = LoggerFactory.getLogger(UnixFile.class);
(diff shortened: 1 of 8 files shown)


sylvesterkaczmarek (commented):
`fromKey()` normalizes magic-file keywords with `Locale.getDefault()`, so parsing depends on the machine locale: under Turkish, `string` uppercases to `STRİNG` and no longer matches `STRING`. Use `Locale.ROOT` (or explicit ASCII-insensitive matching) and add a Turkish-locale regression test.

PR #1555, opened 2026-09-17
use non-deprecated createPlace constructor in PlaceStarter

sambish5:
the createPlace constructors in the PlaceStarter all pointed to the deprecated constructor, which then to the non-deprecated constructor. Update to no longer use the deprecated constructor and just use the correct one.


src/main/java/emissary/admin/PlaceStarter.java
```diff
@@ -63,7 +63,7 @@ public static IServiceProviderPlace createPlace(final String theLocation, final
 
         final Object[] constructorArgs = {theConfigFile, directory, theLocation};
```
dev-mlb: Should we just create this as a list and pass that in?

cfkoehler (approved):

PR #1565, opened 2026-09-20
java :: use getFirst() on lists; @Serial; dangling doc comments

dev-mlb:


julianvanceblackwood:
@dev-mlb On the JDOM change, would it be useful to add a focused regression test that keeps DOCTYPE declarations with an internal subset working while confirming external DTD and entity resolution remain disabled? That seems like a useful way to pin the intended parser boundary during the Java 21 migration. I can help reduce it to a small test case if useful.

